PR #38, opened 2015-11-19
EAPI 6 needs to be documented

maekke:
See https://bugs.gentoo.org/show_bug.cgi?id=565582

### DIFF
--- a/ebuild-writing/eapi/text.xml
+++ b/ebuild-writing/eapi/text.xml
@@ -857,6 +857,141 @@ If USE flag is set, echo [true output][true suffix] (defaults to "yes"),
 </body>
 </section>
 
+<section>
+<title>EAPI=6</title>
+<body>
+
+<subsection>
+<title>Bash version</title>
+<body>
+<p>Ebuilds can use features of Bash version 4.2 (was 3.2 before).</p>
+</body>
+</subsection>
+
+<subsection>
+<title>Ebuild environment</title>
+<body>
+<ul>
+	<li>
+		<p><b>Locale settings</b></p>
+		<p>It is ensured that the behaviour of case modification and collation order for ASCII characters (<c>LC_CTYPE</c> and <c>LC_COLLATE</c>) are the same as in the POSIX locale.</p>
+	</li>
+	<li>
+		<p><b><c>failglob</c> enabled</b></p>
+		<p>
+		The <c>failglob</c> option of Bash is set in global scope, so that unintentional pattern expansion will be caught as an error.
+		</p>
+	</li>
+</ul>
+</body>
+</subsection>
+
+<subsection>
+<title>Phases</title>
+<body>
+<ul>
+	<li>
+		<p><b>New <c>src_prepare</c> Phase Function</b></p>
+		<p>
+			This phase is no longer a no-op, it supports applying patches via the <c>PATCHES</c> variable and applying user patches via <c>eaply_user</c>. The default <c>src_prepare</c> looks like this:
+			<codesample lang="ebuild">
+src_prepare() {
+	if declare -p PATCHES | grep -q "^declare -a "; then
+		[[ -n ${PATCHES[@]} ]] &amp;&amp; eapply "${PATCHES[@]}"
+	else
+		[[ -n ${PATCHES} ]] &amp;&amp; eapply ${PATCHES}
+	fi
+	eapply_user
+}
+			</codesample>
+		</p>
+	</li>
+	<li>
+		<p><b>New <c>src_install</c> Phase Function</b></p>
+		<p>
+			This phase uses the new <c>einstalldocs</c> function for installation of documentation. The default <c>src_install</c> looks like this:
+			<codesample lang="ebuild">
+src_install() {
+	if [[ -f Makefile ]] || [[ -f GNUmakefile ]] || [[ -f makefile ]]; then
+		emake DESTDIR="${D}" install
+	fi
+	einstalldocs
+}
+			</codesample>
+		</p>
+	</li>
+</ul>
+</body>
+</subsection>
+
+<subsection>
+<title>Helpers</title>
+<body>
+<ul>
+	<li>
+		<p><b><c>einstall</c> banned</b></p>
+		<p>
+			The <c>einstall</c> helper has been banned with <c>EAPI=6</c>.
+		</p>
+	</li>
+	<li>
+		<p><b><c>nonfatal die</c></b></p>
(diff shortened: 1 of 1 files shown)


jlec:
@mgorny Could you please take a look if that is all correct?

ebuild-writing/eapi/text.xml
```diff
@@ -857,6 +857,141 @@ If USE flag is set, echo [true output][true suffix] (defaults to "yes"),
...
+<body>
+<ul>
+	<li>
+		<p><b>Locale settings</b></p>
+		<p>It is ensured that the behaviour of case modification and collation order for ASCII characters (<c>LC_CTYPE</c> and <c>LC_COLLATE</c>) are the same as in the POSIX locale.</p>
+	</li>
+	<li>
+		<p><b><c>failglob</c> enabled</b></p>
+		<p>
+		The <c>failglob</c> option of Bash is set in global scope, so that unintentional pattern expansion will be caught as an error.
```
mgorny: Well, that's not 100% correct. The failing pattern expansions will be caught, so rather potential unintentional pattern expansion ;-). If the expansion actually happens, it will still work (and cause weird results).

ebuild-writing/eapi/text.xml
```diff
@@ -857,6 +857,141 @@ If USE flag is set, echo [true output][true suffix] (defaults to "yes"),
...
+</ul>
+</body>
+</subsection>
+
+<subsection>
+<title>Phases</title>
+<body>
+<ul>
+	<li>
+		<p><b>New <c>src_prepare</c> Phase Function</b></p>
```
mgorny: It's a new default implementation, not a new phase function ;-).

ebuild-writing/eapi/text.xml
```diff
@@ -857,6 +857,141 @@ If USE flag is set, echo [true output][true suffix] (defaults to "yes"),
...
+	<li>
+		<p><b><c>nonfatal die</c></b></p>
+		<p>
+			When <c>die</c> or <c>assert</c> are called under the <c>nonfatal</c> command and with the <c>-n</c> option, they will not abort the build process but return with an error.
+		</p>
+	</li>
+	<li>
+		<p><b><c>eapply</c> support</b></p>
+		<p>
+			The <c>eapply</c> command is a simplified substitute for <c>epatch</c> (from <c>eutils.eclass</c>), implemented in the package manager. The patches from its file or directory arguments are applied using patch <c>-p1</c>.
```
mgorny: You miss the fact that it accepts `patch(1)` options, so you can make it with with non-`-p1` patches.

ebuild-writing/eapi/text.xml
```diff
@@ -857,6 +857,141 @@ If USE flag is set, echo [true output][true suffix] (defaults to "yes"),
...
+		</p>
+	</li>
+	<li>
+		<p><b><c>eapply_user</c> support</b></p>
+		<p>
+			The <c>eapply_user</c> command permits the package manager to apply user-provided patches. It must be called from every <c>src_prepare</c> function.
```
mgorny: I'd mention that it doesn't need to be called explicitly when `default_src_prepare` is called.

ebuild-writing/eapi/text.xml
```diff
@@ -857,6 +857,141 @@ If USE flag is set, echo [true output][true suffix] (defaults to "yes"),
...
+	<li>
+		<p><b><c>econf</c> adds <c>--docdir</c> and <c>--htmldir</c></b></p>
+		<p>
+			Options <c>--docdir</c> and <c>--htmldir</c> are passed to <c>configure</c>, in addition to the existing options.
+		</p>
+	</li>
+	<li>
+		<p><b><c>in_iuse</c> support</b></p>
+		<p>
+			The <c>in_iuse</c> function returns <c>true</c> if the <c>USE</c> flag given as its argument is available in the ebuild for <c>USE</c> queries.
```
mgorny: This is too PMS-y (i.e. muddy) for devmanual. Not that I can think of a good replacement.

ebuild-writing/eapi/text.xml
```diff
@@ -857,6 +857,141 @@ If USE flag is set, echo [true output][true suffix] (defaults to "yes"),
...
+</body>
+</subsection>
+
+<subsection>
+<title>Ebuild environment</title>
+<body>
+<ul>
+	<li>
+		<p><b>Locale settings</b></p>
+		<p>It is ensured that the behaviour of case modification and collation order for ASCII characters (<c>LC_CTYPE</c> and <c>LC_COLLATE</c>) are the same as in the POSIX locale.</p>
```
mgorny: Please try to word that less PMS-y and more practically ;-). Probably the most important point would be that you don't have to use `LC_CTYPE=C` and `LC_COLLATE=C` explicitly.

ebuild-writing/eapi/text.xml
```diff
@@ -857,6 +857,141 @@ If USE flag is set, echo [true output][true suffix] (defaults to "yes"),
...
+	<li>
+		<p><b><c>in_iuse</c> support</b></p>
+		<p>
+			The <c>in_iuse</c> function returns <c>true</c> if the <c>USE</c> flag given as its argument is available in the ebuild for <c>USE</c> queries.
+		</p>
+	</li>
+	<li>
+		<p><b><c>unpack</c> changes</b></p>
+		<ul>
+			<li><p><c>unpack</c> supports absolute and relative paths.</p></li>
```
mgorny: If I'm supposed to be picky, `unpack` already supported relative paths via `./foo.tar.gz`. The new feature is that `foo/bar.tar.gz` works too (but plain `bar.tar.gz` references DISTDIR).

ebuild-writing/eapi/text.xml
```diff
@@ -857,6 +857,141 @@ If USE flag is set, echo [true output][true suffix] (defaults to "yes"),
...
+			<li><p><c>unpack</c> supports <c>.txz</c> (xz compressed tarball).</p></li>
+			<li><p><c>unpack</c> matches filename extensions case-insensitively.</p></li>
+		</ul>
+	</li>
+	<li>
+		<p><b><c>einstalldocs</c> support</b></p>
+		<p>
+			The <c>einstalldocs</c> function will in-stall the files specified by the <c>DOCS</c> variable (or a default set of files if <c>DOCS</c> is unset) and by the <c>HTML_DOCS</c> variable.
```
mgorny: `in-stall`? ;-)

PR #40, opened 2016-02-22
EAPI for src_pretend is 4 

vice:
EAPI for src_pretend is 4  as documented in Package Manager Specification


vice:
See pull request #41

PR #41, opened 2016-02-22
EAPI for src_pretend is 4 not 3

vice:
EAPI for src_pretend is 4 not 3 as documented in Package Manager Specification


vice:
Now with a commit message....

ulm:
Merged, thanks.
It's "pkg_pretend" not "src_pretend", though. I've taken the liberty to amend your commit message accordingly.

PR #42, opened 2016-03-24
typo corrected

Lightyagami1:
last section describing EAPI=6, under Helpers subsection in unpack changes the first point has a typo.

>unpack supports "elative" paths without leading ./ (unpack foo/bar.tar.gz is valid as relative path).


gktrk:
this is fixed in 7f647d3e3a3b3f029a9c44ef20fd8f9a3077181b

PR #43, opened 2016-03-31
Fix a typo, update copyright year and replace an instance of 'herd' with 'project'

sbraz:
This will not fix all occurrences of 'herd' but it is a start.


gktrk:
all merged. thanks!

PR #44, opened 2016-05-12
devbook.xsl: Change HTML title format

Whissi:
This PR changes devmanual's HTML title format from
```
Gentoo Development Guide: $document_title
```
into
```
$document_title – Gentoo Development Guide
```

Bug: https://bugs.gentoo.org/show_bug.cgi?id=582750


ulm:
Merged, thanks.

PR #46, opened 2016-05-19
ebuild-writing/error-handling: remove outdated info regarding die

rindeal:
The outdated info is only misleading and EAPI < 4 is long deprecated anyway.

### DIFF
--- a/ebuild-writing/error-handling/text.xml
+++ b/ebuild-writing/error-handling/text.xml
@@ -47,11 +47,8 @@ function can die in multiple places.
 </p>
 
 <p>
-Some portage-provided functions will automatically die upon failure. Others will
-not. It is for example safe to omit the <c>|| die</c> after a call to <c>epatch</c>,
-but not <c>emake</c>. The reason is that external binaries are not able to call
-die that is a bash function. You can see what commands are external binaries
-with <c>ls /usr/lib*/portage/bin/ebuild-helpers</c>. In <uri link="::ebuild-writing/eapi/#eapi=4">EAPI>=4</uri> all ebuild-helpers automatically die upon failure.
+In <uri link="::ebuild-writing/eapi/#eapi=4">EAPI>=4</uri> all portage-provided functions
+and ebuild-helpers automatically die upon failure.
 </p>
 
 <p>


gktrk:
I agree that it's misleading but I do not agree with the proposed patch. EAPI dependent failure defined by the PMS only applies to (some) ebuild helpers. "Some portage-provided functions will automatically die upon failure. Others will not." is still valid. egit_clean() from eutils does not die on failure for example.

I think "The reason is that external binaries are not able to call die that is a bash function." refers to the PMS's description of ebuild functions. For example:
<b>econf:</b> econf must be implemented internally—that is, as a bash function and not an external script.
<b>emake:</b> emake must be an external program and cannot be a function or alias—it must be callable from e. g. xargs.
It is probably little less relevant nowadays because emake is a separate "program" (in the form of a shell script) and it will die on failure for EAPI >=4.

We should probably remove the path to the ebuild helpers since that path is incorrect, but still keep an accessible link to them in some form or shape.

Regarding EAPI < 4 being long deprecated, I agree with you. But there is an EAPI 0 question in one of the developer quizzes and it points to devmanual for an answer.

rindeal:
> I agree that it's misleading but I do not agree with the proposed patch. EAPI dependent failure defined by the PMS only applies to (some) ebuild helpers. "Some portage-provided functions will automatically die upon failure. Others will not." is still valid. egit_clean() from eutils does not die on failure for example.

I don't think that's intentional, it looks like an untouched legacy code from the pre-EAPI4 era. And the reason for it being untouched is probably because it doesn't trouble anyone if the function fails or not. Show me a serious function written for EAPI>=4 that doesn't die.

> I think "The reason is that external binaries are not able to call die that is a bash function." refers to the PMS's description of ebuild functions. For example:
econf: econf must be implemented internally—that is, as a bash function and not an external script.
emake: emake must be an external program and cannot be a function or alias—it must be callable from e. g. xargs.
It is probably little less relevant nowadays because emake is a separate "program" (in the form of a shell script) and it will die on failure for EAPI >=4.

Yes, it's irrelevant.

> We should probably remove the path to the ebuild helpers since that path is incorrect, but still keep an accessible link to them in some form or shape.

I cannot help you with this as that's beyond my expertise.

> Regarding EAPI < 4 being long deprecated, I agree with you. But there is an EAPI 0 question in one of the developer quizzes and it points to devmanual for an answer.

Keeping outdated and often misleading information in the manual because of some question with a really questionable intent would be really unjustified. I consider anything (except historical books), even mentioning something from the pre-EAPI4, to be a bloat.

PR #48, opened 2016-09-26
Update keepdir description

mrueg:
The file has a different name.

### DIFF
--- a/function-reference/install-functions/text.xml
+++ b/function-reference/install-functions/text.xml
@@ -267,7 +267,7 @@ the first is the source name, the second the name to use when installing.
       <c>keepdir</c>
     </ti>
     <ti>
-      Create a directory with an empty <c>.keep</c> file in it.
+      Create a directory which will be preserved by the package manager. This is typically accomplished by inserting an empty <c>.keep_${CATEGORY}_${PN}-${SLOT%/*}</c> file in the directory.
     </ti>
   </tr>
   <tr>


gktrk:
PMS doesn't say anything about the `.keep_${CATEGORY}_${PN}-${SLOT%/*}` syntax, it only says `an empty file whose name starts with .keep in that directory`. Should we rather say something like "Create an empty directory which will be preserved by the package manager. This is typically accomplished by inserting an empty file to the directory with the name .keep_${CATEGORY}_${PN}-${SLOT%/*}." ?

ulm:
>>>>> On Mon, 26 Sep 2016, Gokturk Yuksek wrote:

> PMS doesn't say anything about the .keep_${CATEGORY}_${PN}-${SLOT%/*} syntax,
> it only says an empty file whose name starts with .keep in that directory.
> Should we rather say something like "Create an empty directory which will be
> preserved by the package manager. This is typically accomplished by inserting
> an empty file to the directory with the name .keep_${CATEGORY}_${PN}-${SLOT%/
> *}." ?

If there is a file in the dir then the dir isn't empty.
Therefore: s/an empty directory/a directory/

mrueg:
@ulm @gktrk mind to check the wording again? Thanks!

gktrk:
LGTM

gktrk:
Merged with minor punctuation changes (so that we don't have a full stop at the end). Thanks!

PR #49, opened 2016-10-03
quickstart: Update to latest EAPI

mrueg:
Drop dohtml, as it is deprecated in EAPI=6 and we should not encourage
new users to use it.

### DIFF
--- a/quickstart/text.xml
+++ b/quickstart/text.xml
@@ -38,7 +38,7 @@ can see real ebuilds in the main tree).
 # Distributed under the terms of the GNU General Public License v2
 # &#36;Id&#36;
 
-EAPI=5
+EAPI=6
 
 DESCRIPTION="Exuberant ctags generates tags files for quick source navigation"
 HOMEPAGE="http://ctags.sourceforge.net"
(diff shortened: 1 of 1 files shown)


gktrk:
I think we should include "EXTENDING.html ctags.html" there too just to show that this is how html files are installed in EAPI 6. What do you think?

mrueg:
Do you want to do that via HTML_DOCS? I thought for the introduction staying backwards compatible is a good thing, that's why I just removed it.

ulm:
I'd rather not add anything complicated to the "first ebuild" example. That is, I believe neither "docinto html; doins foo.html bar.html" nor HTML_DOCS would be appropriate there.

One could consider adding one of those to the last "more complicated example", i.e. to the sylpheed ebuild (or replace it by another ebuild taken from the tree that has them).

gktrk:
Sounds fair to me, let's keep it simple.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/commit/003c9dc975e405c979b5fd285e23bd25ce9711e4. Thanks!

Just changed the commit message to indicate that only the first ebuild example is updated as the rest still shows EAPI 5.

PR #50, opened 2016-11-03
file-format: Fix missing <c/> style around _p suffix

mgorny:


gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/9623823...0854537. We tend to use the full path instead of the basename, so I've amended the commit before merge. Hope that's ok. Thanks!

PR #51, opened 2016-11-13
function-reference/build-install-functions: removed banned dohard, added specific eapi ban notes for einstall and dohtml, #564620

wimmuskee:
Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=564620

Considered the latest PMS in which EAPI 0-3 are banned in Gentoo. 
- dohard is banned in EAPI 4, and should not be used and (like dosed already) be removed from the devmanual
- dohtml and einstall could still be used in valid EAPI's, just a note for those

### DIFF
--- a/function-reference/install-functions/text.xml
+++ b/function-reference/install-functions/text.xml
@@ -144,14 +144,6 @@ the first is the source name, the second the name to use when installing.
   </tr>
   <tr>
     <ti>
-      <c>dohard</c>
-    </ti>
-    <ti>
-      Create a hardlink to the first argument from the second argument
-    </ti>
-  </tr>
-  <tr>
-    <ti>
       <c>doheader</c>
     </ti>
     <ti>
@@ -164,6 +156,7 @@ the first is the source name, the second the name to use when installing.
     </ti>
     <ti>
       Installs HTML document files into <c>/usr/share/doc/${PF}/html</c>
+      <b>Note</b>: Banned in EAPI 7.
     </ti>
   </tr>
   <tr>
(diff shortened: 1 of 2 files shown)


function-reference/install-functions/text.xml
```diff
@@ -164,6 +156,7 @@ the first is the source name, the second the name to use when installing.
     </ti>
     <ti>
       Installs HTML document files into <c>/usr/share/doc/${PF}/html</c>
+      <b>Note</b>: Banned in EAPI 7.
```
gktrk: EAPI 7 isn't approved by the council, so it would be wrong of us to claim that.

ulm:
>>>>> On Mon, 14 Nov 2016, Gokturk Yuksek wrote:

> +      <b>Note</b>: Banned in EAPI 7.

> EAPI 7 isn't approved by the council, so it would be wrong of us to
> claim that.

From the summary of the 2014-09-09 council meeting:

# - Deprecate dohtml now, ban it in EAPI 7?
#   Accepted (6 yes, 1 no).

Still, I would suggest that we don't explicitly mention EAPI 7,
but add a note that dohtml is deprecated.

wimmuskee:
changed the "banned in EAPI 7" to "deprecated in EAPI 6"

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/0854537...7765886. Thanks!

Squashed two commits into one and mangled the commit message a bit. Hope that's not a problem for you.

PR #52, opened 2016-11-13
ebuild-writing/variables: add explanation for no homepage to HOMEPAGE

wimmuskee:
Found the usage of https://wiki.gentoo.org/wiki/No_homepage back in several bugreports as a suggestion. Also have some mention for it in the devmanual.

### DIFF
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -168,8 +168,10 @@ The following variables may or must be defined by every ebuild.
     Package's homepage. Mandatory (except for virtuals). If you are
     unable to locate an official one, try to provide a link to
     <uri link="http://freshmeat.net">freshmeat.net</uri> or a similar
-    package tracking site. Never refer to a variable name in the
-    string; include only raw text.
+    package tracking site. If the homepage of the package is unavailable
+    and no new location can be found, please update it to
+    <c>https://wiki.gentoo.org/wiki/No_homepage</c>.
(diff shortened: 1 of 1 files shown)


gktrk:
I wonder if we should remove https://devmanual.gentoo.org/ebuild-maintenance/index.html#homepage-unavailable and replace it with your patch then.

wimmuskee:
In most cases changing the homepage to a "no_homepage" is a maintenance use case. I suspect when adding a new package, it is actively maintained and has an actual homepage.

In a more general sense regarding documentation, I think most people don't read the whole manual, but only the things that relate to their use case at that time. In that way, the devmanual functions as a reference (with lists of all functionality), but also as a manual (connecting the functionality to use cases).

That being said, I don't know if that is the official design of the devmanual.

ulm:
>>>>> On Tue, 15 Nov 2016, Gokturk Yuksek wrote:

> I wonder if we should remove 
> https://devmanual.gentoo.org/ebuild-maintenance/index.html#homepage-unavailable
> and replace it with your patch then.

I wouldn't. It is only one short sentence, and some redundancy doesn't
harm. Also, disappearing homepages tend to occur later in an ebuild's
life, so IMHO the primary mention belongs in the maintenance section.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/7765886...9663bc6. Thanks!

PR #53, opened 2016-11-13
general-concepts/tree: Add mention for metadata directory in portage tree, #572150

wimmuskee:
Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=572150


gktrk:
we should note that only `layout.conf` exists in the git tree. the rest are only visible on user systems. I don't have `timestamp` or `projects.xml` on my system for example. The injection is explained more in depth in [git to rsync](https://devmanual.gentoo.org/general-concepts/git-to-rsync/index.html).

wimmuskee:
I see now that most of the Github tree metadata contents does not include any of my listed files. For instance, also news and xml-schema or not there.
Are they all being generated/included in the git-to-rsync process? Because if so, I'll make a more general mention of it.

gktrk:
`projects.xml` is part of the [Gentoo API](https://api.gentoo.org/) and autogenerated from the wiki. [news](https://gitweb.gentoo.org/data/gentoo-news.git/), [glsa](https://gitweb.gentoo.org/data/glsa.git/), and [xml-schema](https://gitweb.gentoo.org/data/xml-schema.git/) are separate repos. The dtd is only used by the old versions of portage, and kept for compatibility I think.

general-concepts/tree/text.xml
```diff
@@ -38,6 +38,24 @@ The basic layout of the portage tree is as follows:
   <li>Eclasses directory, <c>eclass/</c></li>
   <li>Licenses directory, <c>licenses/</c></li>
   <li>
+  Metadata directory, <c>metadata/</c>. Most of the listed contents are not
+  included in the main Git tree, but are generated by the
+  <uri link="::general-concepts/git-to-rsync">Git to Rsync</uri> process or
+  included there from other repositories.
```
gktrk: s/there//

gktrk: The inclusion of files from other repos is also a part of the git-to-rsync process. So we should rewrite this sentence. How about the following? :

```
Metadata directory, <c>metadata/</c>. Most of the listed contents are not kept directly in the main Git tree but instead auto-generated or included from other repositories as part of the <uri link="::general-concepts/git-to-rsync">Git to Rsync</uri> process.
```

general-concepts/tree/text.xml
```diff
@@ -38,6 +38,24 @@ The basic layout of the portage tree is as follows:
...
+  <ul>
+    <li>
+    Various control files, for example <c>layout.conf</c>, <c>timestamp</c>
+    and <c>projects.xml</c>. Only <c>layout.conf</c> exists in the main Git tree.
+    </li>
+    <li>
+    XML validation files in <c>metadata/dtd/</c> and <c>metadata/xml-schema/</c>.
+    </li>
+    <li>
+    News updates in <c>metadata/glsa/</c> and <c>metadata/news/</c>.
```
gktrk: `Security advisories in <c>metadata/glsa/</c> and news items in <c>metadata/news/</c>` How does this sound?

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/9bbf459...fe224c4 with some changes. Thanks!

PR #54, opened 2016-11-13
Portage is the package manager. The repository is named gentoo.

aliceinwire:
The (previously referred) Portage Tree as been recently re-defined as the Gentoo repository. 
for don't make confusion with the Portage package manager.

### DIFF
--- a/ebuild-maintenance/text.xml
+++ b/ebuild-maintenance/text.xml
@@ -15,14 +15,14 @@ developers may not be familiar with.
 <body>
 
 <subsection>
-<title>What (not) to put in the Gentoo repository</title>
+<title>What (not) to put in the gentoo repository</title>
 <body>
 
 <p>
 Before writing a new ebuild, check
 <uri link="https://bugs.gentoo.org/">bugs.gentoo.org</uri>
 to see if an ebuild has already been written for the package, but has not yet
-been added to the Gentoo repository.  Go to <uri
+been added to the gentoo repository.  Go to <uri
 link="https://bugs.gentoo.org/">bugs.gentoo.org</uri>, choose query and select
 Advanced Search; as product select <e>Gentoo Linux</e>, as component select
 <e>ebuilds</e>.  In the search field put the name of the ebuild and as status
@@ -31,7 +31,7 @@ select all possible fields, then submit the query. For you lazy people, click
 </p>
 
 <p>
-In general, the Gentoo repository should only be used for storing
+In general, the gentoo repository should only be used for storing
 <path>.ebuild</path> files, as well as any relatively small companion
 files, such as patches or sample configuration files.  These types of
 files should be placed in the <path>/usr/portage/mycat/mypkg/files</path>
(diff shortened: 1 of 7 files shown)


wimmuskee:
Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=594258

gktrk:
I've been meaning to make a post on gentoo-dev about this. Some use Portage tree, some say gentoo-x86, Gentoo ebuild repository, Gentoo package repository, git repository, git tree etc.

aliceinwire:
@gktrk https://archives.gentoo.org/gentoo-dev/message/d791d9351c4fb2a8ecbd83882695631a

gktrk:
Thanks for the link, I was just about to post to -dev. I think I'm ok with these changes. We also use `tree` and `repository` interchangeably. I think that's fine as well.

@ulm any objections?

ulm:
>>>>> On Tue, 15 Nov 2016, Gokturk Yuksek wrote:

> Thanks for the link, I was just about to post to -dev. I think I'm
> ok with these changes. We also use tree and repository
> interchangeably. I think that's fine as well.

> @ulm any objections?

The rsync module is still called "gentoo-portage" which is also the
default in make.conf:
https://gitweb.gentoo.org/proj/portage.git/tree/cnf/make.conf.example?id=fb25c5d908c22bdd6868c7db44c56ef5cd459fe4#n207

But yes, go for "gentoo repository", since we have to start somewhere.
(Upper or lower case, though? Technically, the repository's name is
"gentoo".)

aliceinwire:
@ulm ah you are right, maybe is better lower caps. I thought that we usually write Portage all caps, so I did also with Gentoo. I think is more grammatically correct to write the initial of a name all caps, but also lower caps is ok for me.

aliceinwire:
@ulm replaced Gentoo repository with gentoo repository

kensington:
If we're going with lowercase, how about some formatting to make it clear that "gentoo" is the repository name - unless one is familiar with that particular detail it just looks like we have bad grammar.

Compare:

> In general, the gentoo repository should only be used for storing <path>.ebuild</path> files

and

> In general, the ```gentoo``` repository should only be used for storing <path>.ebuild</path> files

aliceinwire:
@kensington In that case I prefer the uppercase version Gentoo instead of ```gentoo```
IMHO

gktrk:
&lt;c&gt;gentoo&lt;/c&gt; will lose its attention value if we use it everywhere throughout the manual I think

aliceinwire:
@gktrk yes, I think so

gktrk:
IMO, "gentoo repository" == the Gentoo Repository. We should be able to use them interchangeably.

aliceinwire:
@gktrk  :+1:

aliceinwire:
@gktrk can we accept this ? or I need to send a new pull request with uppercase ?

ulm:
We could also define the term once (in general-concepts?) and use "Gentoo repository" otherwise.

aliceinwire:
resending pull request with "Gentoo repository"

PR #55, opened 2016-11-14
function-reference/query-functions: add has_version flag explanation, bug #307065

wimmuskee:
Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=307065

Added functionality as described in the bugreport, as well as adding a full example.

### DIFF
--- a/function-reference/query-functions/text.xml
+++ b/function-reference/query-functions/text.xml
@@ -105,9 +105,9 @@ query variables and similar state.
       <c>has_version pkg[flag]</c>
     </ti>
     <ti>
-      True if <c>pkg</c> (can include version specifiers) is installed. The use of <c>flag</c> is
-      optional and <c>flag</c> can be a list of multiple USE flags. Example:
-      <c>if has_version "=app-editors/nano-2.5.3"["nls,spell"] ; then</c>.
+      True if <c>pkg</c> (can include version specifiers and 
+      <uri link="::general-concepts/dependencies#built with use dependencies"/>) is installed. 
+      Example: <c>if has_version "=app-editors/nano-2.5.3[nls,spell]" ; then</c>.
     </ti>
   </tr>
 </table>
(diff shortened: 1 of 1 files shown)


function-reference/query-functions/text.xml
```diff
@@ -102,10 +102,12 @@ query variables and similar state.
   </tr>
   <tr>
     <ti>
-      <c>has_version pkg</c>
+      <c>has_version pkg[flag]</c>
     </ti>
     <ti>
-      True if <c>pkg</c> (can include version specifiers) is installed.
+      True if <c>pkg</c> (can include version specifiers) is installed. The use of <c>flag</c> is
+      optional and <c>flag</c> can be a list of multiple USE flags. Example:
```
gktrk: `list of multiple USE flags` can be simplified to `list of USE flags`

function-reference/query-functions/text.xml
```diff
@@ -102,10 +102,12 @@ query variables and similar state.
...
+      <c>if has_version "=app-editors/nano-2.5.3"["nls,spell"] ; then</c>.
```
gktrk: the entire `pkg[flag]` can be contained in `""`

function-reference/query-functions/text.xml
gktrk: hmm, now that I think about it again, the proper name for those is [built with use dependencies](https://devmanual.gentoo.org/general-concepts/dependencies/index.html#built-with-use-dependencies). Maybe we can say something like: `True if <c>pkg</c> (can include version specifiers and built with use dependencies) is installed. Example:` ....

function-reference/query-functions/text.xml
wimmuskee: would you like me to do an uri refer to the "built with use dependencies" page?

wimmuskee:
would you like me to do an uri refer to the "built with use dependencies" page?

wimmuskee:
updated with regard to your comments, feel free to squash to one commit
and thx for the comments

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/commit/9101c681314f17ff94abc075afc1d80c42742383. Thanks!

PR #56, opened 2016-11-17
function-reference/install-functions: improve difference between doexe and dobin

wimmuskee:
Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=586632

Tried to make the explanations as concise as possible (and not verbatim copy the PMS texts).

### DIFF
--- a/function-reference/install-functions/text.xml
+++ b/function-reference/install-functions/text.xml
@@ -99,8 +99,8 @@ the first is the source name, the second the name to use when installing.
       <c>dobin</c>
     </ti>
     <ti>
-      Install a binary into <c>/usr/bin</c>, set it to mode 0755 and
-      set ownership to the superuser
+      Install a binary into <c>/usr/bin</c>, set the file mode to 0755
+      and set the ownership to superuser
     </ti>
   </tr>
   <tr>
--- a/function-reference/install-functions/text.xml
+++ b/function-reference/install-functions/text.xml
@@ -99,7 +99,8 @@ the first is the source name, the second the name to use when installing.
       <c>dobin</c>
     </ti>
     <ti>
-      Install a binary
+      Install a binary into <c>/usr/bin</c>, set it to mode 0755 and
+      set ownership to the superuser
     </ti>
   </tr>
   <tr>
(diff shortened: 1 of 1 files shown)


function-reference/install-functions/text.xml
```diff
@@ -99,7 +99,8 @@ the first is the source name, the second the name to use when installing.
       <c>dobin</c>
     </ti>
     <ti>
-      Install a binary
+      Install a binary into <c>/usr/bin</c>, set it to mode 0755 and
```
gktrk: s/it/its/ or s/it to mode/the file mode to/

function-reference/install-functions/text.xml
```diff
@@ -99,7 +99,8 @@ the first is the source name, the second the name to use when installing.
...
+      set ownership to the superuser
```
gktrk: `set the ownership to superuser` is more correct I think.

ulm:
This seems to be one of the places where PMS provides a more
comprehensive explanation, and even comes with examples. Here is what
PMS says about dobin:

   "Gives the files mode 0755 and transfers file ownership to the
   superuser or its equivalent on the system or installation at hand.
   For instance on Gentoo Linux in a non-offset-prefix installation
   this ownership is root:root, while on an offset-prefix aware
   installation this may be joe:users."

wimmuskee:
@ulm, you are disagreeing with gktrk's proposed changes? or maybe even just provide a link to the PMS, and not bother with a bit more context in the devmanual

ulm:
>>>>> On Mon, 21 Nov 2016, Wim Muskee wrote:

> @ulm, you are disagreeing with gktrk's proposed changes?

I was just noting that "superuser" isn't accurate in all situations.

> or maybe even just provide a link to the PMS, and not bother with a
> bit more context in the devmanual

IMHO the devmanual should be stand-alone, at least for such topics
like ebuild functions.

gktrk:
I think saying `superuser` as opposed to `the superuser` covers `or its equivalent on the system or installation at hand.` We don't really document prefix much in devmanual anyway unfortunately.

gktrk:
On a second though, I think I'm gonna append `or its equivalent on the system or installation at hand.` anyway

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/9486c65...7aaea52. Thanks!

PR #57, opened 2016-11-18
Portage is the package manager. The repository is named gentoo.

aliceinwire:
The (previously referred) Portage Tree as been recently re-defined as the Gentoo repository.
for don't make confusion with the Portage package manager.
Changed Portage Tree with Gentoo Repository.

Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=594258

### DIFF
--- a/ebuild-maintenance/text.xml
+++ b/ebuild-maintenance/text.xml
@@ -15,14 +15,14 @@ developers may not be familiar with.
 <body>
 
 <subsection>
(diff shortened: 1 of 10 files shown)


gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/9663bc6...9aebef2. Thanks!

PR #58, opened 2016-11-19
ebuild-maintenance: update fedora repository link, #581608

wimmuskee:
Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=581608


gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/9aebef2...b7e1c71. Thanks!

PR #59, opened 2016-12-03
Make naming scheme more defined in regards to tracking upstream naming

kentfredric:
The real point to get here is people use upstream names in API's far
more than they do gentoo ones, so we should model gentoo names after
upstream identifiers.

I type "firefox &" more than I type "emerge www-client/firefox"
I type "use Moose" more than I type "emerge dev-perl/Moose"

maintainers also need to know identifiers by guesswork, and so that
guesswork should be reliable, so it should track the same case they're
already using in their code their packaging.

@mgorny

### DIFF
--- a/ebuild-writing/file-format/text.xml
+++ b/ebuild-writing/file-format/text.xml
@@ -19,9 +19,39 @@ An ebuild should be named in the form <c>name-version.ebuild</c>.
 </p>
 
 <p>
-The name section should contain only lowercase non-accented letters, the digits
-0-9, hyphens, underscores and plus characters. Uppercase characters are strongly
-discouraged, but technically valid.
+The name section should mimic the identifier most frequently used by the user in order
+to run or consume the package.
+</p>
+
+<p>
+For example, <c>www-client/firefox</c>, <c>dev-lang/perl</c> and <c>sys-devel/gcc</c>
+all make sense to be written as lowercase, as they all ship executables in
+<c>/usr/bin</c>, which the user also calls them in lowercase.
+</p>
+
+<p>
+<c>media-libs/jpeg</c>, <c>media-libs/libpng</c> and <c>dev-libs/expat</c> all make sense
+to be written in lowercase, as they're all primarily referred to with <c>gcc -ljpeg</c>
+etc.
+</p>
+
+<p>
+For some programming languages, like Perl and Haskell, the primary interface is however
+case sensitive where the user is accustomed to regularly using a specific case in their
+code. A library that users primarily consume by <c>use Foo::Bar</c> ( and might be called
+<c>Foo-Bar</c> as an upstream tarball ) would be in a more expected place if it was
+instead called <c>dev-perl/Foo-Bar</c>
+</p>
+
+<p>
+Legal characters include:
+<ul>
+  <li>Upper and lower case non-accented letters</li>
+  <li>the digits 0-9</li>
+  <li>hyphens</li>
+  <li>underscores</li>
+  <li>plus characters</li>
+</ul>
 </p>
 
 <note>


ebuild-writing/file-format/text.xml
```diff
@@ -19,9 +19,30 @@ An ebuild should be named in the form <c>name-version.ebuild</c>.
 </p>
 
 <p>
-The name section should contain only lowercase non-accented letters, the digits
-0-9, hyphens, underscores and plus characters. Uppercase characters are strongly
-discouraged, but technically valid.
+The name section should mimic the identifier most frequently used by the user in a
```
mgorny: Me nitpicking but I think you're putting too much information in a single sentence. One sentence for allowed chars, one sentence about what should it resemble. Also 'non-Gentoo' context sounds weird… maybe just 'in context to the package' or something like that. It's obvious that the name isn't supposed to resemble itself ;-P.

ulm:
I am strictly opposed against this. There was no consensus in gentoo-dev about such a change of policy.

Also the proposed patch would replace a clear cut rule by something rather vague. "Most frequently used by the user" is an ill-defined criterium.

kentfredric:
It wasn't *that* clear, we had to have an -ml discussion about it :)

I read the current rule as "Don't use uppercase (unless you want to)"

And then you look at actual tree and go "Oh .... maybe its more than just no, maybe there have to be good reasons".

Either way, the tree does not reflect the documentation.

It seems more sensible to update the documentation and the policy to reflect the real world accepted usage.

>  "Most frequently used by the user" is an ill-defined criterium.

It is however more specific than "whatever upstream calls themselves"

Its "Whatever upstream calls itself in the primary API it provides"

ie: "/usr/bin/firefox" is the API for the users to "use firefox"
"/usr/lib64/libjpeg.so.1" is the API for tools to "use libjpeg"
"Foo::Bar" is the API perl modules use to "use Foo::Bar"

etc.

The point being that upstream is typically arbitrarily restricted in what they call themselves
at the "load layer", because it has to serve as a unique identifier in "some space", and that "space"
has to have well defined conventions.

And it just so happens that binaries and libraries all typically use a lower case convention because of their lineage from the CLI, where CLI terms were predominantly lower case.

So a vast supply of atoms will have naturally lower case letters for the same reason we want them to have :) 

I am of course open to any better wording suggestions.

ulm:
Github isn't the right place to push through policy changes. Get consensus in the mailing lists first, before writing a patch for the devmanual.

That said, "the identifier most frequently used by the user" isn't suitable as a Gentoo package name. Otherwise we would have sys-apps/cat or sys-apps/ls instead of sys-apps/coreutils, and sys-apps/emerge instead of sys-apps/portage. Also I refuse to call a package e.g. app-doc/PMS (because certainly PMS is used more often than pms to refer to it).

The current policy that uppercase characters are strongly discouraged has served us well for decades, but your patch throws it all overboard. There is no more mention at all that lowercase is the default, and uppercase to be used only in special and exceptional circumstances.

kentfredric:
> That said, "the identifier most frequently used by the user" isn't suitable as a Gentoo package name. 
> Otherwise we would have sys-apps/cat or sys-apps/ls instead of sys-apps/coreutils, and 
> sys-apps/emerge instead of sys-apps/portage. Also I refuse to call a package e.g. app-doc/PMS > (because certainly PMS is used more often than pms to refer to it).

Right. This is obviously not the case with packages like that that are toolboxes of parts.

However, I said *mimic*, not *be*

There are no "primary" pieces in coreutils that the user "interfaces with" that share the same letters.

There is only: 
- its name as used in .info pages
- its name as used in /usr/share/locale

Here, I'd say "these are good proxies to use as a reference for the name".

However, even *if* those files where not present, my guide is to *mimic* those names where it makes sense, 

But obviously you have to use judgement here, and I guess, preferring lowercase if its unclear.

( coreutils itself consists of lots of bin/ files people directly use, and they're all lower-case named, so coreutils itself should follow that precedent )

ulm:
>>>>> On Sat, 03 Dec 2016, Kent Fredric wrote:

> There is only:

>   * its name as used in .info pages
>   * its name as used in /usr/share/locale

> Here, I'd say "these are good proxies to use as a reference for the name".

Not at all. The Info page happens to use both "coreutils" and
"Coreutils". But even if both the Info page and the locale file would
use uppercase exclusively, all-lowercase coreutils would still be more
appropriate as a name.

Also, any change of policy that doesn't reflect current usage (and
the wording of your patch doesn't) would require hundreds of package
moves, with very little benefit.

kentfredric:
> Not at all. The Info page happens to use both "coreutils" and "Coreutils".

In this case:
1. The file is still called 'coreutils.info' 
2. info CoreUtils works in any case, and if "any case is fine", I guess preferring lower-case makes sense there as a default.

The point with literal binaries is they're not case-insensitive, neither are libraries, neither are Perl Modules.

> Also, any change of policy that doesn't reflect current usage (and the wording of your patch doesn't) would require hundreds of package moves, with very little benefit.

Right, the point is not to mandate change, but allow it given right grounds.

Its hard to say "its extremely discouraged" when we have such significant use of it, most of it being represented in "ecosystem groups"

Obviously Perl is well known, but 
```
      2 dev-libs/D
      2 dev-python/B
      2 dev-python/S
      2 media-libs/S
      2 x11-plugins/wmM
      3 dev-libs/O
      3 dev-python/C
      3 x11-plugins/wmS
      4 dev-php/S
     12 dev-python/P
     34 x11-libs/libX
    145 dev-php/P
```

And those x11-libs are following my suggested interpretation.

orlitzky:
> 145 dev-php/P

Most of these suffer from greater brain-damage, since "PEAR" is where the files are hosted, but is not part of those packages' names.

Regardless, I find this interesting: I'm guilty of using upper-case in a few places. And while I couldn't articulate it, it's for the reasons Kent gave. For example, dev-php/PHPMailer is so-named because the user is going to do something like `require 'PHPMailer'` to use it. That simplifies the ebuild and our instructions.

But, on the other hand, I do see a lot of value in just picking a standard (e.g. lowercase) and sticking with it. Not least because every time I add a package, I have this argument with myself, and the results are inconsistent.

gktrk:
I don't understand the problem we are trying to solve here. The language in devmanual is meant to make people think twice about it. After all, you cite many packages with uppercase letters in their names with justifications, which is the intent here. Doesn't seem like the specific wording has stopped anyone and it has worked for us so far.

kentfredric:
@orlitzky the other rule of thumb I guess is if you find yourself needing to set S= and MY_PN= etc a *lot* for related packages, it may make sense not to do that.

ulm:
This is stalled since 30 months, and there was no consensus in the discussion. Closing.

PR #60, opened 2016-12-04
ebuild-writing/variables: add HTML_DOCS, #591476

wimmuskee:
Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=591476

adding a description for HTML_DOCS below the existing DOCS description

### DIFF
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -280,9 +280,8 @@ The following variables may or must be defined by every ebuild.
     <ti><c>HTML_DOCS</c></ti>
     <ti>
     An array or space-delimited list of documentation files for
-    the default src_install function to install using dodoc into
-    the <c>html/</c> subdirectory of docdir (Requires
-    <uri link="::ebuild-writing/eapi/#eapi=6">EAPI>=6</uri>.)
+    the <c>einstalldocs</c> function to install using dodoc -r. 
+    (Requires <uri link="::ebuild-writing/eapi/#eapi=6">EAPI>=6</uri>.)
     </ti>
   </tr>
 </table>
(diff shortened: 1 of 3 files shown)


ebuild-writing/variables/text.xml
```diff
@@ -276,6 +276,15 @@ The following variables may or must be defined by every ebuild.
     <uri link="::ebuild-writing/eapi/#eapi=4">EAPI>=4</uri>.)
     </ti>
   </tr>
+  <tr>
+    <ti><c>HTML_DOCS</c></ti>
+    <ti>
+    An array or space-delimited list of documentation files for
+    the default src_install function to install using dodoc into
```
gktrk: It's technically the `einstalldocs` function, not `src_install`. the fact that `src_install` calls `einstalldocs` is irrelevant.

We should mention that it's done recursively per `dodoc -r`.

I also find the text a bit hard to follow. I'm not sure if the developers need to care where the files are installed, they just need to know it's for the html documentation files.

wimmuskee: einstalldocs is not mentioned (yet) in the devmanual, before I refer to that one, should I add that one to the install functions part? if yes, where? alphabetically after dosym?

dodoc -r, yes, I can add that.

And indeed, if developers are depending on the einstalldocs functionality, they don't need to care where the files are installed. I'll change that as well.

Depending on the answer for the einstalldocs, I will make a new commit.

gktrk: We can insert `einstall` after `dodoc`. I don't think the list is alphabetically sorted.

wimmuskee:
ok, had some trouble with the Github commit yesterday ([similar](http://stackoverflow.com/questions/35052210/unexpected-system-error-after-push-was-received/35052290)), had to push an empty commit to get the actual commit visible here.

Made the requested changes, and also an addition to the src_install page, showing the new einstalldocs in the default EAPI6 version of that function.

gktrk:
And I'll rebase these into 3 commits if you don't mind

wimmuskee:
of course rebase as you please

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/fe224c4...4552670 with modifications. Thanks!

PR #61, opened 2016-12-10
ebuild-writing/metadata: indentation rules, #591548

wimmuskee:
Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=591548

more or less copied the seemingly agreed upon minimal rules
left out the valid XML part, this seems obvious, described in the GLEP, or not relevant to indentation

### DIFF
--- a/ebuild-writing/misc-files/metadata/text.xml
+++ b/ebuild-writing/misc-files/metadata/text.xml
@@ -18,6 +18,12 @@ A metadata file follows the syntax defined in
 <uri link="https://wiki.gentoo.org/wiki/GLEP:68">GLEP 68</uri>.
 The character set <b>must</b> be UTF-8 as specified by
 <uri link="https://wiki.gentoo.org/wiki/GLEP:31">GLEP 31</uri>.
+Concerning indentation there is no strict rule governing the style and width.
+However the following minimal set of rules need to be followed:
+<ul>
+  <li>Indentation should be consistent, i.e. either spaces or tabs, but not both.</li>
+  <li>Keep the existing style when touching another developer's file.</li>
+</ul>
 </p>
 
 <p>


ebuild-writing/misc-files/metadata/text.xml
```diff
@@ -18,6 +18,12 @@ A metadata file follows the syntax defined in
 <uri link="https://wiki.gentoo.org/wiki/GLEP:68">GLEP 68</uri>.
 The character set <b>must</b> be UTF-8 as specified by
 <uri link="https://wiki.gentoo.org/wiki/GLEP:31">GLEP 31</uri>.
+Concerning indentation there is no strict rule governing the style and width.
```
gktrk: Starting this on a new paragraph looks better.

ebuild-writing/misc-files/metadata/text.xml
```diff
@@ -18,6 +18,12 @@ A metadata file follows the syntax defined in
...
+However the following minimal set of rules need to be followed:
+<ul>
+  <li>Indentation should be consistent, i.e. either spaces or tabs, but not both.</li>
+  <li>Keep the existing style when touching another developer's file.</li>
```
gktrk: `Keep the existing style when touching metadata.xml files that belong to other developers.` reads better for me.

wimmuskee: "belong to" -> "are made by"
might be even better

gktrk: Well, if I see an orphaned package, I'm re-indenting its metadata.xml when claiming it ;)

gktrk: My point is, it is more of a concern when the package is owned by a developer which may not be the same person who "made" the package.

wimmuskee: If that is the case, then "belong to" is indeed the better choice.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/0c32c62...49fb72b. Thanks!

PR #62, opened 2017-01-21
shortened to facts and made it more objective

jonasstein:
The user who reads this does not need 3x please. Well sorted and readable attachments are good for everyone. Gentoo-devs, bug reporter, proxy-maintainer, packagers from other distributions looking for solutions...

### DIFF
--- a/ebuild-writing/common-mistakes/text.xml
+++ b/ebuild-writing/common-mistakes/text.xml
@@ -440,7 +440,7 @@ link="::ebuild-maintenance#adding-a-new-ebuild">Adding new Ebuild</uri> tutorial
 <body>
 
 <p>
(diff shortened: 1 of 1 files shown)


EvaSDK:
+1 for just one please but you imho loose the point about not using tarballs making it less cumbersome to review the actual contribution.

Maybe something like "Please do not attach ebuilds or patches as tarballs. It avoids extra operations when reviewing." ?

Also this would make integration with reviewing tools such as found in gnome bugzilla unusable but that point is speculation only for our own infrastructure for the moment.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/ef00b42...1ad3d23. Thanks!

I've squashed the commits, edited the commit message to have a path prefix and reflow the text before committing, but the content hasn't been modified.

PR #64, opened 2017-01-24
Update example to EAPI=6, 2017

jonasstein:


ebuild-writing/eapi/text.xml
```diff
@@ -42,11 +42,11 @@ Most developers prefer to set the EAPI version without quotes. However, the PMS
...
 
 <codesample lang="ebuild">
-# Copyright 1999-2016 Gentoo Foundation
+# Copyright 1999-2017 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 # &#36;Id&#36;
 
-EAPI=5
+EAPI=6
 inherit eutils
```
gktrk: I think we should also drop the inherit from there

orlitzky:
I think the `inherit` was there to show that the EAPI line always comes first, but seeing an `EAPI=5` -> `EAPI=6` bump that retains `eutils.eclass` does make me twitch...

gktrk:
I think it's a poor place to show that inherit comes after EAPI. We also hope to nuke this page altogether in the future. The PMS Appendix E contains almost the exact same thing.

gktrk:
Anyway, merged in https://github.com/gentoo/devmanual.gentoo.org/compare/e098b2f...f4c6229.

I squashed the commits and mangled the commit message to prefix with the path of the directory. I also deleted the empty line right after `EAPI=6`

PR #65, opened 2017-05-07
Improve src_test documentation

mgorny:
As noted in gentoo/gentoo#4544, we have some old stuff there.

### DIFF
--- a/ebuild-writing/functions/src_test/text.xml
+++ b/ebuild-writing/functions/src_test/text.xml
@@ -31,23 +31,13 @@
 <section>
 <title>Default <c>src_test</c></title>
 <body>
-<p> From sys-apps/portage-2.1.2.9 </p>
+<p>The default test phase in EAPI 6 is equivalent to the following:</p>
 <codesample lang="ebuild">
 src_test() {
-	if emake -j1 check -n &amp;> /dev/null; then
-		vecho ">>> Test phase [check]: ${CATEGORY}/${PF}"
-		if ! emake -j1 check; then
-			hasq test $FEATURES &amp;&amp; die "Make check failed. See above for details."
-			hasq test $FEATURES || eerror "Make check failed. See above for details."
-		fi
-	elif emake -j1 test -n &amp;> /dev/null; then
-		vecho ">>> Test phase [test]: ${CATEGORY}/${PF}"
-		if ! emake -j1 test; then
-			hasq test $FEATURES &amp;&amp; die "Make test failed. See above for details."
-			hasq test $FEATURES || eerror "Make test failed. See above for details."
-		fi
-	else
-		vecho ">>> Test phase [none]: ${CATEGORY}/${PF}"
+	if emake check -n &amp;> /dev/null; then
+		emake check
+	elif emake test -n &amp;> /dev/null; then
+		emake test
 	fi
 }
 </codesample>
--- a/ebuild-writing/functions/src_test/text.xml
+++ b/ebuild-writing/functions/src_test/text.xml
@@ -107,19 +107,8 @@ suite is extremely helpful for arch maintainers.
 <title>Skipping Tests</title>
 <body>
 <p>
-Sometimes it is necessary to skip tests entirely. This can be done
-using a dummy <c>src_test</c> function:
-</p>
-
-<codesample lang="ebuild">
-src_test() {
-    # Tests don't even remotely work inside portage
-    true
-}
-</codesample>
-
-<p>
-Another option would be to set <c>RESTRICT="test"</c> in the ebuild.
+Sometimes it is necessary to skip tests entirely. This can be done by
+setting <c>RESTRICT="test"</c> in the ebuild.
 </p>
 
 <note>
--- a/ebuild-writing/functions/src_test/text.xml
+++ b/ebuild-writing/functions/src_test/text.xml
@@ -113,9 +113,9 @@ setting <c>RESTRICT="test"</c> in the ebuild.
 
 <note>
 If upstream provide a test suite which doesn't work, consider talking
-to them about getting it fixed. A broken test suite is worse than no
-test suite at all, since we are unable to tell whether a test failure
-indicates a genuine fault.
+to them about getting it fixed. A broken test suite requires developers
+to investigate each test failure in order to determine whether it is
+genuine or indicates a broken test.
 </note>
 </body>
 </section>
--- a/ebuild-writing/functions/src_test/text.xml
+++ b/ebuild-writing/functions/src_test/text.xml
@@ -59,10 +59,10 @@ src_test() {
 <body>
 <codesample lang="ebuild">
 src_test() {
-    cd "${S}"/src/testdir
+    cd "${S}"/src/testdir || die
 
     # Test 49 won't work inside a portage environment
-    sed -i -e 's~test49.out~~g' Makefile
+    sed -i -e 's~test49.out~~g' Makefile || die
 
     # Try to run the non-gui tests only
     make test-nongui \
--- a/ebuild-writing/functions/src_test/text.xml
+++ b/ebuild-writing/functions/src_test/text.xml
@@ -82,7 +82,6 @@ portage environment. Reasons for such a failure could include:
 </p>
 
 <ul>
-  <li>Needing to use X.</li>
   <li>
     Needing to work with files which are disallowed by the sandbox.
   </li>
--- a/ebuild-writing/functions/src_test/text.xml
+++ b/ebuild-writing/functions/src_test/text.xml
@@ -63,6 +63,107 @@ src_test() {
 </section>
 
 <section>
+<title>Tests that require network or service access</title>
+<body>
+<p>
+Sometimes test suites (and other build-time programs) attempt to use
+remote or local network, or production servers running on the host. All
+of these are strictly forbidden. Developers should either fix such tests
+to work in an isolated environment, or disable them completely unless
+explicitly allowed by the user. At the bare minimum, the tests must
+not fail with <c>FEATURES=network-sandbox</c> being enabled.
+</p>
+
+<p>
+Internet access within the build procedure is forbidden for
+the following reasons:
+</p>
+<ul>
+  <li>
+    the build may be running in an environment with no or restricted
+    Internet access, and this must not cause the tests (build) to fail;
+  </li>
+
+  <li>
+    the Internet connection may be unstable (e.g. poor reception)
+    in which case an interrupted connection or packet loss must not
+    cause the tests to fail or hang, and it should not cause unnecessary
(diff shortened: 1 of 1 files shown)


ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,13 +59,13 @@ src_test() {
...
+    cd "${S}"/src/testdir || die
 
     # Test 49 won't work inside a portage environment
-    sed -i -e 's~test49.out~~g' Makefile
+    sed -i -e 's~test49.out~~g' Makefile || die
 
-    # Try to run the non-gui tests only
-    make test-nongui \
+    # Try to run the non-gui tests only, non-parallel
+    emake -j1 test-nongui \
```
gktrk: If you switch to `emake`, do you still need `|| die` there?

ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,13 +59,13 @@ src_test() {
...
-    cd "${S}"/src/testdir
+    cd "${S}"/src/testdir || die
 
     # Test 49 won't work inside a portage environment
-    sed -i -e 's~test49.out~~g' Makefile
+    sed -i -e 's~test49.out~~g' Makefile || die
 
-    # Try to run the non-gui tests only
-    make test-nongui \
+    # Try to run the non-gui tests only, non-parallel
```
gktrk: Should we rather say `for non-parallelism`? I'm guessing this is related to the hunk about parallelism you deleted below.

ebuild-writing/functions/src_test/text.xml
```diff
@@ -82,7 +82,6 @@ portage environment. Reasons for such a failure could include:
 </p>
 
 <ul>
-  <li>Needing to use X.</li>
```
gktrk: Is this related to the comment `patched out the tests that don't work without X.` in PR 4544? Instead of removing this sentence, should we expand it with this?

```
Reasons for such a failure could include:
- Needing to use X (Unless the test system can be configured/patched to exclude those certain tests).
```

ebuild-writing/functions/src_test/text.xml
```diff
@@ -96,12 +95,6 @@ using <c>sed</c> or skipping a particular <c>make</c> target is
 sufficient.
 </p>
 
-<note>
-<c>emake</c> should not be used for <c>src_test</c> <d/> trying to
-parallelise tests unless the <c>Makefile</c> was specifically designed
-for this can cause all sorts of strange problems.
-</note>
-
```
gktrk: I like the direction of saying `use emake -j1` instead of `do not use emake`. I wonder if we should update this sentence and make it compliant with the example above or simple remove it. Not sure personally.

ebuild-writing/functions/src_test/text.xml
```diff
@@ -114,26 +107,15 @@ suite is extremely helpful for arch maintainers.
...
 </p>
 
 <note>
 If upstream provide a test suite which doesn't work, consider talking
-to them about getting it fixed. A broken test suite is worse than no
-test suite at all, since we are unable to tell whether a test failure
-indicates a genuine fault.
+to them about getting it fixed. A broken test suite requires developers
+to investigate each test failure in order to determine whether it is
+genuine or indicates a broken test.
```
gktrk: I personally like the old version better. Likewise a broken test suite, a non-broken test suite also requires developers to investigate each test failure since there's no way `to tell whether a test failure indicates a genuine fault.`

ebuild-writing/functions/src_test/text.xml
mgorny: Sounds like a weird word. It's just not run in parallel.

ebuild-writing/functions/src_test/text.xml
mgorny: Depends on EAPI, obviously.

ebuild-writing/functions/src_test/text.xml
mgorny: Tests needing to use X should be run via virtualx.eclass.

ebuild-writing/functions/src_test/text.xml
mgorny: I don't think there's a point in making a special verbose notes about past state not applying anymore.

ebuild-writing/functions/src_test/text.xml
mgorny: If test suite is not broken, then a failure indicates a genuine problem, obviously.

The old version stated falsely that 'we are unable to tell [...]' — we *are* able to tell that, it just requires additional effort.

ebuild-writing/functions/src_test/text.xml
kentfredric: > Try to run the non-gui tests only, single threaded

?

gktrk: That works too

ebuild-writing/functions/src_test/text.xml
gktrk: Let's assume EAPI5+

ebuild-writing/functions/src_test/text.xml
gktrk: Why not put that information in there instead?

mgorny: Feel free to do that. Last I checked, I wasn't on devmanual team or responsible for fixing horribly bad documentation.

mgorny:
Updated the non-parallel comment and removed die from emake.

mgorny:
@gentoo/devmanual, please review now. Updated the obsolete snippet from Portage to a modern version, and added detailed section on network/service access and on dealing with X11 (which is basically a specific example of the generic local service problem).

kensington:
These new changes are really, really nice. Good work!

ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,14 +49,161 @@ src_test() {
...
+forbidden for the following reasons:
+</p>
+<ul>
+  <li>
+    the tests must run reliably independently of whether a particular
+    server is running throughout the build process or not,
+  </li>
+
+  <li>
+    using production services for running tests is frequently
```
gktrk: s/frequently/extremely/

ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,14 +49,161 @@ src_test() {
...
+    <b>dangerous</b> as it may inadvertently expose bugs in those
+    services, causing instability, data loss or even exposing security
+    vulnerabilities.
+  </li>
+</ul>
+
+<p>
+Fixing tests that require access to the local services is usually done
```
gktrk: s/the local services/local services/

ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,14 +49,161 @@ src_test() {
...
+via starting additional isolated instances of those services during
+the test phase. Those services must either be running on an UNIX
+socket or on the loopback interface, to reliably prevent remote access.
+</p>
+
+<p>
+For all networked services exposed during the test phase (either by
+the ebuild or the tests itself), UNIX sockets are strongly preferred
```
gktrk: `tests themselves` seems more correct to me

ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,14 +49,161 @@ src_test() {
...
+over IP sockets as they provide better means for unique naming
+and access control mechanisms. The IP sockets can be subject to port
+collisions with other local services and they can be connected to
```
gktrk: s/connected to/accessed/

ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,14 +49,161 @@ src_test() {
...
+by local system users who may exploit a vulnerability through the tests.
+</p>
+
+<p>
+Additional protection against those issues is provided through
+<c>FEATURES=network-sandbox</c>. However, this is only an optional
+Portage feature relying on specific Linux kernel namespace mechanisms
+and the developers should not rely on it being enabled.
```
gktrk: s/the developers/developers/

ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,14 +49,161 @@ src_test() {
...
+Some packages include tests (or other build-time applications) that
+attempt to use the user's X11 session and fail being unable to connect
+to it. Those tests need to be fixed to work independently of the X11
+server that might or might not be running when packages are being built.
+</p>
+
+<p>
+If the program in question does not strictly need X11 but merely
+attempts to take opportunity of the <c>DISPLAY</c> variable being set,
+the best solution is to simply ensure to unset this variable
```
gktrk: Either `ensure that this variable is unset` or `simply unset this variable`

ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,14 +49,161 @@ src_test() {
...
+
+	default
+}
+</codesample>
+
+<p>
+If the package actually requires a running X11 server to run
+the complete test suite, you can use the <c>virtualx</c> eclass to
+provide an isolated Xvfb environment for the tests to use. It provides
+a virtual X11 display that is not connected to any physical devices
```
gktrk: s/devices/device/

ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,14 +49,161 @@ src_test() {
...
+  </li>
+
+  <li>
+    the remote network services used by the tests may become unavailable
+    temporarily or permanently, causing unexpected test failures;
+  </li>
+
+  <li>
+    accessing remote sites always poses a privacy issue, and possibly
+    even a threat to security (e.g. through inadvertently exposing
```
gktrk: s/even//

ebuild-writing/functions/src_test/text.xml
```diff
@@ -59,14 +49,161 @@ src_test() {
...
+    information about the system).
+  </li>
+</ul>
+
+<p>
+Fixing tests that require Internet access usually requires cooperation
+with upstream, and porting the tests to use one of the test techniques
+such as mocking or using replay data. For this reason, developers
+more frequently report the issue upstream and skip tests that require
```
mgorny: This changes the meaning of the sentence completely.

mgorny:
Updated.

gktrk:
http://dev.gentoo.org/~gokturk/devmanual/pr65/ebuild-writing/functions/src_test/index.html to check how it renders.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/9456029...1118c72. Thanks!

I've mangled the summary line prefix before merging.

PR #67, opened 2017-08-06
Improve documentation on revisions and revision bumps

mgorny:
Here's a refurbished version of the doc to help people figure out better to revbump or not to revbump.

The doc ended up a bit list-ish but I think it's going to improve the readability in the long run. The structure is pretty much:
1. what revision is and what is its purpose,
2. rule of thumb when to revbump,
3. example of when to revbump or not.

If you can think of more examples, I think it'd be useful to add them. The thing is always a little tricky for everyone, so the more samples, the less likely someone will do things wrong ;-).

### DIFF
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -62,13 +62,50 @@ of thumb could be used as a guideline:
   </li>
 </ol>
 
-<p>
-Simple compile fixes do <b>not</b> warrant a revision bump; this is because they do
-not affect the installed package for users who already managed to compile it.
-Small documentation fixes are also usually not grounds for a new revision.
-In particular, this applies if the package has a substantial compilation
-time; developers should use their best judgement in these circumstances.
-</p>
+<p>Examples of changes that warrant a new revision are:</p>
+<ul>
+  <li>adding a patch to fix a runtime issue,</li>
+  <li>installing additional files that could be useful to the user,</li>
+  <li>adding a missing runtime dependency to one of the existing blocks,</li>
+  <li>
+    adding a missing build-time dependency that contributed to
+    a successfully yet incorrect build (e.g. missing some feature),
+  </li>
+  <li>
+    restricting a runtime dependency version, unless the <c>:=</c>
+    subslot operator is going to trigger a rebuild.
+  </li>
+</ul>
+
+<p>Examples of changes that can be done without a revision bump are:</p>
+<ul>
+  <li>
+    adding a patch to fix a build-time issue that prevented users from
+    building the package (since it does not affect users who already
+    managed to build it),
+  </li>
+  <li>adding a trivial documentation fix,</li>
+  <li>
+    installing an additional file of relatively little value (minor
+    documentation, editor syntax file, bash completion) compared
+    to the cost of rebuilding the package (especially if a new bump
+    is expected soon),
+  </li>
+  <li>
+    adding a missing build-time dependency that caused a build failure,
+  </li>
+  <li>
+    adding a new USE flag or removing an existing one (since change
+    in USE flags is going to trigger <c>--changed-use</c> rebuild),
+  </li>
+  <li>
+    a trivial stylistic / ebuild code change (as long as the new code
+    is equivalent to the old code),
+  </li>
+  <li>
+    a dependency change that is a result of a package move (slot move).
+  </li>
+</ul>
 
 </body>
 </chapter>
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -8,8 +8,9 @@
 Ebuilds may have a Gentoo revision number associated with them. This is a
 <c>-rX</c> suffix, where <c>X</c> is an integer <d/> see <uri
 link="::ebuild-writing/file-format#File Naming Rules"/>. This
-component must only be used for Gentoo changes, not upstream releases. By
-default, <c>-r0</c> is implied.
+component must only be used for Gentoo changes, not upstream releases.
+An ebuild with no explicit revision number has the implicit <c>-r0</c>
+revision.
 </p>
 
 <p>
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -28,13 +28,39 @@ revision.
 </ol>
 
 <p>
-Ebuilds should have their <c>-rX</c> incremented whenever a change is made which
-will make a substantial difference to what gets installed by the package <d/> by
-substantial, we generally mean "something for which many users would want to
-upgrade". This is usually for bugfixes.
-For any such revision bump, the new ebuild should be based on the
-previous revision to ensure that fixes aren't dropped accidentally.
+Developers are encouraged to use common sense when determining
+whether to introduce a new <c>-rX</c> revision. The following rule
+of thumb could be used as a guideline:
 </p>
+<ol>
+  <li>
+    If the change can cause the package to be broken to the point
+    of requiring users to revert to the previous version (in the case
+    of packages marked stable, every non-trivial change is classified
+    as such), then a new revision should be introduced and the old one
+    kept. If the package has stable keywords, the new revision should
+    be dropped to <c>~arch</c> (see
+    <uri link="::keywording#Keywording on Upgrades"/>).
+    For any such revision bump, the new ebuild should be based
+    on the previous revision to ensure that fixes aren't dropped
+    accidentally.
+  </li>
+
+  <li>
+    If the change makes a substantial difference to the user who already
+    installed the package (fixes runtime issues, changes installed files,
+    etc.) and it would not be propagated using other means, then
+    the ebuild should be renamed to a new revision. If the package has
+    stable keywords, they should be moved to the new revision without
+    dropping. To commit the ebuild, <c>repoman commit
+    --straight-to-stable</c> option should be used.
+  </li>
+
+  <li>
+    Otherwise, the change can be done in place in the current revision
+    of the ebuild.
+  </li>
+</ol>
 
 <p>
 Simple compile fixes do <b>not</b> warrant a revision bump; this is because they do
@@ -42,17 +68,6 @@ not affect the installed package for users who already managed to compile it.
 Small documentation fixes are also usually not grounds for a new revision.
 In particular, this applies if the package has a substantial compilation
 time; developers should use their best judgement in these circumstances.
-</p>
-
-<important>
-For ebuilds marked stable on at least one arch, only trivial edits can be made
-without a bump (e.g. typo fixes in elog messages). Even simple changes may
-result in a breakage. <b>Modifying stable ebuilds should be avoided.</b>
-</important>
-
-<p>
-When doing a revision bump, the usual rules about dropping to <c>~arch</c> apply.
-See <uri link="::keywording#Keywording on Upgrades"/>.
 </p>
 
 </body>
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -13,6 +13,20 @@ An ebuild with no explicit revision number has the implicit <c>-r0</c>
 revision.
 </p>
 
+<p>Ebuild revisions usually serve two purposes:</p>
+<ol>
+  <li>
+    keeping an older copy of an ebuild around when doing a potentially
+    breaking change, and
+  </li>
+
+  <li>
+    propagating the rebuild of a package when performing a meaningful
+    change that would otherwise go unnoticed by users who have installed
+    the current version already.
+  </li>
+</ol>
+
 <p>
 Ebuilds should have their <c>-rX</c> incremented whenever a change is made which
 will make a substantial difference to what gets installed by the package <d/> by


orlitzky:
I have an old patch for this at https://bugs.gentoo.org/show_bug.cgi?id=565560

Most of this looks good. I think there's one more important exception that can be added -- when you do a package or slot move, you're allowed to edit dependencies in place (to rename the package or slot).

I'm also wary of the `--changed-use` exception, because it's portage-specific.

general-concepts/ebuild-revisions/text.xml
```diff
@@ -54,13 +54,48 @@ of thumb could be used as a guideline:
...
-not affect the installed package for users who already managed to compile it.
-Small documentation fixes are also usually not grounds for a new revision.
-In particular, this applies if the package has a substantial compilation
-time; developers should use their best judgement in these circumstances.
-</p>
+<p>The examples of changes that warrant a new revision are:</p>
+<ul>
+  <li>adding a patch to fix a runtime issue,</li>
+  <li>installing additional files that could be useful to the user,</li>
+  <li>adding a missing runtime dependency to one of the existing blocks,</li>
```
Whissi: I would add the general rule to do a rev bump whenever you make changes to any installed files or changes the set of installed files (remove/add new files). Then add exceptions (like the trivial documentation fix which maybe changes a man file or program's `-h` output).

orlitzky: I agree, it's simpler to just say "always revbump except..." and then list the exceptions. But this page is so outdated, I'll take whatever I can get.

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,94 @@
...
+  </li>
+</ol>
+
+<p>The examples of changes that warrant a new revision are:</p>
+<ul>
+  <li>adding a patch to fix a runtime issue,</li>
+  <li>installing additional files that could be useful to the user,</li>
+  <li>adding a missing runtime dependency to one of the existing blocks,</li>
+  <li>
+    adding a missing build-time dependency that caused the package
```
mrueg: What's the difference to https://github.com/gentoo/devmanual.gentoo.org/pull/67/files#diff-3d4b64fa6c27f002d5d97ebaff9af61eR87

mgorny: Here the package *was built* but the result is incorrect, so it's important that people rebuild it. There the package *could not be built*.

general-concepts/ebuild-revisions/text.xml
mgorny: This is kinda implied by point 2 in the list above it. I might make that more detailed but I'd rather keep 'good' and 'bad' examples separate.

mgorny:
> Most of this looks good. I think there's one more important exception that can be added -- when you do a package or slot move, you're allowed to edit dependencies in place (to rename the package or slot).

Isn't that kinda implied by move logic itself? I dare say this kind of info belongs in the place where moves are explained since they don't follow the usual package logic.

mgorny:
Oh, sorry, now I get what you mean by package move ;-). I thought you meant changing stuff inside the package being moved.

mgorny:
Please look at it now. Changes:
* included some examples to indicate what a user-visible change is,
* tried to clear the wording as to the two kind of build-dep problems,
* added example with package move dep changes.

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,98 @@
...
+    to be built successfully but incorrectly (e.g. miss some feature),
```
SoapGentoo: miss -> missing

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,98 @@
...
+    to the cost of rebuilding the package (especially if a new bump
+    is expected soon),
+  </li>
+  <li>
+    adding a missing build-time dependency that caused the package to
+    fail to build,
+  </li>
+  <li>
+    adding a new USE flag or removing an existing one (since change
+    in USE flags is going to trigger <c>--changed-use</c> rebuild),
```
SoapGentoo: this implicitly requires that the great majority of users consider `--changed-use` a "default" flag that is in general always used. I know we've debated this at length, but it would be nice to make this assumption more explicit somewhere

orlitzky: It also requires that the `--changed-use` behavior be _available_ in the user's package manager, and it isn't in the PMS. This probably isn't the place to get into this, but I would rather see us do revbumps for USE flag changes than try to work around it on the user's end. There are two viable scenarios:

**Scenario 1**: Developers make new revisions whenever they change USE flags. All users immediately see the changes, and no special package manager support is needed.

**Scenario 2**: We document that it's OK for developers to change USE flags without a new revision. We insist that all package managers implement flags like `--newuse` and `--changed-use`. Moreover, we insist that users enable those flags when performing upgrades.

The outcomes of both scenarios are identical, but the second one is a hell of a lot more complicated. We would be better off requiring new revisions for USE changes, and then getting rid of the "required" `--newuse` and `--changed-use` that are littered everywhere.

SoapGentoo: @mgorny will disagree here. `--changed-use` will only affect that users where the new flag would've flipped from its "default" implicit state before, that is, `--changed-use` is guaranteed not to cause more rebuilds than revbumps and will very likely lead to less rebuilds.

orlitzky: If you're going to add a flag whose default setting maintains the status quo, you can just wait until the next version or the next time you need to do a revision bump for some other reason before you mess with the flag. That can be used to avoid rebuilds in extreme cases like libreoffice and chromium. Most other packages, it doesn't really matter what you do, because quick rebuilds don't bother anyone.

mgorny: If you want to change the status quo, then please take this to the mailing list and convince others. I'm merely documenting what's the widely-accepted standard these days.

Besides:
* all package managers implement some kind of option for this (though in paludis it's more like 'do not rebuild when options change'),
* expecting people to delay changes is unrealistic for most of multi-developer workflows.

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,98 @@
...
+    adding a missing build-time dependency that caused the package
+    to be built successfully but incorrectly (e.g. missing some feature),
+  </li>
+  <li>
+    restricting a runtime dependency version, unless the <c>:=</c>
+    subslot operator is going to trigger a rebuild.
+  </li>
+</ul>
+
+<p>The examples of changes that can be done without a revison bump are:</p>
```
orlitzky: Typo: revison

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,98 @@
...
+of thumb could be used as a guideline:
 </p>
+<ol>
+  <li>
+    If the change is likely to break the package and require users to
+    downgrade (especially if the package is stable, all non-trivial
+    changes are considered as such), then a new revision should be
+    introduced and the old one kept. If the package has stable keywords,
+    the new revision should be dropped to <c>~arch</c>
+    (see <uri link="::keywording#Keywording on Upgrades"/>).
```
gktrk: It seems as though the importance of treating stable with extra caution does not come out as strong compared to the old text. I would really like to drive home the point of "if you have the slightest doubt about a change to a stable package, revbump` here somehow.

I'm also opposed to the language of "a change is likely to break the package and require users to downgrade". In such situations, the developer should be doing more testing, not pushing it to unstable. I understand that you didn't mean it that way either. How about something like `if a developer is uncertain whether a change may break the package despite their best testing efforts and think that users are likely to benefit from having a fallback after an upgrade`

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,98 @@
...
-For ebuilds marked stable on at least one arch, only trivial edits can be made
-without a bump (e.g. typo fixes in elog messages). Even simple changes may
-result in a breakage. <b>Modifying stable ebuilds should be avoided.</b>
-</important>
+  <li>
+    If the change makes a substantial difference to the user who already
+    installed the package (fixes runtime issues, changes installed files,
+    etc.) and it would not be propagated using other means, then
+    the ebuild should be renamed to a new revision. If the package has
+    stable keywords, they should be kept intact.
```
gktrk: Are there any other means to propagate such things? If not, let's not even say that. Also, renaming an ebuild vs. bump&remove is still a controversial topic and I'd rather stick with the old rule of bump&remove until we discuss it on gentoo-dev. So instead of saying that it should be renamed, let's say it should be revision bumped and the old problematic revision should be removed with the exception of ebuilds with stable keywords.

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,98 @@
...
-When doing a revision bump, the usual rules about dropping to <c>~arch</c> apply.
-See <uri link="::keywording#Keywording on Upgrades"/>.
-</p>
+  <li>
+    Otherwise, the change can be done in place in the current revision
+    of the ebuild.
+  </li>
+</ol>
+
+<p>The examples of changes that warrant a new revision are:</p>
```
gktrk: Don't need `The` here

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,98 @@
...
+<ul>
+  <li>adding a patch to fix a runtime issue,</li>
+  <li>installing additional files that could be useful to the user,</li>
+  <li>adding a missing runtime dependency to one of the existing blocks,</li>
```
gktrk: What do you mean by `existing blocks`?

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,98 @@
...
+  <li>
+    adding a missing build-time dependency that caused the package
+    to be built successfully but incorrectly (e.g. missing some feature),
```
gktrk: Alternative suggestion: `adding a missing build-time dependency that contributed to a successful yet incorrect build (e.g. missing features)

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,98 @@
...
+<p>The examples of changes that can be done without a revision bump are:</p>
+<ul>
+  <li>
+    adding a patch to fix a build-time issue that prevented users from
+    building the package (since it does not affect users who already
+    managed to build it),
+  </li>
+  <li>adding a trivial documentation fix,</li>
+  <li>
+    installing an addition file of relatively little value (minor
```
gktrk: s/addition/additional/

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,98 @@
...
+    documentation, editor syntax file, bash completion) compared
+    to the cost of rebuilding the package (especially if a new bump
+    is expected soon),
+  </li>
+  <li>
+    adding a missing build-time dependency that caused the package to
```
gktrk: s/the package to fail to build/a build failure/

general-concepts/ebuild-revisions/text.xml
gktrk: I am also against this change unless it's in the PMS or there's a council decision or a gentoo-dev thread where there's a consensus.

general-concepts/ebuild-revisions/text.xml
mgorny: Not as strong? It says exactly the same thing — *all non-trivial changes*.

Ok, I get that the 'likely' is not the best word. However, your version makes me feel sick. I'll try to reword it without going pre-school.

general-concepts/ebuild-revisions/text.xml
mgorny: Yes, there are. USE flags and slot operator rebuilds, for example.

The argument about bump/remove is nonsense here. It's not a version bump but a trivial change such as a dependency adjustment. Doing it in two commits only causes confusion. It's a single logical change, and the revbump is only an implied necessity, and not a meaningful step of its own.

general-concepts/ebuild-revisions/text.xml
mgorny: Done.

general-concepts/ebuild-revisions/text.xml
mgorny: It mean you don't add them along with new USE flags, so people with existing set of flags are unaffected.

general-concepts/ebuild-revisions/text.xml
mgorny: Done.

general-concepts/ebuild-revisions/text.xml
mgorny: Done.

general-concepts/ebuild-revisions/text.xml
mgorny: Done.

general-concepts/ebuild-revisions/text.xml
mgorny: There is a reasonable consensus on the gentoo-dev thread. From what I've established on IRC, Michael is not going to require an explicit QA vote on overriding it.

general-concepts/ebuild-revisions/text.xml
```diff
@@ -28,13 +28,34 @@ revision.
 </ol>
 
 <p>
-Ebuilds should have their <c>-rX</c> incremented whenever a change is made which
-will make a substantial difference to what gets installed by the package <d/> by
-substantial, we generally mean "something for which many users would want to
-upgrade". This is usually for bugfixes.
-For any such revision bump, the new ebuild should be based on the
-previous revision to ensure that fixes aren't dropped accidentally.
+The developers are encouraged to use common sense when determining
```
ulm: s/The developers/Developers/

mgorny: Fixed.

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,99 @@
...
-For ebuilds marked stable on at least one arch, only trivial edits can be made
-without a bump (e.g. typo fixes in elog messages). Even simple changes may
-result in a breakage. <b>Modifying stable ebuilds should be avoided.</b>
-</important>
+  <li>
+    If the change makes a substantial difference to the user who already
+    installed the package (fixes runtime issues, changes installed files,
+    etc.) and it would not be propagated using other means, then
+    the ebuild should be renamed to a new revision. If the package has
+    stable keywords, they should be kept intact.
```
gktrk: When you say `they should be kept intact`, are you suggesting committing straight to stable as part of the rename? Or did you want to say `it should be kept intact`, meaning that the stable ebuild should be left alone and a revbump should be made with all ~arch. Also, `package` has stable keywords, or `ebuild` has stable keywords?

mgorny: The keywords should be kept intact, i.e. straight to stable.

Ebuild is a kind of package. In this context, they're synonymous since other kind of packages don't apply here.

mgorny:
Ping.

general-concepts/ebuild-revisions/text.xml
aballier: This is not gonna fly I'm afraid:

```
$ git mv ffmpeg-3.3.3{,-r1}.ebuild
$ repoman
[...]
  KEYWORDS.stable [fatal]       1
   media-video/ffmpeg/ffmpeg-3.3.3-r1.ebuild added with stable keywords: amd64 arm ia64 x86
```

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,99 @@
...
-substantial, we generally mean "something for which many users would want to
-upgrade". This is usually for bugfixes.
-For any such revision bump, the new ebuild should be based on the
-previous revision to ensure that fixes aren't dropped accidentally.
-</p>
+<p>Ebuild revisions usually serve two purposes:</p>
+<ol>
+  <li>
+    keeping an older copy of an ebuild around when doing a potentially
+    breaking change, and
```
aballier: Having `</li> <li>` in the middle of a sentence seems weird to me

general-concepts/ebuild-revisions/text.xml
mgorny: There's `-S`/`--straight-to-stable` option for exactly that purpose, for over two years now.

aballier: it's probably worth mentioning it then; I didn't know about it and would have used --force

general-concepts/ebuild-revisions/text.xml
```diff
@@ -8,37 +8,99 @@
...
+revision.
 </p>
 
-<p>
-Ebuilds should have their <c>-rX</c> incremented whenever a change is made which
-will make a substantial difference to what gets installed by the package <d/> by
-substantial, we generally mean "something for which many users would want to
-upgrade". This is usually for bugfixes.
-For any such revision bump, the new ebuild should be based on the
-previous revision to ensure that fixes aren't dropped accidentally.
```
ulm: IIRC we had introduced this phrase ("the new ebuild should be based on the previous revision to ensure that fixes aren't dropped accidentally") because there were some issues with ebuilds merged from overlays that were dropping previous changes. So could this sentence be kept please? Just above the examples might be a good place.

mgorny:
Reintroduced the sentence about reusing the old ebuild in the first point, and added a tip about `repoman commit --straight-to-stable` to the second point.

mgorny:
@gktrk, ping.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/f3e6cc0...253075d. Thanks!

PR #68, opened 2017-08-07
variables: Advise against proactive use of constant-value variables

mgorny:
As I've suggested on the ml and on [the bug](https://bugs.gentoo.org/show_bug.cgi?id=627170).

### DIFF
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -588,5 +588,53 @@ when joining paths. For example: <c>${D%/}/</c>, <c>${ED%/}/</c>,
 </body>
 </section>
 
+<section>
+<title>Use of constant-value variables in ebuilds</title>
+<body>
+
+<p>
+Variables have significant value in ebuilds, making it possible to avoid
+unnecessary repetitions and make maintenance easier.  However,
+references to constant-value variables should be used with care as their
+excessive use can harm readability and increase maintenance burden (e.g.
+when renaming a package). In particular, using variables whose values
+have no direct correlation with the expected string should be avoided.
+</p>
+
+<p>The examples of beneficial constant-value variable references are:</p>
+<ul>
+  <li>
+    using <c>${PV}</c> and the variables derived from it (<c>${P}</c>,
+    <c>${MY_P}</c>…) in <c>SRC_URI</c> and <c>S</c> to avoid having
+    to update those variables for every version bump (assuming that
+    <c>${PV}</c> is used to indicate the upstream version);
+  </li>
+
+  <li>
+    using <c>${PF}</c> in <c>--docdir</c> path — this is a canonical
+    Gentoo path that is always required to match <c>${PF}</c>.
+  </li>
+</ul>
+
+<p>The examples of bad constant-value variable references are:</p>
+<ul>
+  <li>
+    using <c>${PN}</c> in <c>HOMEPAGE</c>, <c>EGIT_REPO_URI</c>
+    or the domain of <c>SRC_URI</c> — it breaks URL parsing in editors
+    and terminals, and makes it hard to copy the link for external use
+    (e.g. while reviewing via gitweb),
+  </li>
+
+  <li>
+    using <c>${PN}</c> (or other unnecessary helper variables)
+    excessively in <c>src_install()</c> — it impairs readability for
+    little benefit and causes a lot of trouble when the package needs to
+    be renamed.
+  </li>
+</ul>
+
+</body>
+</section>
+
 </chapter>
 </guide>


ebuild-writing/variables/text.xml
```diff
@@ -588,5 +588,54 @@ when joining paths. For example: <c>${D%/}/</c>, <c>${ED%/}/</c>,
...
+<section>
+<title>Use of constant-value variables in ebuilds</title>
+<body>
+
+<p>
+Variables frequently have significant value in ebuilds, making it
+possible to avoid unnecessary repetitions and make maintenance easier.
+However, references to constant-value variables should be used with care
+as their proactive use can harm readability and increase maintenance
+burden (e.g. when renaming the package). In particular, using variables
```
kensington: How does using these variables increase the maintenance burden? Using the example of renaming a package, if I use ``${PN}`` everywhere isn't the maintenance burden reduced?

ebuild-writing/variables/text.xml
```diff
@@ -588,5 +588,54 @@ when joining paths. For example: <c>${D%/}/</c>, <c>${ED%/}/</c>,
...
+whose values are not strictly guaranteed to match the expected string
+should be avoided.
```
kensington: I've read this sentence a few times and I'm not sure I understand. Could you please clarify?

ebuild-writing/variables/text.xml
mgorny: No because now you have to figure out which `${PN}` need to be replaced by the old name (i.e. can't change the value) and which need to be left as `${PN}` (i.e. are supposed to switch to the new name). When people use paths like `/usr/share/${PN}` (where the original path comes from upstream Makefile), this is PITA.

ebuild-writing/variables/text.xml
mgorny: The idea is that it's fine to use a variable if it's 100% the same thing — i.e. you use PV to indicate upstream version, so it's fine to use PV when you mean upstream version. However, PN is Gentoo package name, so it's not fine to randomly replace upstream paths which are not in ebuild control with PN. Can you think of a better way of saying that?

ebuild-writing/variables/text.xml
```diff
@@ -588,5 +588,54 @@ when joining paths. For example: <c>${D%/}/</c>, <c>${ED%/}/</c>,
...
+    the domain or path prefix part of <c>SRC_URI</c> — it breaks URL
+    parsing in editors and terminals, and makes it hard to copy the link
+    for external use (e.g. while reviewing via gitweb),
+  </li>
+
+  <li>
+    using <c>${PN}</c> (or other unnecessary helper variables)
+    proactively in <c>src_install()</c> — it impairs readability for
+    little benefit and causes a lot of trouble when the package needs to
+    be renamed.
```
gktrk: Conversely, I could claim that it's helpful when a rename follows an upstream rename. I don't see what's wrong with something like `doman ${PN}.8` in `src_install()`. If upstream moves from `foo` to `bar`, that `doman` will transition flawlessly. I admit that I don't know how common such mistakes are during package moves given how rare they are.

ebuild-writing/variables/text.xml
```diff
@@ -588,5 +588,54 @@ when joining paths. For example: <c>${D%/}/</c>, <c>${ED%/}/</c>,
...
+  </li>
+</ul>
+
+<p>The examples of bad constant-value variable references are:</p>
+<ul>
+  <li>
+    using <c>${PN}</c> in <c>HOMEPAGE</c>, <c>EGIT_REPO_URI</c>,
+    the domain or path prefix part of <c>SRC_URI</c> — it breaks URL
+    parsing in editors and terminals, and makes it hard to copy the link
+    for external use (e.g. while reviewing via gitweb),
```
gktrk: There are 11264 ebuilds that contain `${PN}` in `SRC_URI`. Unless we convince people to fix them first, there's little value in putting this in devmanual.

Per https://devmanual.gentoo.org/ebuild-writing/variables/index.html, developers are advised to not use variables inside `HOMEPAGE`, so that's  covered.

ebuild-writing/variables/text.xml
```diff
@@ -588,5 +588,54 @@ when joining paths. For example: <c>${D%/}/</c>, <c>${ED%/}/</c>,
...
+  <li>
+    using <c>${PV}</c> and the variables derived from it (<c>${P}</c>,
+    <c>${MY_P}</c>…) in <c>SRC_URI</c> and <c>S</c> to avoid having
+    to update those variables for every version bump (presuming that
+    <c>${PV}</c> is used to indicate the upstream version);
+  </li>
+
+  <li>
+    using <c>${PF}</c> in <c>--docdir</c> path — this is a canonical
+    Gentoo path that is always required to match <c>${PF}</c>.
```
gktrk: This is a good one and very easily missed. Maybe this will work better under the Common Mistakes section.

ebuild-writing/variables/text.xml
floppym: There's no need to "fix" existing ebuilds before setting a guideline. The 11k ebuilds can be updated during normal maintenance.

ebuild-writing/variables/text.xml
mgorny: Those are two separate problems. Here I'm just telling it's fine to use. The formal requirement for using it belongs elsewhere.

ebuild-writing/variables/text.xml
mgorny: https://devmanual.gentoo.org/ebuild-writing/variables/index.html is exactly the page that is being altered by this pull request…

ebuild-writing/variables/text.xml
mgorny: No, it won't transition because the old version will still be named `foo`.

ebuild-writing/variables/text.xml
gktrk: That presumes that people agree to it. I agree that it's incorrect to use `${PN}` when you actually mean the literal string that it evaluates to. The argument is that if you rename a package from `foo` to `bar` with the upstream staying constant, your `${PN}` in SRC_URI will break things. In that case, however, `${P}` will also change. So if you had `${P}.tar.gz` in SRC_URI, that will also break. If you had `-> ${P}.tar.gz`, that will require a re-run of `repoman manifest`.

The sole argument that it makes copy-paste harder mostly applies to users submitting pull requests as opposed to developers pushing their ebuilds directly to the repo. That, in turn, can be enforced by proxy-maint directly.

gktrk:
Overall, I'm sympathetic to the changes. If we ask developers to not use variables in `HOMEPAGE`, we should also be able to ask them to use more reviewer-friendly syntax in `SRC_URI` and `EGIT_REPO_URI`. Like many people say, an ebuild is written once read many times.

However, I'm not in favor of this being some kind of enforceable policy. There are practical use cases. For example, mate-desktop.org eclass has:

```
# Set SRC_URI or EGIT_REPO_URI based on whether live                                                                                                                                  
if [[ ${PV} == 9999 ]]; then
    EGIT_REPO_URI="https://github.com/mate-desktop/${MATE_DESKTOP_ORG_PN}.git"
    SRC_URI=""
else
    SRC_URI="https://pub.mate-desktop.org/releases/${MATE_BRANCH}/${MATE_DESKTOP_ORG_PN}-${MATE_DESKTOP_ORG_PV}.tar.${MATE_TARBALL_SUFFIX}"
fi
```

where

```
# @ECLASS-VARIABLE: MATE_DESKTOP_ORG_PN                                                                                                                                               
# @DESCRIPTION:                                                                                                                                                                       
# Name of the package as hosted on mate-desktop.org.                                                                                                                                  
# Leave unset if package name matches PN.                                                                                                                                             
: ${MATE_DESKTOP_ORG_PN:=$PN}
```

If we go with this change, the eclass would enforce every ebuild to define `MATE_DESKTOP_ORG_PN` manually *without* using a variable. I would think that for the MATE ecosystem, `MATE_DESKTOP_ORG_PN:=$PN` mostly holds and the deduplication is meaningful.

mgorny:
It does not enforce anything. It says 'should be avoided', i.e. discourages. Proxy-maint enforcing a different policy is stupid. What I'm actually trying to do is to actually have a single useful source of documentation for everyone, not have to repeatedly explain the same things again.

ulm:
>>>>> On Wed, 16 Aug 2017, Gokturk Yuksek wrote:

> +<p>The examples of bad constant-value variable references are:</p>
> +<ul>
> +  <li>
> +    using <c>${PN}</c> in <c>HOMEPAGE</c>, <c>EGIT_REPO_URI</c>,
> +    the domain or path prefix part of <c>SRC_URI</c> — it breaks URL
> +    parsing in editors and terminals, and makes it hard to copy the link
> +    for external use (e.g. while reviewing via gitweb),

I am strongly opposed against this change. Especially, SRC_URI
typically contains other variables anyway, so it is not a literal
string that could be copied. Also not allowing PN will cause conflicts
with another policy, namely "try to keep lines no wider than
80 positions", if the package name is long.

(I would agree to avoiding variables in the _host_ part of URLs. OTOH,
I cannot remember having ever seen that in an ebuild.)

Generally this looks like micro-management to me, which we should
avoid. Why would we have variables like PN or CATEGORY if people are
not allowed to used them? (Next time it will be proposed that ${P} and
${PF} should be avoided too, because they can be spelt out as
<packagename>-${PV} and <packagename>-${PVR}.)

> There are 11264 ebuilds that contain ${PN} in SRC_URI. Unless we
> convince people to fix them first, there's little value in putting
> this in devmanual.

Right, and IMHO there is nothing wrong with them. Don't fix what ain't
broken.

Two examples from the app-emacs category:
SRC_URI="https://github.com/${PN}/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
SRC_URI="mirror://hackage/packages/archive/${PN}/${PV}/${P}.tar.gz"

I believe this is perfectly valid use, and it would make no sense to
expand part of the ${PN}/${PV} if it's followed by another ${P}.
Besides, you cannot copy a mirror: URI to your browser anyway.

> Per https://devmanual.gentoo.org/ebuild-writing/variables/index.html,
> developers are advised to not use variables inside HOMEPAGE, so that's
> covered.

Yes, and I think there is consensus not to change that rule.

Ulrich

mgorny:
> I am strongly opposed against this change. Especially, SRC_URI
typically contains other variables anyway, so it is not a literal
string that could be copied. Also not allowing PN will cause conflicts
with another policy, namely "try to keep lines no wider than
80 positions", if the package name is long.

That's a no-argument. Variables and line length are purely coincidental, so you shouldn't use the other as an excuse for the first. Otherwise, you break `ekeyword` by starting to play with `KEYWORDS`.

> (I would agree to avoiding variables in the _host_ part of URLs. OTOH,
I cannot remember having ever seen that in an ebuild.)

I did. I also did see scary things like `http://github.com/${PN}-project/${PN}/...`. Do you think the first one is really a good idea?

> Generally this looks like micro-management to me, which we should
avoid. Why would we have variables like PN or CATEGORY if people are
not allowed to used them? (Next time it will be proposed that ${P} and
${PF} should be avoided too, because they can be spelt out as
<packagename>-${PV} and <packagename>-${PVR}.)

Don't turn this into a drama, please.

ebuild-writing/variables/text.xml
ulm: I disagree about this hunk too. IMHO the general guideline from above ("references to constant-value variables should be used with care") is enough, and we should avoid too much micro-management here, like specific rules for `SRC_URI` or `${PN}`.

mgorny: Examples are not micro-management. Examples tip you what the formal rules actually mean. Would it be fine if I killed the part about `SRC_URI`?

mgorny:
Changed to cover only domain part of SRC_URI.

ebuild-writing/variables/text.xml
```diff
@@ -588,5 +588,54 @@ when joining paths. For example: <c>${D%/}/</c>, <c>${ED%/}/</c>,
 </body>
 </section>
 
+<section>
+<title>Use of constant-value variables in ebuilds</title>
+<body>
+
+<p>
+Variables frequently have significant value in ebuilds, making it
```
gktrk: s/frequently//, it doesn't boost the meaning of the sentence in the way that is intended IMO, frequency relates to time but we talk about the value of variables in general here.

ebuild-writing/variables/text.xml
gktrk: s/the package/a package/

ebuild-writing/variables/text.xml
gktrk: Here's my take on this sentence. I'm not sure if I've made it more complex and simpler: `In particular, using variables whose values have no direct correlation with the expected string other than matching it by coincidence should be avoided.` Maybe following it with a concrete example can clarify it better.

ebuild-writing/variables/text.xml
```diff
@@ -588,5 +588,54 @@ when joining paths. For example: <c>${D%/}/</c>, <c>${ED%/}/</c>,
...
+whose values are not strictly guaranteed to match the expected string
+should be avoided.
+</p>
+
+<p>The examples of beneficial constant-value variable references are:</p>
+<ul>
+  <li>
+    using <c>${PV}</c> and the variables derived from it (<c>${P}</c>,
+    <c>${MY_P}</c>…) in <c>SRC_URI</c> and <c>S</c> to avoid having
+    to update those variables for every version bump (presuming that
```
gktrk: s/presuming/assuming/, https://en.oxforddictionaries.com/usage/assume-or-presume

gktrk:
With that being said, I'm not super picky about `proactive`. I think it's incorrect to use it, I don't think it will make a substantial difference in understanding given the context.

ebuild-writing/variables/text.xml
mgorny: I've took the first part of it but stripped the 'other than ...'. I think it's unnecessary given the context, and the rest sums it up well.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/dd80781...f3e6cc0. Thanks!

PR #69, opened 2017-08-26
install-functions: Verbosely explain dosym target parameter

mgorny:
It is a common misconception that dosym takes two paths, both of them
interpreted relatively to ${ED}. To avoid that, explicitly explain
that the first parameter is interpreted as a verbatim symlink target.

### DIFF
--- a/function-reference/install-functions/text.xml
+++ b/function-reference/install-functions/text.xml
@@ -257,7 +257,11 @@ the first is the source name, the second the name to use when installing.
       <c>dosym</c>
     </ti>
     <ti>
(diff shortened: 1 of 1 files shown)


ulm:
Please use &lt;e&gt; (rather than &lt;b&gt;) for emphasis in running text.

Otherwise LGTM.

mgorny:
Updated.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/7fda20c...dd80781. Thanks!

Please try to use the parent path in the prefix as much as possible in the commit summary line, I missed this one.

PR #70, opened 2017-09-06
Replace nonfatal example with something correct

mgorny:
Replace the nonfatal example from a poor programming practice to a code
collecting logs for test failures. Copied from sys-apps/paludis.

Closes: https://bugs.gentoo.org/551190

// I've attached this to the bug a long time ago, let's try to finally get it merged.

### DIFF
--- a/ebuild-writing/eapi/text.xml
+++ b/ebuild-writing/eapi/text.xml
@@ -369,19 +369,31 @@ src_compile() {
 		</p>
 		<p>Example:</p>
 		<codesample lang="ebuild">
-EAPI=1
+EAPI=2
 ...
-src_install() {
-	emake DESTDIR="${D}" install || die "make install failed"
-	dodoc ChangeLog README
+src_test() {
+	if ! emake check ; then
+		local a
+		eerror "Tests failed. Looking for files for you to add to your bug report..."
+		while IFS='' read -r -d $'\0' a ; do
+			eerror "    ${a}"
+		done &lt; &lt;(find "${S}" -type f '(' -name '*.epicfail' -o -name '*.log' ')' -print0)
+		die "Make check failed"
+	fi
 }
 		</codesample>
 		<codesample lang="ebuild">
 EAPI=4
 ...
-src_install() {
-	emake DESTDIR="${D}" install
-	nonfatal dodoc ChangeLog README
+src_test() {
+	if ! nonfatal emake check ; then
+		local a
+		eerror "Tests failed. Looking for files for you to add to your bug report..."
+		while IFS='' read -r -d $'\0' a ; do
+			eerror "    ${a}"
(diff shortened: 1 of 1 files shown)


orlitzky:
"Make check failed" is probably meaningless to the end-user, personally I would go with `die "test suite failed"`, but otherwise LGTM.

ebuild-writing/eapi/text.xml
```diff
@@ -369,19 +369,29 @@ src_compile() {
...
-EAPI=1
+EAPI=2
 ...
-src_install() {
-	emake DESTDIR="${D}" install || die "make install failed"
-	dodoc ChangeLog README
+src_test() {
+	if ! emake check ; then
+		eerror "Tests failed. Looking for files for you to add to your bug report..."
+		find "${S}" -type f -name '*.epicfail' -or -name '*.log' | while read a ; do
```
ulm: Shouldn't there be parentheses around the two `-name` tests? Otherwise, non-regular `*.log` will match too.
Also, any good reason for using `-or` (which is a GNU extension) instead of POSIX `-o`?

ebuild-writing/eapi/text.xml
```diff
@@ -369,19 +369,29 @@ src_compile() {
...
 		<codesample lang="ebuild">
 EAPI=4
 ...
-src_install() {
-	emake DESTDIR="${D}" install
-	nonfatal dodoc ChangeLog README
+src_test() {
+	if ! nonfatal emake check ; then
+		eerror "Tests failed. Looking for files for you to add to your bug report..."
+		find "${S}" -type f -name '*.epicfail' -or -name '*.log' | while read a ; do
```
ulm: See above.

ebuild-writing/eapi/text.xml
```diff
@@ -369,19 +369,29 @@ src_compile() {
...
 ...
-src_install() {
-	emake DESTDIR="${D}" install || die "make install failed"
-	dodoc ChangeLog README
+src_test() {
+	if ! emake check ; then
+		eerror "Tests failed. Looking for files for you to add to your bug report..."
+		find "${S}" -type f -name '*.epicfail' -or -name '*.log' | while read a ; do
+			eerror "    ${a}"
+		done
```
ulm: The while loop is executed in a subshell, therefore you don't have to declare `a` as a local variable. So the example is correct, but maybe this is a little too subtle?

ebuild-writing/eapi/text.xml
mgorny: It was just a cheap copy-paste, I'll look into improving it.

mgorny:
Updated to a nice, modern loop.

mgorny:
@gktrk, ping.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/253075d...b5f6752. Thanks!

PR #71, opened 2017-09-06
dependencies: Improve the description of blockers

mgorny:
The current text is not very explanatory. I still see a lot of people committing pure DEPEND soft blocks that don't even work! Let's try to improve it:
* clearly define the distinction between soft and hard blocks,
* explain how blocks work (or don't work) in RDEP/DEP.

### DIFF
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -215,6 +215,10 @@ to the package manager.
 </p>
 
 <p>
+The following description applies to all EAPIs starting with EAPI 2.
+</p>
+
+<p>
 There are two kinds of blockers: soft blockers and hard blockers.
 </p>
 
@@ -233,11 +237,17 @@ to solve file collisions between packages and are meaningful only
 in <c>RDEPEND</c>.
 </p>
 
+<warning>
+Soft blockers that are pure <c>DEPEND</c> <e>do not work correctly</e>.
+While Portage seemingly queues the package for removal, it <e>does not</e>
+exempt their contents from file collision checks. Always include your
+soft blockers in <c>RDEPEND</c>!
+</warning>
+
 <p>
 If it is strictly necessary to resolve the blocker before the package
-is built (installed) or if automatic resolution is undesired,
-a hard blocker must be used instead. Hard blockers are not resolved
-automatically, and are expressed using the following syntax:
+is built (installed), a hard blocker must be used instead. Hard blockers
+are expressed using the following syntax:
 </p>
 
 <codesample lang="ebuild">
@@ -251,10 +261,6 @@ is installed (but do not prevent building binary packages). Blockers
 defined purely in <c>DEPEND</c> are enforced only for building
 the package from source, and may not apply once the package is installed
 or when it is installed from a binary package.
-</p>
-
-<p>
-Hard blocks require EAPI 2 or newer.
 </p>
 
 <p>
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -245,6 +245,15 @@ RDEPEND="!!app-misc/foo"
 </codesample>
 
 <p>
+Hard blockers apply accordingly to the dependency type defining them.
+Blockers defined in <c>RDEPEND</c> are enforced as long as the package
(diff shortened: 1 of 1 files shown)


general-concepts/dependencies/text.xml
```diff
@@ -209,28 +209,46 @@ asterisk. Also note that when selecting all versions in a specific
...
 RDEPEND="!!app-misc/foo"
 </codesample>
 
 <p>
-Hardblockers always take precedence over softblockers and need at least EAPI=2.
-Also note that blockers are usually <e>runtime</e> rather than buildtime.
+EAPIs 0 and 1 do not define the distinction between soft and hard
+blocks. In those EAPIs, only the single exclamation mark syntax is
+allowed and it is undefined whether it is interpreted as a soft or hard
+block.
```
ulm: I'd rather not add such long explanations about deprecated EAPIs. The previous wording that at least EAPI 2 is required was just fine.

mgorny: I was actually wondering if we need to mention EAPI at all. But sure, I can go for the old concept.

mgorny: Fixed.

mgorny:
Ping.

general-concepts/dependencies/text.xml
```diff
@@ -209,28 +209,59 @@ asterisk. Also note that when selecting all versions in a specific
...
+While Portage seemingly queues the package for removal, it <e>does not</e>
+exempt their contents from file collision checks. Always include your
+soft blockers in <c>RDEPEND</c>!
+</warning>
+
+<p>
+If it is strictly necessary to resolve the blocker before the package
+is built (installed) or if automatic resolution is undesired,
+a hard blocker must be used instead. Hard blockers are not resolved
+automatically, and are expressed using the following syntax:
```
aballier: Actually, portage is supposed to remove hard blockers before compiling the package. It does not currently and there's a bug about it

general-concepts/dependencies/text.xml
```diff
@@ -209,28 +209,59 @@ asterisk. Also note that when selecting all versions in a specific
...
+Hard blockers apply accordingly to the dependency type defining them.
+Blockers defined in <c>RDEPEND</c> are enforced as long as the package
+is installed (but do not prevent building binary packages). Blockers
+defined purely in <c>DEPEND</c> are enforced only for building
+the package from source, and may not apply once the package is installed
+or when it is installed from a binary package.
+</p>
+
+<p>
+Hard blocks require EAPI 2 or newer.
```
aballier: I think soft blockers also change behavior between EAPIs 0-1 and 2+
(or rather: behave as you describe in EAPI 2+)

mgorny: Not exactly — they are not soft blockers but undefined-kind-of-blockers then. The previous version explained that in detail but @ulm requested that I revert to the previous wording which is this.

general-concepts/dependencies/text.xml
mgorny: The text here is supposed to explain what will realistically happen to users, now what some random bug requests to happen even though the 'bug' will probably never be fixed. PMS leaves this undefined.

general-concepts/dependencies/text.xml
aballier: Why not simply put a note that this section applies to EAPI2+ then ? (because everyone should be using something even newer actually)
I don't see any value in documenting something that has been improved/fixed in EAPI2 for a reason.

general-concepts/dependencies/text.xml
aballier: soft and hard blockers have nothing to do with automatic resolution so it is certainly inaccurate to describe hard blockers as a method for avoiding automatic solving

mgorny: Whatever. Removed it.

general-concepts/dependencies/text.xml
mgorny: Added.

mgorny:
@gktrk, ping.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/b5f6752...d82fac3. Thanks!

PR #72, opened 2017-09-06
ebuild-maintenance: Correct the tags used in footer of commit messages

mgorny:
Replace the self-invented 'Gentoo-Bug' tag with 'Bug' and 'Closes' tags
that are used by Portage and Gentoo Infra, and are compatible with
CGIT (gitweb), GitHub, GitLab, Bitbucket...

Also, indicate how they affect Gentoo Bugzilla and GitHub issues/PRs.

### DIFF
--- a/ebuild-maintenance/text.xml
+++ b/ebuild-maintenance/text.xml
@@ -161,17 +161,21 @@ testing the changes, and revelant bugs. Use RFC822/git style tags as
 explained in the
 <uri link="https://kernel.org/doc/html/latest/process/submitting-patches.html">
 Linux Kernel patch guideline</uri>. Additionally, the following tags
-are used in Gentoo:
+are optionally used in Gentoo:
 
 <ul>
-<li><c>Gentoo-Bug:</c> Use this to reference bugs in Gentoo Bugzilla
-by either the bug ID or the bugzilla URI. Multiple bugs can be
-referenced by inserting more tags of this type or separating bug IDs
-with a comma in the same tag. The bug IDs may include an optional '#'
-prefix. There is no consensus on referencing bugs in the summary line
-versus referencing with a tag in the message body. The developer has
-the option to choose either.
-</li>
+<li><c>Bug:</c> Use this to reference bugs <e>without</e> closing them.
+The value is a URL to the bug. If multiple bugs need to be referenced,
+each one should be listed in a separate <c>Bug</c> tag. If a bug
+on Gentoo Bugzilla, or an issue or a pull request on GitHub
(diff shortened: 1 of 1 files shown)


ebuild-maintenance/text.xml
```diff
@@ -164,14 +164,16 @@ Linux Kernel patch guideline</uri>. Additionally, the following tags
...
-versus referencing with a tag in the message body. The developer has
-the option to choose either.
-</li>
+<li><c>Bug:</c> Use this to reference bugs <e>without</e> closing them.
+The value needs to include a full URL to the bug, and can contain only
+a single URL. If a bug on Gentoo Bugzilla, or an issue or a pull request
+on GitHub is referenced, a reference to the commit will be added
+automatically.</li>
+<li><c>Closes:</c> Use this to reference bugs and close them
+automatically. Alike <c>Bug:</c>, the value needs to include a full URL
```
ulm: "Full URL" may be misleading. We allow abbreviated https://bugs.gentoo.org/123456 but don't require the full https://bugs.gentoo.org/show_bug.cgi?id=123456 form, right?

mgorny: Just 'URL'?

mgorny: Fixed.

mgorny:
gentoo-dev is completely useless. I don't have time to listen to another 6 developers demand their personal style even though for 3+ years they haven't done anything to actually prove its value except for demanding others to do things their way.

I'd say there's no point in keeping them in summary line, given that it does not affect git hooks.

This is merely a statement on how things work on Infra end. Finally something purely technical and not useless bikeshed.

kensington:
I don't think it's necessary to bikeshed the tag names as it's pretty clear you have to use those exact values if you want the hook to do anything. I say that as someone whose preferred tag style does not match how the hook is implemented.

However, it might be nice to clarify that clarify that using any of these tags is optional ('Additionally, the following tags are optionally used in Gentoo:').

I'm also in favour of keeping some text indicating that there's no consensus on referencing bugs.

Without some clarifications like this we're saying you must use these tags and there's no consensus on this (nor are we likely to ever get it about any commit message styles).

mgorny:
Updated the text to make them optional, improved the descriptions a bit and moved the notion on summary line after the tag block.

kensington:
Nice!

mgorny:
Ping.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/d82fac3...790ba94. Thanks!

PR #74, opened 2017-10-14
Git guide

mgorny:
This is still work-in-progress but some early feedback would be appreciated.

[GLEP 66](https://www.gentoo.org/glep/glep-0066.html) only aims to provide a minimal formal specification. [Gentoo git workflow](https://wiki.gentoo.org/wiki/Gentoo_git_workflow) turned into half-reviewed mess.

What I'd like to see is a single *good* official documentation, and I think devmanual is the correct place for it. It already describes some specific details like commit messages. I'd like to add an additional section specific to git, and move those to it while leaving the other stuff as generic maintenance tasks.


mgorny:
That's all I thought of adding. If you have additional suggestions, I'm open to them.

ebuild-maintenance/git/text.xml
```diff
@@ -0,0 +1,322 @@
...
+<title>Preparing a development checkout</title>
+<body>
+
+<subsection>
+<title>Cloning the gentoo.git repository</title>
+<body>
+
+<p>
+In order to commit or submit ebuild changes, you should work on top
+of the development git repository. In order to have push access, you should use
```
gktrk: Avoid the repetitive use of `in order to`. If you assume basic git knowledge, you can take advantage of the lingo and use something like: `The main ebuild development happens on the official git repository. You can push your changes only through the SSH protocol. Therefore, clone the repository using:`

ebuild-maintenance/git/text.xml
```diff
@@ -0,0 +1,322 @@
...
+
+</body>
+</subsection>
+
+<subsection>
+<title>Configuration specific to the Gentoo repository</title>
+<body>
+
+<p>
+To ensure that the Gentoo policies are followed, you should ensure to set
```
gktrk: Similarly, get rid of the repetitive `ensure`

ebuild-maintenance/git/text.xml
```diff
@@ -0,0 +1,322 @@
...
+<body>
+
+<subsection>
+<title>Committing and verifying commits</title>
+<body>
+
+<p>
+The recommended way of committing to the Gentoo repository is to use <c>repoman
+commit</c>. It automatically performs the necessary QA checks on the package
+being committed and has other quirks helping with the Gentoo workflow. However,
```
gktrk: I doubt `quirk` the right word to use here: `an idiosyncrasy; a slight glitch, mannerism; something unusual about the manner or style of something or someone`

ebuild-maintenance/git/text.xml
```diff
@@ -0,0 +1,322 @@
...
+it is currently limited to creating a single commit to a single package.
+</p>
+
+<p>
+For any other use case, <c>git commit</c> and other git commands need to be
+used. The valid uses of git include:
+</p>
+
+<ul>
+<li>creating commits spanning across multiple packages and/or multiple areas
```
gktrk: `Span` is transitive, you don't need the word `across`.

ebuild-maintenance/git/text.xml
```diff
@@ -0,0 +1,322 @@
...
+of Gentoo repository (eclasses, licenses, profiles…),</li>
```
gktrk: missing `the`

ebuild-maintenance/git/text.xml
```diff
@@ -0,0 +1,322 @@
...
+<li>amending a commit created via <c>repoman commit</c> with additional files
+or fixups,</li>
+<li>combining multiple commits created via <c>repoman commit</c> via <c>git
```
gktrk: Avoid the double use of `via` by saying `using git rebase`

ebuild-maintenance/git/text.xml
```diff
@@ -0,0 +1,322 @@
...
+rebase</c>.</li>
+</ul>
+
+<p>
+Whenever repoman is not used to commit, you need to manually verify that all
```
gktrk: s/that//

ebuild-maintenance/git/text.xml
```diff
@@ -0,0 +1,322 @@
...
+</body>
+</subsection>
+
+<subsection>
+<title>Atomic commits</title>
+<body>
+
+<p>
+Whenever possible, use atomic commits. Try to split your changes into logical
+commits, abiding the following three rules as much as possible:
```
gktrk: s/abiding/abiding by/

ebuild-maintenance/git/text.xml
```diff
@@ -0,0 +1,322 @@
...
+ebuilds relying on them. Usually it is also acceptable to include those changes
+along with the commit adding the package.
+</li>
+</ol>
+
+<p>
+Please note that revision bumps count as side effects of the changes requiring
+them and do not belong in separate commits. When doing multiple irrelevant
+changes that require a revision bump, it is only necessary to bump the revision
+in the first commit in the series introduced by a single push.
```
gktrk: I am personally against this. I think each commit should move the tree from one stable state to another and it has been this way so far. A user should be able to pick any commit without requiring the knowledge of the push boundaries and expect a "stable" tree state. Has this been discussed on gentoo-dev or is there a QA team statement?

ebuild-maintenance/git/text.xml
mgorny: Done.

ebuild-maintenance/git/text.xml
mgorny: Done.

ebuild-maintenance/git/text.xml
mgorny: Replaced with 'features'.

ebuild-maintenance/git/text.xml
mgorny: Done.

ebuild-maintenance/git/text.xml
mgorny: Done.

ebuild-maintenance/git/text.xml
mgorny: Done.

ebuild-maintenance/git/text.xml
mgorny: Done.

ebuild-maintenance/git/text.xml
mgorny: Done.

ebuild-maintenance/git/text.xml
mgorny: Yes, this has been debated to death (around one of the GLEP reviews), with the result of *1:n* where *n* was a fair number of people who agreed with this specific policy, and 1 was against it. So that makes 2. And yes, it's part of GLEP 66.

Long story short, revbumps are only a matter of enforced upgrade on `emerge -... @world`. If the user is doing the handiwork of picking specific commits, it is extremely unlikely he will be doing repetitive `@world` upgrades on successive commits — he is usually just testing stuff and reinstalling a specific package.

orlitzky:
I wasn't the only person against committing changes that require a revbump without the revbump =P

I think we should recommend `git config --local diff.renames 1` as well, because it lets you see the diff between the old ebuild and the new one after you do a revbump and make an edit.

orlitzky:
By my count, @rich0 and I were for revbumping each commit, while @mgorny and @floppym were against it. I don't care what you do any more and I don't want a vote or anything, but for the record, n=2 and you were off by one.

floppym:
Rich basically said that we should keep the tree in working condition with each commit, which we all agree with.

There was some later debate as to what that actually means in relation to revbumps. I don't see any message where he conclusively "took a side".

rich0:
If I thought this would benefit from more bike-shedding input I'd be offering it...

Seriously, counting votes is about the worst way to handle something like this.  IMO the folks spearheading this ought to do what they think is best, and if somebody has THAT big of a problem it becomes their job to escalate things.  You can't make progress if those trying to change things have to try to win over some kind of nebulous consensus.  WP:BOLD and all that...

I'm flattered that my opinion seems to mean something though.  :)

ebuild-maintenance/git/text.xml
gktrk: I retract my opposition since GLEP 66 has been officially approved.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/f84cfb4...5121332. Thanks!

PR #76, opened 2018-01-13
Major profile/metadata description refurbishing [wip]

mgorny:
The current documentation on profiles looks horrible. It seems to mix up (or is it just some historical state?) files from inside profiles with top-level profiles/ file, also is hard to read and missing a lot.

So my idea is to split it in two chapters: one dedicated to metadata files (unrelated to profiles), and the other dedicated specifically to profiles. This also means describing some of the Gentoo profile structure eventually.

TODO:
metadata section:
- [x] metadata/layout.conf
- [ ] profiles/arch.list
- [x] profiles/categories
- [ ] profiles/desc/ (integrate with use.desc?)
- [x] profiles/eapi
- [x] profiles/info*
- [ ] profiles/license_groups
- [x] profiles/repo_name
- [ ] profiles/thirdpartymirrors
- [x] profiles/use.desc
  - [ ] it probably needs modernization
- [x] profiles/updates
  - [ ] link it to ebuild maintenance section on moving packages
  - [ ] do we have the warning that you can't use it on existing package somewhere? if not, we should add it

profile section:
- [ ] describe what profiles are
- [ ] profiles.desc (it's stub now)
- [ ] eapi
- [ ] parent
- [x] make.defaults
- [x] packages
  - [ ] verify it
  - [ ] packages.build ?
- [x] package.mask
  - [ ] we seriously need to improve that section
- [x] use.mask
  - [ ] use.force
  - [ ] package.use*
  - [ ] *stable* variants


ulm:
Looking at the general structure, ACK on flattening the profiles chapter. Also ACK on separating the top-level profiles directory from profiles.

However, I think that describing the top-level profiles dir in a chapter named metadata is confusing. IMHO it should rather be called "Profiles Directory" or similar.

Also not sure if having two chapters for this at the top level is appropriate. Maybe leave only the profiles chapter at top level, and place the profiles dir under general-concepts or even general-concepts/tree?

mgorny:
Note there's one file in metadata. Not sure how to fit that.

mgorny:
Abandoning.

PR #78, opened 2018-05-24
Remove einstall reference and add Autotools Mythbuster

Zlogene:
It is banned since EAPI=6, older EAPIs are deprecated. Also one more reference for autotools


gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/de8434b...c4df531. Thanks!

PR #79, opened 2018-06-15
patches: add warning about strip level

Zlogene:
Lots of people ignores it even for really simple things, lets at least warn them.

### DIFF
--- a/ebuild-writing/misc-files/patches/text.xml
+++ b/ebuild-writing/misc-files/patches/text.xml
@@ -37,6 +37,12 @@ here. Mirrored patches are often compressed with <c>xz</c> or
 Patches included in <c>${FILESDIR}</c> should never be compressed.
 </note>
 
+<warning>
+Starting from EAPI=6 a strip patch level was limited to the <c>-p1</c>.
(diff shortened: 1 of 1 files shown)


ebuild-writing/misc-files/patches/text.xml
```diff
@@ -37,6 +37,12 @@ here. Mirrored patches are often compressed with <c>xz</c> or
 Patches included in <c>${FILESDIR}</c> should never be compressed.
 </note>
 
+<warning>
+Starting from EAPI=6 a strip patch level was limited to the <c>-p1</c>.
+Although it can be overridden with a <c>eapply -p&lt;strip_level&gt;</c> command
+it is higly recommended to enforce portage defaults.
```
ulm: This is not a "portage default" in particular.
Maybe "... recommended to adapt the patch itself to work with the -p1 default."?

Zlogene: @ulm, wfm, pushed

ulm:
Merged, with an comma inserted, and the paragraph rewrapped.

PR #80, opened 2018-06-16
ebuild-writing/functions/src_prepare: modernize src_prepare()

Zlogene:
Not really happy with the conditional patching in the sample, but modernize at least

### DIFF
--- a/ebuild-writing/functions/src_prepare/text.xml
+++ b/ebuild-writing/functions/src_prepare/text.xml
@@ -48,8 +48,10 @@ src_prepare() {
 <body>
 <codesample lang="ebuild">
 src_prepare() {
-	epatch "${FILESDIR}/${PV}/${P}-fix-bogosity.patch"
-	use pam &amp;&amp; epatch "${FILESDIR}/${PV}/${P}-pam.patch"
+	eapply "${FILESDIR}/${PV}/${P}-fix-bogosity.patch"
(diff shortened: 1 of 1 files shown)


ebuild-writing/functions/src_prepare/text.xml
```diff
@@ -48,8 +48,9 @@ src_prepare() {
 <body>
 <codesample lang="ebuild">
 src_prepare() {
-	epatch "${FILESDIR}/${PV}/${P}-fix-bogosity.patch"
-	use pam &amp;&amp; epatch "${FILESDIR}/${PV}/${P}-pam.patch"
+	default
```
ulm: IMHO "default" doesn't make much sense unless patches are applied via the PATCHES variable. So I'd rather have an explicit eapply_user after the eapply calls.

ebuild-writing/functions/src_prepare/text.xml
```diff
@@ -48,8 +48,9 @@ src_prepare() {
...
+	eapply "${FILESDIR}/${PV}/${P}-fix-bogosity.patch"
+	use pam &amp;&amp; eapply "${FILESDIR}/${PV}/${P}-pam.patch"
```
ulm: Right, conditional patching is frowned upon. Simply omit the "use pam &&" part?

Zlogene:
@ulm makes sense, fixed

PR #81, opened 2018-06-16
function-reference/install-functions: add ban notes about banned funs in EAPI=7

Zlogene:
Just a note as per PMS

### DIFF
--- a/function-reference/install-functions/text.xml
+++ b/function-reference/install-functions/text.xml
@@ -89,6 +89,7 @@ the first is the source name, the second the name to use when installing.
   <tr>
     <ti>
       <c>libopts</c>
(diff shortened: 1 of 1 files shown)


function-reference/install-functions/text.xml
```diff
@@ -65,6 +65,7 @@ the first is the source name, the second the name to use when installing.
   <tr>
     <ti>
       <c>insopts</c>
+      <b>Note:</b> Banned in EAPI=7
```
ulm: Wrong command. It is libopts that has been banned, not insopts.

Zlogene:
@ulm indeed, fixed

ulm:
Cherry picked to master.

PR #82, opened 2018-06-16
general-concepts/use-flags: deprecate LINGUAS in favour of L10N

Zlogene:
do not know f it is worth to mention about gettext and `LINGUAS` now, anyway, `L10N` is default now, will open yet another PR to add `LINGUAS` if you want me to.


ulm:
Cherry picked to master.

PR #85, opened 2018-06-22
WIP: tools-reference: remove useless head and tail section (and further more)

Zlogene:
Useless and has to die


Zlogene:
To clarify: it was only somehow useful when we had the question about these tools in the end quiz, now it is just redundant for our documentation, it can be easily found just googling

ulm:
Why do you want to remove this section in particular? head and tail are widely used throughout the tree, and on the other hand we also document commands like cut or tr.

Zlogene:
@ulm cut and co. are on my removal list too (however we still have these ones as the quiz questions that recruiters will prune soon as well). I just do not see the point to pull in too generic info in the devmanual

ulm:
Maybe remove all in one go then (after adjusting the quizzes)?
And a technical remark, you must remove the include in ../text.xml too.

ulm:
I've changed my mind in the meantime. Keeping these sections doesn't harm, and there's enough bad bash code out there that having some explanation and a few examples in the devmanual might even add some value.

Also, these tools are basically immutable, therefore very little maintenance is required for them.

ulm:
Closing, per my comment of 2020-01-08.

PR #86, opened 2018-06-22
xsl: highlight eapply calls

Zlogene:
now as per https://devmanual.gentoo.org/ebuild-writing/functions/src_prepare/index.html xsl has no clue about it.


Zlogene:
@ulm ping, too small changes :)

ulm:
Sorry, I haven't found the time yet. If you want to look at it yourself, there is a complete list of package manager and eclass functions in ebuild-mode:
https://gitweb.gentoo.org/proj/ebuild-mode.git/tree/ebuild-mode-keywords.el#n32

I think it should be safe to add at least the package manager keywords (lines 32 to 96).

gktrk:
@ulm how about we add `eapply` now, file a bug for an xsl highlight revision and address the rest later?

PR #87, opened 2018-06-23
ebuild-writing/functions/src_install: remove einstall reference

Zlogene:
round two, the first one was #78

### DIFF
--- a/ebuild-writing/functions/src_install/text.xml
+++ b/ebuild-writing/functions/src_install/text.xml
@@ -93,48 +93,6 @@ src_install() {
 </section>
 
 <section>
-<title>Easy Installs</title>
-<body>
-<p>
-Often, especially with autotools-powered packages, there is a <c>Makefile</c>
-<c>install</c> target which will honour the <c>DESTDIR</c> variable to tell it to
-install to a non-root location. If possible, this should be used:
-</p>
-
-<codesample lang="ebuild">
-	emake DESTDIR="${D}" install
-</codesample>
-
-<note>
-<c>emake</c> should be used to parallelise here. Some installs are
-not designed to be parallelised, use <c>emake -j1</c> or <c>make</c>
-if you hit an error.
-</note>
-
-<p>
-  Sometimes this will end up installing a few things into strange
-  places. If and only if this is the case, the <c>einstall</c>
-  function can be used. It is usually necessary to include additional
-  <c>dodoc</c> statements for the <c>README</c>, <c>ChangeLog</c>, etc
-  in these cases:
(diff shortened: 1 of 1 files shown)


ebuild-writing/functions/src_install/text.xml
```diff
@@ -92,48 +92,6 @@ src_install() {
...
-<codesample lang="ebuild">
-	emake DESTDIR="${D}" install
-</codesample>
-
-<note>
-<c>emake</c> should be used to parallelise here. Some installs are
-not designed to be parallelised, use <c>emake -j1</c> or <c>make</c>
-if you hit an error.
-</note>
-
```
ulm: This is way too intrusive. IMHO the section with the "emake install" example should remain in place.

ebuild-writing/functions/src_install/text.xml
```diff
@@ -92,48 +92,6 @@ src_install() {
...
-<p>
-  Sometimes this will end up installing a few things into strange
-  places. If and only if this is the case, the <c>einstall</c>
-  function can be used. It is usually necessary to include additional
-  <c>dodoc</c> statements for the <c>README</c>, <c>ChangeLog</c>, etc
-  in these cases:
-</p>
```
ulm: This paragraph should indeed be removed. Or maybe better, updated to something like this:

    Usually the package's build system will not install the <c>README</c>,
    <c>ChangeLog</c>, etc. files, so it is necessary to include additional
    <c>dodoc</c> statements for them:

ebuild-writing/functions/src_install/text.xml
```diff
@@ -92,48 +92,6 @@ src_install() {
...
-
-<codesample lang="ebuild">
-	einstall
```
ulm: Leave in place, only update the example to use:

    emake DESTDIR="${D}" install

ebuild-writing/functions/src_install/text.xml
```diff
@@ -92,48 +92,6 @@ src_install() {
...
-	dodoc README CHANGES
-</codesample>
-
-<note>
-There is no need to <c>dodoc</c> <c>COPYING</c>! The license belongs
-to <c>${PORTDIR}/licenses</c>. Sometimes though, you might want to
-install <c>COPYING</c> regardless, if it explains how different
-licenses are applied to different parts of the application, for
-example.
-</note>
```
ulm: Leave this note in place too, it also has nothing to do with einstall.

ulm:
This was resolved less intrusively in #137. Closing.

PR #89, opened 2018-07-18
ebuild-writing/functions/pkg_postrm: sample update

Zlogene:
`fdo-mime.eclass` is dead


gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/5334a50...033e606. Thanks!

PR #91, opened 2018-07-19
keywording: sort keywords out

Zlogene:
As https://github.com/gentoo/portage/commit/f4faf6d7950150a3068f92a3f3615b884db897d6 was introduced  (erhm?), let's not to introduce minorsyn bugs in the devmanual

### DIFF
--- a/keywording/text.xml
+++ b/keywording/text.xml
@@ -29,7 +29,7 @@ A sample <c>KEYWORDS</c> entry might look like:
 </p>
 
 <codesample lang="ebuild">
-KEYWORDS="x86 sparc ~mips ~ppc ~ppc-macos -ia64"
+KEYWORDS="-ia64 ~mips ~ppc sparc x86 ~ppc-macos"
 </codesample>
 
 <p>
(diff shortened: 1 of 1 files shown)


keywording/text.xml
```diff
@@ -29,7 +29,7 @@ A sample <c>KEYWORDS</c> entry might look like:
 </p>
 
 <codesample lang="ebuild">
-KEYWORDS="x86 sparc ~mips ~ppc ~ppc-macos -ia64"
+KEYWORDS="-ia64 ~mips ~ppc sparc x86 ~ppc-macos"
```
gktrk: Shouldn't `ppc-macos` come after `ppc` ?

Zlogene: No, prefix keywords go last

Zlogene: @gktrk As I said, no. Arch teams are using `ekeyword` and it works like this (random package example)

```sh
eb3eadc3b326 /usr/portage/app-editors/moe # ekeyword ~ppc moe-1.9.ebuild
moe-1.9: amd64 arm ~ppc x86 ~amd64-linux ~x86-linux
eb3eadc3b326 /usr/portage/app-editors/moe # ekeyword ~ppc-macos moe-1.9.ebuild
moe-1.9: amd64 arm ~ppc x86 ~amd64-linux ~x86-linux ~ppc-macos
eb3eadc3b326 /usr/portage/app-editors/moe # 
```

PR #93, opened 2019-04-13
Implement search functionality via lunr.js

oxr463:
Bug: https://bugs.gentoo.org/674378
Signed-off-by: Lucas Ramage <ramage.lucas@protonmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,12 +8,17 @@ prereq:
 	@type convert >/dev/null 2>&1 || { echo "media-gfx/imagemagick with corefonts, svg and truetype required" >&2; exit 1; }; \
 		type xsltproc >/dev/null 2>&1 || { echo "dev-libs/libxslt is required" >&2; exit 1; }
 
+index:
+	@echo -n "var documents = " > documents.js
+	@./search_index.py text.xml >> documents.js
+
 %.png : %.svg
 	convert $< $@
 
 clean:
 	@find . -name "*.png" -a \! -path "./icons/*" -exec rm -v {} +
 	@find . -name "index.html" -exec rm -v {} +
+	@rm documents.js
 
 # Given a directory with text.xml in it, return its immediate children as prerequisites
 # Hypothetical example:
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,7 @@ prereq:
 		type xsltproc >/dev/null 2>&1 || { echo "dev-libs/libxslt is required" >&2; exit 1; }
 
 index:
-	@echo -n "var documents = " > documents.js
-	@./search_index.py text.xml >> documents.js
+	@./search_index.py text.xml > documents.js
 
 %.png : %.svg
 	convert $< $@
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ all: prereq $(text_files) $(image_files)
 
 prereq:
 	@type convert >/dev/null 2>&1 || { echo "media-gfx/imagemagick with corefonts, svg and truetype required" >&2; exit 1; }; \
-		type xsltproc >/dev/null 2>&1 || { echo "dev-libs/libxslt is required" >&2; exit 1; }
+		type xsltproc >/dev/null 2>&1 || { echo "dev-libs/libxslt with python is required" >&2; exit 1; }
 
 index:
 	@./search_index.py text.xml > documents.js
--- a/devbook.xsl
+++ b/devbook.xsl
@@ -520,7 +520,45 @@
             </div>
           </div>
         </nav>
+        <nav class="navbar navbar-grey navbar-stick" id="devmanual-actions" role="navigation">
+          <div class="container">
+            <div class="row">
+              <div class="navbar-header">
+                <button type="button" class="navbar-toggle" data-toggle="collapse" data-target="#gw-toolbar">
+                  <span class="sr-only">Toggle navigation</span>
+                  <span class="icon-bar"></span>
+                  <span class="icon-bar"></span>
+                  <span class="icon-bar"></span>
+                </button>
+              </div>
+              <div class="collapse navbar-collapse" id="gw-toolbar">
+              <div class="input-group">
+                <input type="search" name="search" placeholder="Search" title="Search Gentoo Developer Manual [f]" accesskey="f" id="searchInput" class="form-control"/>
+                <div class="input-group-btn">
+                  <input type="submit" name="fulltext" value="Search" title="Search the pages for this text" id="mw-searchButton" class="searchButton btn btn-default" onclick="search()"/>
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+      </nav>
       </header>
+      <div id="searchResults" class="modal fade" role="dialog">
+        <div class="modal-dialog">
+          <div class="modal-content">
+            <div class="modal-header">
+              <button type="button" class="close" data-dismiss="modal">x</button>
+              <h4 class="modal-title">Search Results</h4>
+            </div>
+            <div class="modal-body">
+              <p>No results found.</p>
+            </div>
+            <div class="modal-footer">
+              <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
+            </div>
+          </div>
+        </div>
+      </div>
       <div class="container">
         <div class="row">
           <div class="col-md010">
@@ -568,6 +606,16 @@
       </footer>
       <script src="https://assets.gentoo.org/tyrian/jquery.min.js"></script>
       <script src="https://assets.gentoo.org/tyrian/bootstrap.min.js"></script>
+      <script src="https://unpkg.com/lunr/lunr.js"></script>
+      <xsl:variable name="relative_path_depth" select="string-length(/guide/@self)-string-length(translate(/guide/@self, '/' , ''))"/>
+      <xsl:variable name="relative_path_depth_recursion">
+          <xsl:call-template name="str:repeatString">
+            <xsl:with-param name="count" select="$relative_path_depth"/>
+            <xsl:with-param name="append">../</xsl:with-param>
+          </xsl:call-template>
+      </xsl:variable>
+      <script src="{$relative_path_depth_recursion}documents.js"></script>
+      <script src="{$relative_path_depth_recursion}search.js"></script>
     </body>
     </html>
   </xsl:template>
--- a/devbook.xsl
+++ b/devbook.xsl
@@ -606,7 +606,7 @@
       </footer>
       <script src="https://assets.gentoo.org/tyrian/jquery.min.js"></script>
       <script src="https://assets.gentoo.org/tyrian/bootstrap.min.js"></script>
-      <script src="https://unpkg.com/lunr/lunr.js"></script>
+      <script src="https://assets.gentoo.org/lunr/lunr.min.js"></script>
       <xsl:variable name="relative_path_depth" select="string-length(/guide/@self)-string-length(translate(/guide/@self, '/' , ''))"/>
       <xsl:variable name="relative_path_depth_recursion">
           <xsl:call-template name="str:repeatString">
--- a/search.js
+++ b/search.js
@@ -1,4 +1,7 @@
-// GPL-2.0-or-later
+/*
+ * Copyright 2019 Gentoo Authors
+ * GPL-2.0-or-later
+ */
 "use strict";
 
 var search_index = lunr(function () {
--- a/search.js
+++ b/search.js
@@ -1,6 +1,6 @@
 /*
  * Copyright 2019 Gentoo Authors
- * GPL-2.0-or-later
+ * Distributed under the terms of the GNU GPL version 2 or later
  */
 "use strict";
 
--- a/search.js
+++ b/search.js
@@ -1,0 +1,56 @@
+"use strict";
+
+var search_index = lunr(function () {
+  this.ref('name');
+  this.field('text');
+  this.field('url');
+
+  documents.forEach(function (doc) {
+    this.add(doc);
+  }, this);
+});
+
+var search_input = document.getElementById("searchInput");
+
+search_input.addEventListener("keyup", function(event) {
+  if(event.keyCode === 13) {
+    event.preventDefault();
+    document.getElementById("mw-searchButton").click();
+  }
+});
+
+function getContents(docs, article) {
+  var contents = { text: "", url: "" };
+
+  for (var i = 0; i< docs.length; i++) {
+    if (docs[i].name == article) {
+      contents.text = docs[i].text;
+      contents.url = docs[i].url;
+    }
+  }
+  return contents;
+}
+
+function search() {
+  var term = document.getElementById("searchInput").value;
+  if (term !== "") {
+    var results = search_index.search(term);
+    if (results.length > 0) {
+      $("#searchResults .modal-body").empty();
+      $.each(results, function(index, result) {
+        var title = result.ref;
+        var contents = getContents(documents, title);
+
+        $("#searchResults .modal-body").append(`<article><h5><a href="${contents.url}">
+                                                ${title}</a></h5><p>${contents.text}</p></article>`);
+      });
+    } else {
+      $("#searchResults .modal-body").empty();
+      $("#searchResults .modal-body").append("<p>No results found.</p>");
+    }
+  } else {
+      $("#searchResults .modal-body").empty();
+      $("#searchResults .modal-body").append("<p>No search term defined.</p>");
+  }
+  $("#searchResults").modal();
+}
--- a/search.js
+++ b/search.js
@@ -1,3 +1,4 @@
+// GPL-2.0-or-later
 "use strict";
 
 var search_index = lunr(function () {
--- a/search_index.py
+++ b/search_index.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 # Copyright 2019 Gentoo Authors
-# GPL-2.0-or-later
+# Distributed under the terms of the GNU GPL version 2 or later
 import json
 import os
 import sys
--- a/search_index.py
+++ b/search_index.py
@@ -1,4 +1,5 @@
 #!/usr/bin/python
+# Copyright 2019 Gentoo Authors
 # GPL-2.0-or-later
 import json
 import os
--- a/search_index.py
+++ b/search_index.py
@@ -1,4 +1,5 @@
 #!/usr/bin/python
+# GPL-2.0-or-later
 import json
 import os
 import sys
--- a/search_index.py
+++ b/search_index.py
@@ -6,7 +6,9 @@ import xml.etree.ElementTree as ET
 
 xmlFile = sys.argv[1]
 documents = []
+url_root = 'https://devmanual.gentoo.org/'
 
+print('var documents = ', end = '')
 for path, dirs, files in os.walk('.'):
     if xmlFile in files:
         tree = ET.parse(path + '/' + xmlFile)
@@ -15,7 +17,7 @@ for path, dirs, files in os.walk('.'):
             try:
                 documents.append({"name": chapter.find('title').text,
                     "text": chapter.find('body').find('p').text,
-                    "url": path })
+                                  "url": url_root + path[+2:] })
             except:
                 pass
     if '.git' in dirs:
--- a/search_index.py
+++ b/search_index.py
@@ -1,0 +1,24 @@
+#!/usr/bin/python
+import json
+import os
+import sys
+import xml.etree.ElementTree as ET
+
+xmlFile = sys.argv[1]
+documents = []
+
+for path, dirs, files in os.walk('.'):
+    if xmlFile in files:
+        tree = ET.parse(path + '/' + xmlFile)
+        root = tree.getroot()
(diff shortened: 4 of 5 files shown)


oxr463:
@maffblaster, not sure if you saw this. What do you think?

aliceinwire:
Tested and it works well.
image

oxr463:
> 
> 
> Tested and it works well.
> image
> Acked-by: [alicef@gentoo.org](mailto:alicef@gentoo.org)

Thanks! I still have to resolve the issue above with the relative URLs, but then I think this should be good to go!

aliceinwire:
sending a commit now for that!

aliceinwire:
This is really nice to play :) thanks for this pull request !

aliceinwire:
@gktrk @maffblaster would be nice to have this merged.

Makefile
```diff
@@ -8,12 +8,16 @@ prereq:
 	@type convert >/dev/null 2>&1 || { echo "media-gfx/imagemagick with corefonts, svg and truetype required" >&2; exit 1; }; \
 		type xsltproc >/dev/null 2>&1 || { echo "dev-libs/libxslt is required" >&2; exit 1; }
 
+index:
+	@./search_index.py text.xml > documents.js
```
lu-zero: Do we depend on python (and the deps I'm seeing in the script) to build it already? If not please update the prereq target.

oxr463: Python 2.7 is required by libxslt [already](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/libxslt/libxslt-1.1.33-r1.ebuild#n25), and all of the python dependencies for the script are part of the standard library.

orlitzky:
In the Makefile, the `documents.js` file should be the rule target so that nothing happens if you run "make index" twice. (Right now, it tries to build a file literally named `index`... I think you wanted it to be a "phony" target?). The `documents.js` target should also depend on every `text.xml` file in the repository, so that if one of them changes, the search index will get rebuilt. Finally, the `search_index.py` file is kind of doing double duty here. Half of it is creating the search index (good), but the other half is reimplementing stuff that should be done in the makefile (figuring out which files to add to the index, bad).

The devmanual makefile was a bit of disaster, but I think this can be done elegantly. I have [bug 684688](https://bugs.gentoo.org/684688) open to simplify the makefile, after which I think all of this would be pretty easy to do. I already find all `text.xml` files in the repository, so it would be easy to add them as a dependency of `documents.js`. After that, we could make `search_index.py` just take a list of files on the command line, and pass it the list of `text.xml` files instead of having it search the repo itself.

I'm happy to help if my patch ever gets merged =P

aliceinwire:
> I'm happy to help if my patch ever gets merged =P
Happy to help restructure a better makefile after this is merged.

search.js
```diff
@@ -0,0 +1,56 @@
+"use strict";
```
mgorny: Don't we need license here?

search_index.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/python
```
mgorny: Likewise.

devbook.xsl
```diff
@@ -568,6 +606,16 @@
       </footer>
       <script src="https://assets.gentoo.org/tyrian/jquery.min.js"></script>
       <script src="https://assets.gentoo.org/tyrian/bootstrap.min.js"></script>
+      <script src="https://unpkg.com/lunr/lunr.js"></script>
```
mgorny: Generally relying on third parties here sucks a lot. We should move it to assets instead.

gktrk:
```suggestion
      <script src="https://assets.gentoo.org/lunr/lunr.min.js"></script>
```

gktrk:
@orlitzky since you've called my devmanual Makefile hack a disaster, I defer the help with rebasing to you ;)

aliceinwire:
@gktrk should be better if @orlitzky create a new pull request for rebasing the makefile, instead of stopping others pull request.

devbook.xsl
oxr463: It is assumed that this file will be uploaded to that location, correct?

maffblaster: I agree with this decision. We want to copy any assets we're using on the sites to assets.gentoo.org. This is the www version of distfiles.gentoo.org, which is important in the event that a CDN or external site disappears, we still have functional features on our site(s).

search.js
oxr463: In accordance with the [Gentoo Social Contract](https://www.gentoo.org/get-started/philosophy/social-contract.html), I release these scripts under the terms of the GPL-2.0-or-later.

search_index.py
oxr463: See previous.

devbook.xsl
oxr463: As long as the appropriate party adds this file where it needs to be that is fine with me.

gktrk: Yup, I'm currently working on it.

ulm:
@gktrk See GLEP 76. We need both a copyright notice and a license notice. (An SPDX identifier is no substitute for a license notice.)

gktrk:
> @gktrk See GLEP 76. We need both a copyright notice and a license notice. (An SPDX identifier is no substitute for a license notice.)

So is the following sufficient:

```
/*
 * Copyright 2019 Gentoo Authors
 * GPL-2.0-or-later
 */
```

or do we need the full blown:

```
one line to give the program's name and an idea of what it does.
Copyright (C) 2019 Gentoo Authors

This program is free software; you can redistribute it and/or
modify it under the terms of the GNU General Public License
as published by the Free Software Foundation; either version 2
of the License, or (at your option) any later version.

This program is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU General Public License for more details.

You should have received a copy of the GNU General Public License
along with this program; if not, write to the Free Software
Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
```

devbook.xsl
gktrk: The deployment of lunr.min.js to infra is almost ready: https://github.com/gktrk/devmanual-assets

ulm:
> So is the following sufficient:
> 
> ```
> /*
>  * Copyright 2019 Gentoo Authors
>  * GPL-2.0-or-later
>  */
> ```
> 
> or do we need the full blown:
> 
> ```
> Copyright (C) 2019 Gentoo Authors
> 
> This program is free software; you can redistribute it and/or
> modify it under the terms of the GNU General Public License
> as published by the Free Software Foundation; either version 2
> of the License, or (at your option) any later version.
> 
> This program is distributed in the hope that it will be useful,
> but WITHOUT ANY WARRANTY; without even the implied warranty of
> MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
> GNU General Public License for more details.
> 
> You should have received a copy of the GNU General Public License
> along with this program; if not, write to the Free Software
> Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
> ```

Yes, either the full notice (which is what I would prefer), or at least something like this:
```
Copyright 2019 Gentoo Authors
Distributed under the terms of the GNU GPL version 2 or later
```

See https://gitweb.gentoo.org/proj/eselect.git/tree/ for examples of both (current version of full GPL license notice in `libs/*.in`, short notice in `modules/*.eselect`).

gktrk:
> > So is the following sufficient:
> > ```
> > /*
> >  * Copyright 2019 Gentoo Authors
> >  * GPL-2.0-or-later
> >  */
> > ```
> > 
> > 
> > or do we need the full blown:
> > ```
> > Copyright (C) 2019 Gentoo Authors
> > 
> > This program is free software; you can redistribute it and/or
> > modify it under the terms of the GNU General Public License
> > as published by the Free Software Foundation; either version 2
> > of the License, or (at your option) any later version.
> > 
> > This program is distributed in the hope that it will be useful,
> > but WITHOUT ANY WARRANTY; without even the implied warranty of
> > MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
> > GNU General Public License for more details.
> > 
> > You should have received a copy of the GNU General Public License
> > along with this program; if not, write to the Free Software
> > Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
> > ```
> 
> Yes, either the full notice (which is what I would prefer), or at least something like this:
> 
> ```
> Copyright 2019 Gentoo Authors
> Distributed under the terms of the GNU GPL version 2 or later
> ```
> 
> See https://gitweb.gentoo.org/proj/eselect.git/tree/ for examples of both (current version of full GPL license notice in `libs/*.in`, short notice in `modules/*.eselect`).

@oxr463 ^

orlitzky:
Well I don't know how to update someone else's PR, but I promised I'd help rebase it, so here's the Makefile diff:

```patch
diff --git a/Makefile b/Makefile
index 3da7c99..dd3d271 100644
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 # These "find" commands match text.xml and *.svg files, respectively,
 # but only after excluding the .git directory from the search for
 # performance and overall sanity reasons.
-HTMLS := $(subst text.xml,index.html,\
-  $(shell find ./ -name .git -prune -o \( -type f -name 'text.xml' -print \)))
+XMLS := $(shell find ./ -name .git -prune -o \( -type f -name 'text.xml' -print \))
+HTMLS := $(subst text.xml,index.html,$(XMLS))
 IMAGES := $(patsubst %.svg,%.png,\
   $(shell find ./ -name .git -prune -o \( -type f -name '*.svg' -print \)))
 
@@ -16,6 +16,12 @@ prereq:
 	{ echo "dev-libs/libxslt is required" >&2;\
 	  exit 1; }
 
+# Since search_index.py rebuilds the index from scratch instead of
+# updating it, we pass it the names of ALL prerequisites ($^) and not
+# just the names of the ones that are new ($?).
+documents.js: $(XMLS)
+	./search_index.py $^ > documents.js
+
 %.png : %.svg
 	convert $< $@
 
@@ -35,6 +41,6 @@ prereq:
 	xsltproc devbook.xsl $< > $@
 
 clean:
-	rm -f $(HTMLS) $(IMAGES)
+	rm -f $(HTMLS) $(IMAGES) documents.js
 
 .PHONY: all prereq clean
```

And the updated `search_index.py` that takes a list of files as arguments:

```python
#!/usr/bin/python
# Copyright 2019 Gentoo Authors
# Distributed under the terms of the GNU GPL version 2 or later
import json
import os.path
import sys
import xml.etree.ElementTree as ET

files = sys.argv[1:]
documents = []
url_root = 'https://devmanual.gentoo.org/'

for f in files:
    tree = ET.parse(f)
    root = tree.getroot()
    for chapter in root.findall('chapter'):
        try:
            documents.append({"name": chapter.find('title').text,
                "text": chapter.find('body').find('p').text,
                 "url": url_root + os.path.dirname(f) + '/'})
        except AttributeError:
            pass

print('var documents = ' + json.dumps(documents) + ';')
```

I have not tested that searching actually works, but it should be pretty close. Off by a line break or semicolon or something like that at worst.

orlitzky:
It also looks possible to update only the parts of the index that changed. I did a quick experiment, and if you write one `document.js` file for each `text.xml` file and then concatenate them all together at the end, then (just like with the html files), you only have to update the `document.js` files for pages that change. One final rule then spits out `var documents = ` and all of the individual `document.js` files into a top-level `documents.js`. This still overwrites the whole file, but avoids parsing XML in python a bajillion times pointlessly.

The only trick is that you have to join the individual `document.js` files together with commas, but you can just have `search_index.py` output a comma at the end (and no square brackets). I think JSON ignores a trailing comma? tl;dr it should work but I'm not debugging javascript on thanksgiving =P

gktrk:
> Well I don't know how to update someone else's PR, but I promised I'd help rebase it, so here's the Makefile diff:
> 
> ```diff
> diff --git a/Makefile b/Makefile
> index 3da7c99..dd3d271 100644
> --- a/Makefile
> +++ b/Makefile
> @@ -1,8 +1,8 @@
>  # These "find" commands match text.xml and *.svg files, respectively,
>  # but only after excluding the .git directory from the search for
>  # performance and overall sanity reasons.
> -HTMLS := $(subst text.xml,index.html,\
> -  $(shell find ./ -name .git -prune -o \( -type f -name 'text.xml' -print \)))
> +XMLS := $(shell find ./ -name .git -prune -o \( -type f -name 'text.xml' -print \))
> +HTMLS := $(subst text.xml,index.html,$(XMLS))
>  IMAGES := $(patsubst %.svg,%.png,\
>    $(shell find ./ -name .git -prune -o \( -type f -name '*.svg' -print \)))
>  
> @@ -16,6 +16,12 @@ prereq:
>  	{ echo "dev-libs/libxslt is required" >&2;\
>  	  exit 1; }
>  
> +# Since search_index.py rebuilds the index from scratch instead of
> +# updating it, we pass it the names of ALL prerequisites ($^) and not
> +# just the names of the ones that are new ($?).
> +documents.js: $(XMLS)
> +	./search_index.py $^ > documents.js
> +
>  %.png : %.svg
>  	convert $< $@
>  
> @@ -35,6 +41,6 @@ prereq:
>  	xsltproc devbook.xsl $< > $@
>  
>  clean:
> -	rm -f $(HTMLS) $(IMAGES)
> +	rm -f $(HTMLS) $(IMAGES) documents.js
>  
>  .PHONY: all prereq clean
> ```
> 

`all` is missing `documents.js`, no?

> And the updated `search_index.py` that takes a list of files as arguments:
> 
> ```python
> #!/usr/bin/python
> # Copyright 2019 Gentoo Authors
> # Distributed under the terms of the GNU GPL version 2 or later
> import json
> import os.path
> import sys
> import xml.etree.ElementTree as ET
> 
> files = sys.argv[1:]
> documents = []
> url_root = 'https://devmanual.gentoo.org/'
> 
> for f in files:
>     tree = ET.parse(f)
>     root = tree.getroot()
>     for chapter in root.findall('chapter'):
>         try:
>             documents.append({"name": chapter.find('title').text,
>                 "text": chapter.find('body').find('p').text,
>                  "url": url_root + os.path.dirname(f) + '/'})
>         except AttributeError:
>             pass
> 
> print('var documents = ' + json.dumps(documents) + ';')
> ```
> 
> I have not tested that searching actually works, but it should be pretty close. Off by a line break or semicolon or something like that at worst.

I'm confused as to why we pick the first `<p/>` in every chapter, and not the rest.

gktrk:
For those that wanna play with it without the hassle of building, I deployed it here: https://dev.gentoo.org/~gokturk/devmanual/pr93/index.html

gktrk:
I think xml parsing devmanual this way is a bit problematic and devmanual is most certainly not helping with that. Even with partial indexing, I'm in favor of this change.

orlitzky:
@gktrk

> `all` is missing `documents.js`, no?

That's the way the PR had it initially. Most people don't want to build the search index. You only need it if you're about to publish the live website. Otherwise, `grep -irl` is a 28052935% better way to do this.

> I'm confused as to why we pick the first `<p/>` in every chapter, and not the rest.

Beats me, I didn't intentionally change that, if I did.

orlitzky:
> I think xml parsing devmanual this way is a bit problematic and devmanual is most certainly not helping with that. Even with partial indexing, I'm in favor of this change.

Devil's advocate: there's a reason client-side search isn't a thing. All this is doing is concatenating the whole website into a string in a JS file, and then you're searching that. To make it work, you have to ship a whole 'nother copy of the website to the end user, just so that he can do the Javascript equivalent of `ctrl-f` on it. It would be much more straightforward (and faster, and use less bandwidth/memory) to publish a "single page" version of the devmanual that people can `ctrl-f` on themselves, without javascript.

gktrk:
> @gktrk
> 
> > `all` is missing `documents.js`, no?
> 
> That's the way the PR had it initially. Most people don't want to build the search index. You only need it if you're about to publish the live website. Otherwise, `grep -irl` is a 28052935% better way to do this.

Makes sense

> 
> > I'm confused as to why we pick the first `<p/>` in every chapter, and not the rest.
> 
> Beats me, I didn't intentionally change that, if I did.

No I think there is a bigger problem with parsing it.

gktrk:
> > I think xml parsing devmanual this way is a bit problematic and devmanual is most certainly not helping with that. Even with partial indexing, I'm in favor of this change.
> 
> Devil's advocate: there's a reason client-side search isn't a thing. All this is doing is concatenating the whole website into a string in a JS file, and then you're searching that. To make it work, you have to ship a whole 'nother copy of the website to the end user, just so that he can do the Javascript equivalent of `ctrl-f` on it. It would be much more straightforward (and faster, and use less bandwidth/memory) to publish a "single page" version of the devmanual that people can `ctrl-f` on themselves, without javascript.

I thought about it but lunr says (https://lunrjs.com/guides/index_prebuilding.html):

```
A better way is to pre-build the index, and serve a serialised index that Lunr can load on the client side much quicker.
[...]
Lunr indexes support serialisation in JSON.
[...]
This serialized index can then be written to a file, compressed, and served along side other static assets.
```

Especially with compression, this is very appealing.

gktrk:
The current problem is, if you have something like the following:

```
<p>                                                                                                                                                                                                          
Most ebuilds specify a <c>KEYWORDS</c> variable. This variable is used to                                                                                                                                    
indicate the suitability and stability of both the package and the ebuild on                                                                                                                                 
each given arch (<c>sparc</c>, <c>ppc</c>, <c>x86-obsd</c>, ...).                                                                                                                                            
</p>
```

`.findall('p')` would parse it until `Most ebuilds specify a` and break at `<c>`. Not sure how to get it to move past that yet.

orlitzky:
> 
> I thought about it but lunr says...
> 

Javascript programmers say a lot of things =)

If enough numerical evidence says that up is down, then... I'll have to accept it. But if you're claiming that

* Shipping and rendering an HTML page
* Shipping a serialized JS index
* Starting up the javascript engine
* Deserializing the index
* Performing a search in JS code
* Shipping and rendering the target page

is faster than shipping and rendering ONE page, with ctrl-f to search... then... that's a bit counter-intuitive, and I'm not going to believe it without strong evidence.

The HTML can also be shipped compressed, if you really want to (I don't recommend it).

orlitzky:
>  Not sure how to get it to move past that yet.

Case in point, the` XML -> HTML -> browser` pipeline already knows exactly how to turn that markup into meaningful text.

gktrk:
> > I thought about it but lunr says...
> 
> Javascript programmers say a lot of things =)
> 
> If enough numerical evidence says that up is down, then... I'll have to accept it. But if you're claiming that
> 
> * Shipping and rendering an HTML page
> * Shipping a serialized JS index
> * Starting up the javascript engine
> * Deserializing the index
> * Performing a search in JS code
> * Shipping and rendering the target page
> 
> is faster than shipping and rendering ONE page, with ctrl-f to search... then... that's a bit counter-intuitive, and I'm not going to believe it without strong evidence.
> 

Most likely not, although rendering such a big page vs. searching using JS have their own respective trade-offs.

> The HTML can also be shipped compressed, if you really want to (I don't recommend it).

I assume this happens at the protocol level.

gktrk:
> > Not sure how to get it to move past that yet.
> 
> Case in point, the` XML -> HTML -> browser` pipeline already knows exactly how to turn that markup into meaningful text.

I'm tempted to come up with another xsl transformation that transforms text.xml into pure ascii text for processing :D

I found an annoying solution. You access the test after a `<c/>` by getting that node and accessing its `.tail` selector.

gktrk:
It gets more complicated. I hacked the script a bit more to process more text (https://github.com/gktrk/devmanual.gentoo.org/blob/pr93/search_index.py). The resulting `documents.js` is 166KB. I deployed it at https://dev.gentoo.org/~gokturk/devmanual/pr93/index.html. If there is any slowdown from parsing such a big index file, I didn't notice on an 8-year old macbook air. The search is super ugly though, because I clumped up an entire file into a single entry. Search needs to return the paragraph that it got the hit on. That means we have to insert an entry per paragraph. Which means creating unique name entries for every one of them while also locating in which section/subsection they reside so we can correctly generate the corresponding URI.

gktrk:
Here's the next iteration: https://dev.gentoo.org/~gokturk/devmanual/pr93/index.html

Each paragraph is indexed individually up to the `subsubsection` depth. Hits are highlighted in the output. Matching the ref to text was O(n) before, now it's O(1).

Other things to consider:
- Right now I'm not indexing `<note/>`, `<warning/>` et al
- A search like `src_configure` performs very poorly because I'm not indexing section titles. Not sure how to go about it yet.

Please give me some feedback about its current usefulness.

orlitzky:
>  It would be much more straightforward (and faster, and use less bandwidth/memory) to publish a "single page" version of the devmanual that people can `ctrl-f` on themselves, without javascript.

It looks like someone else had this idea 12+ years ago (with three duplicate reports!), and a guy named Göktürk already did a proof-of-concept =)

https://bugs.gentoo.org/185477

oxr463:
Closing in favor of https://github.com/gentoo/devmanual/pull/120

PR #94, opened 2019-04-13
use-flags: Split out use cases for not using flags, and list more

mgorny:
Split out the paragraph on not using USE flags for runtime dependencies
into a separate section.  Include the policy of not using USE flags
for small files.

### DIFF
--- a/general-concepts/use-flags/text.xml
+++ b/general-concepts/use-flags/text.xml
@@ -27,6 +27,16 @@ dependency is not detected by the package manager tools and can easily
 break, among other issues.
 </p>
 
+<p>
+Automagic dependencies are preferably fixed by preparing a build system patch
+adding appropriate options to control the dependency in question, and submitting
+this patch upstream for the benefit of all users. To avoid carrying additional
+patches downstream, automagic dependencies can usually be worked around using
+special build system options (e.g. cache variables in autotools) or through
+depending on the relevant packages unconditionally (i.e. forcing the check
+to always succeed).
+</p>
+
 <note>
 The status of USE flags is saved in the VDB, and their value in
 <c>pkg_prerm</c> and <c>pkg_postrm</c> is taken from there. This means that
--- a/general-concepts/use-flags/text.xml
+++ b/general-concepts/use-flags/text.xml
@@ -70,6 +70,42 @@ of flags to those programs or modules that have external dependencies
 and/or long build times. The rest of them should be built unconditionally
 instead, or controlled by a flag such as <c>minimal</c>.
 </p>
+
+<p>
+You should not introduce USE flags that merely manipulate <c>CFLAGS</c>,
+<c>FEATURES</c> or similar variables configured directly by the user. Instead,
+packages should avoid manipulating them at all, and let users set them directly.
+Common mistakes include:
+</p>
+
+<ol>
+	<li>
+	Using <c>debug</c> USE flag to force <c>-O0 -g</c> and disable
+	stripping. The correct purpose of <c>debug</c> flag is to control additional
+	debug code paths. The use of correct flags and features to preserve
+	debugging information is user's responsibility.
+	</li>
+
+	<li>
+	Introducing <c>lto</c> flag to force <c>-flto</c>. This is something user
+	should set directly in flag varibles.
+	</li>
+
+	<li>
+	Using <c>CPU_FLAGS_*</c> to control <c>-m*</c> options. Those flags are
+	intended to control code paths explicitly requiring specific CPU extensions,
+	e.g. separate assembly. Compiler-generated assembly should respect user's
+	<c>-march</c> choice.
+	</li>
+</ol>
+
+<p>
+There might be corner cases where these rules do not apply. For example, a few
+upstreams require users to use specific <c>CFLAGS</c> and reject bug reports
+against builds using other values. In this case, it is customary to strip flags
+by default and provide <c>custom-cflags</c> flag to allow users to force their
+preferred flags.
+</p>
 </body>
 </section>
 
--- a/general-concepts/use-flags/text.xml
+++ b/general-concepts/use-flags/text.xml
@@ -5,40 +5,40 @@
 
 <body>
 <p>
-	<c>USE</c> flags are to control optional dependencies and settings which
-	the user may reasonably want to select. For example, <c>app-editors/vim</c>
-	can optionally build with support for the <c>ruby</c> interpreter, and it
-	needs <c>dev-lang/ruby</c> installed to do this <d/> we use the ruby
-	<c>USE</c> flag to provide this option. On the other hand,
-	<c>app-text/glark</c> requires <c>ruby</c> no matter what, so no <c>USE</c>
-	flag is used there.
+<c>USE</c> flags are to control optional dependencies and settings which
+the user may reasonably want to select. For example, <c>app-editors/vim</c>
+can optionally build with support for the <c>ruby</c> interpreter, and it
+needs <c>dev-lang/ruby</c> installed to do this <d/> we use the ruby
+<c>USE</c> flag to provide this option. On the other hand,
+<c>app-text/glark</c> requires <c>ruby</c> no matter what, so no <c>USE</c>
+flag is used there.
 </p>
 
 <p>
-	No combination of <c>USE</c> flags should cause a package to fail to build
-	because users can set any combination of flags.
+No combination of <c>USE</c> flags should cause a package to fail to build
+because users can set any combination of flags.
 </p>
 
 <p>
-	Packages should not configure and link based upon what is available at
-	compile time <d/> any autodetection must be overridden. This is commonly
-	referred to as the dependency being "automagic" - This is bad because the
-	dependency is not detected by the package manager tools and can easily
-	break, among other issues.
+Packages should not configure and link based upon what is available at
+compile time <d/> any autodetection must be overridden. This is commonly
+referred to as the dependency being "automagic" - This is bad because the
+dependency is not detected by the package manager tools and can easily
+break, among other issues.
 </p>
 
 <p>
-	The usage of a <c>USE</c> flag should not control runtime dependencies when
-	the package does not link to it. Doing so will create extra
-	configuration for the package and re-compilation for no underlying file
-	change on disk. This should be avoided and instead can be conveyed to the
-	user via post install messages if needed.
+The usage of a <c>USE</c> flag should not control runtime dependencies when
+the package does not link to it. Doing so will create extra
+configuration for the package and re-compilation for no underlying file
+change on disk. This should be avoided and instead can be conveyed to the
+user via post install messages if needed.
 </p>
 
 <note>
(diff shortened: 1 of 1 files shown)


mgorny:
Ping.

mgorny:
Ok, included a commit reformatting the whole file first.

general-concepts/use-flags/text.xml
```diff
@@ -5,113 +5,179 @@
...
 </p>
 
 <p>
-	Packages should not configure and link based upon what is available at
-	compile time <d/> any autodetection must be overridden. This is commonly
-	referred to as the dependency being "automagic" - This is bad because the
-	dependency is not detected by the package manager tools and can easily
-	break, among other issues.
+Packages should not configure and link based upon what is available at
+compile time <d/> any autodetection must be overridden. This is commonly
```
orlitzky: This might be a good place to mention the easiest solution (and the alternative to USE flags) regarding automagic dependencies: require them unconditionally. If some ./configure script will introduce an automagic dependency when a package is installed, depending on that package unconditionally (i.e. without a USE flag) prevents any breakage from that package's removal.

general-concepts/use-flags/text.xml
```diff
@@ -5,113 +5,179 @@
...
+or modules. Whenever they could result in a large number of <c>USE</c> flags
+that would force the user to spend a lot of time choosing the necessary flags
+and possibly rebuilding after incomplete choices, you should consider
+reducing the use of flags to those programs or modules that have external
+dependencies and/or long build times. The rest of them should be built
+unconditionally instead, or controlled by a flag such as <c>minimal</c>.
+</p>
+
+<p>
+It is customary not to introduce USE flags that merely manipulate
```
orlitzky: I had a hard time reading "customary not to....". I think the phrase "it is customary to" is rather common in english, but "customary not to" is comparatively rare. So when the brain sees that combination of letters, it reads "customary to" regardless of what the text actually says. In any case, this wording could be stronger: "you should not..." and that solves my problem too.

general-concepts/use-flags/text.xml
mgorny: Added a paragraph listing all possible solutions, 'correct' and 'cheap'.

general-concepts/use-flags/text.xml
mgorny: Done.

general-concepts/use-flags/text.xml
```diff
@@ -27,6 +27,23 @@ dependency is not detected by the package manager tools and can easily
 break, among other issues.
 </p>
 
+<note>
+The status of USE flags is saved in the VDB, and their value in
```
gktrk: s/is/are/, s/value/values/

general-concepts/use-flags/text.xml
```diff
@@ -27,6 +27,23 @@ dependency is not detected by the package manager tools and can easily
...
+</note>
+</body>
+
+<section>
+<title>When not to use USE flags?</title>
+<body>
+<p>
+While <c>USE</c> flags are generally considered beneficial to users, there
+are valid use cases for avoiding them. When writing ebuilds, you should
+consider whether you should add flags for particular conditional features,
```
gktrk: s/whether you should/whether to/

general-concepts/use-flags/text.xml
```diff
@@ -27,6 +27,23 @@ dependency is not detected by the package manager tools and can easily
...
+setting or unsetting a USE flag between merge and unmerge has no effect.
+</note>
+</body>
+
+<section>
+<title>When not to use USE flags?</title>
+<body>
+<p>
+While <c>USE</c> flags are generally considered beneficial to users, there
+are valid use cases for avoiding them. When writing ebuilds, you should
```
gktrk: s/you should//

general-concepts/use-flags/text.xml
```diff
@@ -27,6 +27,23 @@ dependency is not detected by the package manager tools and can easily
...
+consider whether you should add flags for particular conditional features,
+or explore one of the alternative solutions.
```
gktrk: either say "explore one of the alternative solutions **described below.**" or "explore alternative solutions."

general-concepts/use-flags/text.xml
```diff
@@ -35,12 +52,25 @@ change on disk. This should be avoided and instead can be conveyed to the
...
 
-<note>
-The status of USE flags is saved in the VDB, and their value in
-<c>pkg_prerm</c> and <c>pkg_postrm</c> is taken from there. This means that
-setting or unsetting a USE flag between merge and unmerge has no effect.
-</note>
+<p>
+<c>USE</c> flags must not be used to control installing files that are small,
+non-intrusive, do not introduce additional build-time dependencies or cause
+significant increase of build time. Examples of such files are bash completion
```
gktrk: s/significant increase of build time/a significant increase in build time/

general-concepts/use-flags/text.xml
```diff
@@ -35,12 +52,25 @@ change on disk. This should be avoided and instead can be conveyed to the
...
+files, init.d scripts, systemd service files. The rationale is same as above.
```
gktrk: s/same/the same/

general-concepts/use-flags/text.xml
```diff
@@ -35,12 +52,25 @@ change on disk. This should be avoided and instead can be conveyed to the
...
+Instead, those files must be installed unconditionally.
+</p>
+
+<p>
+A similar case can be made for packages having multiple conditional programs
+or modules. Whenever they could result in a large number of <c>USE</c> flags
```
gktrk: s/they could result/this results/

general-concepts/use-flags/text.xml
```diff
@@ -35,12 +52,25 @@ change on disk. This should be avoided and instead can be conveyed to the
...
+that would force the user to spend a lot of time choosing the necessary flags
```
gktrk: s/the necessary/compatible/

general-concepts/use-flags/text.xml
```diff
@@ -35,12 +52,25 @@ change on disk. This should be avoided and instead can be conveyed to the
...
+and possibly rebuilding after incomplete choices, you should consider
```
gktrk: s/you should//

general-concepts/use-flags/text.xml
```diff
@@ -69,6 +69,42 @@ reducing the use of flags to those programs or modules that have external
 dependencies and/or long build times. The rest of them should be built
 unconditionally instead, or controlled by a flag such as <c>minimal</c>.
 </p>
+
+<p>
+You should not introduce USE flags that merely manipulate <c>CFLAGS</c>,
+<c>FEATURES</c> and similar variables configured directly by the user. Instead,
```
gktrk: s/and/or/

general-concepts/use-flags/text.xml
```diff
@@ -69,6 +69,42 @@ reducing the use of flags to those programs or modules that have external
...
+the packages should avoid manipulating them at all, and let user set them
```
gktrk: s/the packages/packages/

gktrk: either go "the user" or "users"

general-concepts/use-flags/text.xml
```diff
@@ -69,6 +69,42 @@ reducing the use of flags to those programs or modules that have external
...
+	<li>
+	Using <c>CPU_FLAGS_*</c> to control <c>-m*</c> options. Those flags are
+	intended to control code paths explicitly requiring specific CPU extensions,
+	e.g. separate assembly. Compiler-generated assembly should respect user's
+	<c>-march</c> choice.
+	</li>
+</ol>
+
+<p>
+There might be a few corner cases where those rules do not apply. For example,
```
gktrk: s/might be a few corner/may be/, so that we don't repeat "a few" in two sentences back to back. Also s/those/these/ IMO.

general-concepts/use-flags/text.xml
```diff
@@ -27,6 +27,16 @@ dependency is not detected by the package manager tools and can easily
 break, among other issues.
 </p>
 
+<p>
+Automagic dependencies are preferably fixed by preparing a build system patch
+adding appropriate options to control the dependency in question, and submitting
+this patch upstream for benefit of all users.  To avoid carrying additional
```
gktrk: s/benefit/the benefit/

You also have two whitespace characters before 'To'

general-concepts/use-flags/text.xml
```diff
@@ -27,6 +27,16 @@ dependency is not detected by the package manager tools and can easily
...
+patches downstream, automagic dependencies can usually be worked around using
+special build system options (e.g. cache variables in autotools) or through
```
gktrk: cache variables or cached variables?

general-concepts/use-flags/text.xml
```diff
@@ -5,113 +5,189 @@
...
+the package does not link to it. Doing so will create extra
+configuration for the package and re-compilation for no underlying file
+change on disk. This should be avoided and instead can be conveyed to the
+user via post install messages if needed.
+</p>
+
+<p>
+<c>USE</c> flags must not be used to control installing files that are small,
+non-intrusive, do not introduce additional build-time dependencies or cause
+significant increase of build time. Examples of such files are bash completion
```
orlitzky: If you want to mention logrotate entries by name, then this also closes https://bugs.gentoo.org/471388

general-concepts/use-flags/text.xml
```diff
@@ -27,6 +27,23 @@ dependency is not detected by the package manager tools and can easily
 break, among other issues.
 </p>
 
+<note>
+The status of USE flags is saved in the VDB, and their value in
+<c>pkg_prerm</c> and <c>pkg_postrm</c> is taken from there. This means that
```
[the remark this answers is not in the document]
mgorny: That was actually only moved without touching the contents. But sure, I can change it.

general-concepts/use-flags/text.xml
mgorny: 'status [..] are'?

mgorny: I'll make it 'states'.

general-concepts/use-flags/text.xml
mgorny: 'cache variable', as used in 'info autoconf'.

general-concepts/use-flags/text.xml
mgorny: Done that.

mgorny:
Implemented all grammar suggestions, and also added logrotate files to the list.

ulm:
>>>>> On Thu, 25 Apr 2019, Michał Górny wrote:

> +<note>
> +The status of USE flags is saved in the VDB, and their value in

> 'status [..] are'?

It's "status" both for singular and plural. Latin u-declension. :)

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/858a4bf...168d155.  Thanks!

PR #95, opened 2019-04-22
ebuild-writing: Stop encouraging empty USE_EXPAND hack

mgorny:
Remove the suggestion that 'no flags = all options'.  We certainly
aren't doing it these days, and the problem is better resolved with
defaulting USE_EXPAND and/or REQUIRED_USE these days.


ulm:
Looks good to me.

gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/168d155...059da25. Thanks!

PR #96, opened 2019-04-22
Update code snippets in quickstart, and copyright headers

mgorny:
1. Update copyright header to state 'Gentoo Authors' everywhere.
2. Update the code snippets in quickstart to be more modern:
   * use eapply instead of epatch
   * avoid redefining default src_install whenever it would just work
   * prefer `myconf` array over backslash hell for large number of flags
3. Correct licenses

### DIFF
--- a/quickstart/text.xml
+++ b/quickstart/text.xml
@@ -225,11 +225,6 @@ DEPEND="${RDEPEND}
 src_configure() {
 	econf --with-popt
 }
-
-src_install() {
-	emake DESTDIR="${D}" install
-	dodoc README CHANGES
-}
 </codesample>
 
 <p>
@@ -241,7 +236,9 @@ variables <d/> see
 </p>
 
 <p>
-Again, we define <c>src_configure</c> and <c>src_install</c> functions.
+We define <c>src_configure</c> only.  <c>src_install</c> does not need to
+be defined since the default implementation calling <c>emake install</c>
+and installing common documentation files works correctly for the package.
 </p>
 
 <p>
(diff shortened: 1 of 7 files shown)


quickstart/text.xml
```diff
@@ -397,20 +381,23 @@ DEPEND="${RDEPEND}
...
+		$(use_enable ssl)
+		$(use_enable crypt gpgme)
+		$(use_enable pda jpilot)
+		$(use_enable ldap)
+		$(use_enable ipv6)
+		$(use_enable imlib)
 		$(use_enable xface compface)
+	)
+
+	econf "${myconf[@]}"
```
ulm: That unnecessarily introduces a variable, and I think it is a matter of personal preference.
Also the existing example is easier to read (IMHO), so please leave it as it is.

mgorny: There is no harm in using a local variable. Relying on backslashes is harmful, especially when someone accidentally forgets a backslash in middle of it. And yes, I've seen such mistakes in practice.

mgorny: Also, case of passing flags directly to econf is presented in another example. I think it's beneficial to present more than one possible solution.

ulm: I disagree. This is just your personal preference, not something that needs to be recommended in general. That there can be unrelated mistakes is not an argument. The variable could be misspelt as well, or the quotes be omitted or whatnot.

mgorny: Then just skip that one commit.

ulm:
OK, if there are no further comments until tomorrow, I am going to merge this series, except for commit 
f1fafdc .

ulm:
Cherry-picked commits 1, 2, 3, and 5. Closing.

PR #97, opened 2019-05-24
general-concepts/use-flags: remove redundant 'die' from the example

steils:
Signed-off-by: Stefan Strogin <steils@gentoo.org>

### DIFF
--- a/general-concepts/use-flags/text.xml
+++ b/general-concepts/use-flags/text.xml
@@ -283,7 +283,7 @@ src_compile() {
 		# Other stuff
 		${myconf}
 
-	emake || die "make failed"
+	emake
 }
 </codesample>
 


general-concepts/use-flags/text.xml
```diff
@@ -276,14 +276,14 @@ src_compile() {
 	elif use ssl &amp;&amp; ! use gnutls ; then
 		myconf="${myconf} --enable-ssl --with-ssl=openssl"
 	else
-	myconf="${myconf} --disable-ssl"
+		myconf="${myconf} --disable-ssl"
```
ulm: All these assignments could profit from using +=, e.g. `myconf+=" --disable-ssl"`

steils: Why `+=`? There is no previous assignment in this example.
I think the example should be rewritten if we want to show profit from using `+=`. But that is for another PR.

ulm: Sure, we can be strict and split issues into different PRs. But then that indentation fix doesn't belong here either. :stuck_out_tongue:

steils: Sure:D

ulm:
Pushed.

PR #98, opened 2019-06-17
Correct IEEE Std section number

andrey-utkin:


ulm:
Note that the link was definitely correct when I added it in 2015:
https://web.archive.org/web/20150413074304/http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_278
I think this is the first time that upstream changes things in place, without issuing a new document number.

Patch looks good, but could you please rewrap the rest of the paragraph (lines 31-33)?

andrey-utkin:
I see, it's weird that they've done so.
Rewrapped the paragraph.

ulm:
Sorry for noticing only now (when I was going to merge), but your commits are missing a Signed-off-by line:
https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin

andrey-utkin:
Perhaps add signoff check to CI.

PR #101, opened 2019-08-24
Fix logo in header on mobile screens

oxr463:
This is how it should look based on the Wiki,

photo5030524679764420634

This is how it looks for the Developer Manual,

photo5030524679764420635

### DIFF
--- a/devmanual.css
+++ b/devmanual.css
@@ -1,3 +1,13 @@
+.logo {
+    display: inline;
+}
+
+@media only screen and (min-width: 380px) {
+    .logo {
+        display: inline-block;
+    }
+}
+
 div.header {
     border-bottom: 3px solid #333333;
     background-color: #666666;


oxr463:
@gktrk, @ulm, could somebody please review this PR. Thanks!

oxr463:
ping @maffblaster, @gktrk

devmanual.css
```diff
@@ -1,3 +1,13 @@
+.logo {
+    display: inline;
+}
+
+@media only screen and (min-width: 380px) {
+    .logo {
+        display: inline-block;
+    }
+}
+
```
ulm: Sorry for the late reply.

I don't see any effect of this in my tests. devmanual.css is loaded before tyrian.min.css, and IIUC the latter overrides the former by setting `display: inline-block;` unconditionally for the `.logo`.

What am I missing here?

oxr463: Looks like I got this one wrong. If you navigate to https://devmanual.gentoo.org with Responsive Design Mode set to 320 x 480 in Firefox Developer tools, you will see that `display: inline-block` is causing the image to shift off screen. By setting it to `display: inherit`, the logo then bumps back up to the proper location. I was attempting to only set inline-block if the page width is 380px or greater. Otherwise it should be inherited.

ulm: I still see no difference, even if I change `inline` to `inherit` in your patch. AFAICS it has no effect bacause it is superseded by the `.logo` block in the Tyrian layout.

ulm:
BTW, https://packages.gentoo.org/ suffers from the same problem. This may indicate that it is a more general issue with the Tyrian layout, which should be solved there.

devmanual.css
oxr463: Yes, I think it is in the wrong order.

oxr463:
> 
> 
> BTW, https://packages.gentoo.org/ suffers from the same problem. This may indicate that it is a more general issue with the Tyrian layout, which should be solved there.

Tyrian no longer seems to be available on GitHub, so I guess I will just file a report on bugzilla instead.

gktrk:
> > BTW, https://packages.gentoo.org/ suffers from the same problem. This may indicate that it is a more general issue with the Tyrian layout, which should be solved there.
> 
> Tyrian no longer seems to be available on GitHub, so I guess I will just file a report on bugzilla instead.

@oxr463 https://gitweb.gentoo.org/sites/tyrian-theme.git/

oxr463:
Bug report: https://bugs.gentoo.org/705668

PR #103, opened 2019-09-01
general-concepts/licenses: Elaborate on combining licenses

mgorny:
Add a section explaining how to determine the effective license when
a file combines sources with different licenses.  Give explicit examples
with BSD/GPL and GPL/GPL.  Rephrase the license conflict warning to make
it clearer.  Link to GNU's 'License compatibility' article.

### DIFF
--- a/general-concepts/licenses/text.xml
+++ b/general-concepts/licenses/text.xml
@@ -82,11 +82,34 @@ versions are specified elsewhere.
 </p>
 
 <p>
-Please watch for license conflicts. If the license indicated
-by the package is incompatible with the licenses used by its sources
-(e.g. BSD/MIT package including GPL sources), please contact
-the licenses team for guidance. Do not add packages that seem
-to include license term violations.
+If an installed file is built from sources using different licenses, all those
+licenses must be compatible. The effective license for the file will be the most
+restrictive of input licenses. For example:
+</p>
+
+<ol>
+  <li>
+    If a program includes code that is partially covered by GPL-2
+    and partially by the 3-clause BSD license, the license for
+    the resulting executable will be GPL-2 since it is more restrictive
+    than the other one.
+  </li>
+  <li>
+    If a program includes code that is partially covered by GPL-2+
+    and partially by GPL-3[+], the latter license will be effective
+    since it is the subset of both terms.
+  </li>
+  <li>
+    If a program includes code that is partially covered by GPL-2+
+    and partially by GPL-2, GPL-2 will be the effective subset.
+  </li>
+</ol>
+
+<p>
+Please watch out for license conflicts. If the package combines incompatible
+licenses (e.g. GPL-2 and GPL-3 code without 'or newer' clauses), please contact
+the licenses team for guidance. Do not add packages that seem to include
+license term violations.
 </p>
 
 </body>
--- a/general-concepts/licenses/text.xml
+++ b/general-concepts/licenses/text.xml
@@ -167,6 +167,32 @@ you are unsure as to the meaning of any part of it.
 
 </body>
 </section>
+
+<section>
+<title>Further reading</title>
+<body>
+
+<p>
+The following resources are recommended as source of more information
+on copyright and licensing:
+</p>
+
+<ul>
+  <li>
+    <uri link="https://dwheeler.com/essays/floss-license-slide.html">
+      David A. Wheeler, The Free-Libre / Open Source Software (FLOSS)
+      License Slide
+    </uri>
+  </li>
+  <li>
+    <uri link="https://www.gnu.org/licenses/license-compatibility.en.html">
+      Richard Stallman, License Compatibility and Relicensing
+    </uri>
+  </li>
+</ul>
+
+</body>
+</section>
 </body>
 
 </chapter>


mgorny:
Disclaimer: I've permitted maintainer edits here, so if you have any grammar corrections, please either commit it to PR branch or straight to the repo. It'll be less work for all of us than having the review-edit ping-pong.

general-concepts/licenses/text.xml
```diff
@@ -82,11 +82,38 @@ versions are specified elsewhere.
...
-(e.g. BSD/MIT package including GPL sources), please contact
-the licenses team for guidance. Do not add packages that seem
-to include license term violations.
+If an installed file is built from sources using different licenses, all those
+licenses must be compatible. The effective license for the file will be the most
+restrictive of input licenses. For example:
+</p>
+
+<ol>
+	<li>
```
ulm: Follow the style guide please, indentation in lists is exactly 2 spaces per level.

general-concepts/licenses/text.xml
```diff
@@ -82,11 +82,38 @@ versions are specified elsewhere.
...
+		If a program includes both GPL- and BSD-licensed code, the license
+		for the resulting program will be respective GPL license since it is more
```
ulm: "will be respective" sounds strange.

general-concepts/licenses/text.xml
```diff
@@ -82,11 +82,38 @@ versions are specified elsewhere.
...
+Please watch out for license conflicts. If the package combines incompatible
+licenses (e.g. GPL-2 and GPL-3 code without 'or newer' clauses), please contact
+the licenses team for guidance. Do not add packages that seem to include
+license term violations.
+</p>
+
+<p>
+For more information, please see
+<uri link="https://www.gnu.org/licenses/license-compatibility.en.html">
+License Compatibility and Relicensing</uri> article.
```
ulm: Maybe add that this only describes compatibility for the GPL family?

general-concepts/licenses/text.xml
mgorny: Ok, I will reword the whole list to sound a little bit less strange.

general-concepts/licenses/text.xml
mgorny: I disagree. The first few paragraph cover different lax licenses as well.

mgorny:
I've split the link into 'Further reading' section, and included another as suggested by @gktrk. I've also noticed we're using subsections instead of sections, and fixed that.

general-concepts/licenses/text.xml
```diff
@@ -82,17 +82,39 @@ versions are specified elsewhere.
...
-to include license term violations.
+If an installed file is built from sources using different licenses, all those
+licenses must be compatible. The effective license for the file will be the most
+restrictive of input licenses. For example:
+</p>
+
+<ol>
+  <li>
+    If a program includes code that is partially covered by GPL-2
+    and partially by the BSD license, the license for the resulting
```
ulm: Maybe a clarification like "3-clause BSD license" would be helpful here. The original BSD license (which we list as BSD-4) is not GPL compatible.

mgorny: Well, I meant to use terms from LICENSE but sure.

mgorny:
Updated.

mgorny:
> My only concern is with the `GPL-3[+]` notation. You meant `GPL-3 or GPL-3+` right?

Yes. Do you need me to submit a fix or will you change it while merging?

gktrk:
> > My only concern is with the `GPL-3[+]` notation. You meant `GPL-3 or GPL-3+` right?
> 
> Yes. Do you need me to submit a fix or will you change it while merging?

Don't bother, I'll take care of it.

gktrk:
> > > My only concern is with the `GPL-3[+]` notation. You meant `GPL-3 or GPL-3+` right?
> > 
> > 
> > Yes. Do you need me to submit a fix or will you change it while merging?
> 
> Don't bother, I'll take care of it.

I've decided to convert all `+`s in that subsection to `or later` variants because the `+` syntax is not explained until the section that follows it (`GPL-x vs GPL-x+`)

mgorny:
Ping.

mgorny:
It's a bit different concept.

'Effective license' is what affects a single file. For example, if you have GPL-2+ with public-domain or BSD-2 snippet in one of the source files, you usually don't want to bother the user with the extra license (because he won't see it installed without the GPL-2+ code anyway). Similarly, if package has some GPL-2+ and GPL-3+ source files (that create GPL-3+ executable as a result), listing 'GPL-2+ GPL-3+ would be weird.

Listing multiple licenses is generally applicable if the package installs multiple files with different licenses, e.g. a LGPL library and GPL program. From user's perspective, it means 'you have to look in detail what applies where'.

ulm:
> 'Effective license' is what affects a single file. For example, if you have GPL-2+ with public-domain or BSD-2 snippet in one of the source files, you usually don't want to bother the user with the extra license (because he won't see it installed without the GPL-2+ code anyway).

If you create a binary from (e.g.) GPL and BSD licensed source files, it will be a derivative work of both. This means that the conditions of **both** licenses apply (and clause 2 of the BSD license explicitly requires this). So, the derivative work can be distributed neither under GPL alone, nor under BSD alone. This means that all of these licenses should be listed in the `LICENSE` variable.

https://opensource.stackexchange.com/questions/304/can-i-take-bsd-licensed-code-and-distribute-it-under-gpl

> Similarly, if package has some GPL-2+ and GPL-3+ source files (that create GPL-3+ executable as a result), listing 'GPL-2+ GPL-3+ would be weird.

I agree on this one. Reading your change again, maybe it should explicitly say that in case of the second and third item only the most restrictive license needs to be listed (GPL-3[+] and GPL-2, respectively), but in case of the first item, both BSD and GPL-2 must be included.

orlitzky:
Having `LICENSE="GPL-2+ GPL-3+"` is rather harmless, while (as ulm pointed out) having `LICENSE="GPL-2+"` instead of `LICENSE="GPL-2+ BSD"` is bad. I think giving developers the freedom to think about this will result in mistakes like that, and we won't gain much from it.

That said, I don't feel strongly about it. If this is useful for mgorny, I don't want to stand in the way of it.

ulm:
The ebuild should simply list all licenses of the installed package. The reason for this is that everything else would be error prone, and would require devs to determine if the installed package is a derivative work or a collective work.

So, I'd suggest that we follow the KISS principle, and don't try to be too clever with our rules.

gktrk:
> The ebuild should simply list all licenses of the installed package. The reason for this is that everything else would be error prone, and would require devs to determine if the installed package is a derivative work or a collective work.
> 
> So, I'd suggest that we follow the KISS principle, and don't try to be too clever with our rules.

:+1: 

I don't like the idea of deferring to developers when it comes to deciding what the _effective_ license is. As long as they can spot potential violations, I don't care if they put ugly-yet-valid license combinations in ebuilds. I personally most certainly would put `LICENSE="GPL-2+ GPL-3+"` in an ebuild, because it's straightforward to do so.

If you want to write a blog post about this and list it in further reading, I'm completely okay with that. I propose to merge the hunk about the section on further reading only. What do you guys think?

ulm:
I have cherry-picked commit 46b662b098a16438eea31ac64e8d0fee0a2a9455 to master.

Commit af3ec235e0af117a85dcf89a8e1834686bafc693 has conflicts (with your own commit be80788debfcc0035e9d014956763871fd5df287), which I don't know how to resolve.

gktrk:
> I have cherry-picked commit [46b662b](https://github.com/gentoo/devmanual/commit/46b662b098a16438eea31ac64e8d0fee0a2a9455) to master.
> 
> Commit [af3ec23](https://github.com/gentoo/devmanual/commit/af3ec235e0af117a85dcf89a8e1834686bafc693) has conflicts (with your own commit [be80788](https://github.com/gentoo/devmanual/commit/be80788debfcc0035e9d014956763871fd5df287)), which I don't know how to resolve.

Are we picking commit af3ec23 too? I've objected it but I'd like to have your opinion as well.

ulm:
> Are we picking commit [af3ec23](https://github.com/gentoo/devmanual/commit/af3ec235e0af117a85dcf89a8e1834686bafc693) too? I've objected it but I'd like to have your opinion as well.

I got confused about the status (because it says "approved"). Reading your comment from 2019-12-19 again, I am closing this PR.

PR #104, opened 2019-09-05
Elaborate on package maintainers, kill old herds section

mgorny:
Reuse 'Herds and Projects' chapter as more generic 'Package Maintainers'. Remove obsolete section on herds, and elaborate in detail on maintainers.

### DIFF
--- a/general-concepts/package-maintainers/text.xml
+++ b/general-concepts/package-maintainers/text.xml
@@ -1,0 +1,93 @@
+<?xml version="1.0"?>
+<guide self="general-concepts/package-maintainers/">
+<chapter>
+<title>Package Maintainers</title>
+<body>
+
+<p>
+Package maintainers are responsible for ebuild maintenance tasks <d/>
+bumping ebuilds to new upstream releases, updating them as necessary
+with regards to policies, replying to bugs, etc. The maintainers
+of a package are listed in the package's <uri
+link="::ebuild-writing/misc-files/metadata/">metadata.xml</uri> file.
+When assigning bugs, the first maintainer listed becomes the bug's
+assignee and the remaining maintainers are added to CC, unless stated
+otherwise in the metadata.
+</p>
+
+<section>
+<title>Maintainer authority</title>
+<body>
+
+<p>
+Package maintainers have authority over their maintained packages.
+Unless specified otherwise, you should request maintainers' approval
+before committing to their packages. Try filing a bug, catching them
+on IRC or sending a mail first. The exceptions to this rule are trivial
+changes implied by your own actions, e.g. dependency updates as a result
+of package moves.
+</p>
+
+<p>
+In case of maintainer inactivity, it is a good idea to consult other
+developers on how to handle the situation, especially if it is
+a critical issue that needs to be handled ASAP. A customary tool
+of maintainer timeout may be used in this case. This timeout can be
+invoked if none of the package's maintainers reply to a patch within
+2<d/>4 weeks of it being attached or linked to a correctly assigned bug,
+or being sent to the maintainer.
+</p>
+
+<important>
+Common sense dictates to us that toolchain/base-system/core packages
+and eclasses or anything else which is delicate (e.g. glibc) or widely
+used (e.g. GTK+) should usually be left to their maintainers.
+If in doubt, don't touch it.
+</important>
+
+<p>
+Respect developers' coding preferences. Unnecessarily changing
+the syntax of an ebuild can cause complications for others. Syntax
+changes should only be done if there is a real benefit, such as faster
+compilation, improved information for the end user, or compliance
+with Gentoo policies.
+</p>
+
+</body>
+</section>
+
+<section>
+<title>Adding and removing maintainers</title>
+<body>
+
+<p>
+All packages newly added to the Gentoo repository must have at least
+one maintainer specified. New maintainers can only be added with their
+consent. In particular, it is not acceptable to add generic projects
+(such as the Python project) as package maintainers without the approval
+of their members or against their explicit policy.
+</p>
+
+<p>
+Maintainers without commit access are called proxied maintainers. Their
+changes must be pushed by a Gentoo developer, who acts as their proxy.
+All packages maintained by proxied maintainers must explicitly list
+their proxy developer or project in <c>metadata.xml</c>.
+</p>
+
+<p>
+When adding new maintainers or removing old maintainers, please remember
+to reassign bugs. The last maintainer to resign from maintaining
+the package must place a <c>&lt;!-- maintainer-needed --&gt;</c> comment
+in <c>metadata.xml</c>, in order simplify finding unmaintained packages.
+It is also required to send an 'up for grabs' email to the
+<c>gentoo-dev-announce</c> and <c>gentoo-dev</c> mailing lists,
+providing a list of newly-unmaintained packages.
+</p>
+
+</body>
+</section>
+
+</body>
+</chapter>
+</guide>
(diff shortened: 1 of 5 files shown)


general-concepts/package-maintainers/text.xml
```diff
@@ -5,6 +5,63 @@
 
 <body>
 
+<section>
+<title>Maintainers</title>
+<body>
+
+<p>
+Package maintainers are responsible for ebuild maintenance tasks —
```
ulm: Use `<d/>` for a dash (so it's the same unicode dash throughout the document).

general-concepts/package-maintainers/text.xml
```diff
@@ -5,6 +5,63 @@
...
+bumping them to new upstream releases, updating as necessary wrt
```
ulm: I think abbreviations like "wrt" should be avoided in the devmanual.

general-concepts/package-maintainers/text.xml
```diff
@@ -5,6 +5,63 @@
...
+assignee and the remaining maintainers are added to CC.
+</p>
+
+<p>
+Package maintainers have authority over their maintained packages.
+Unless specified otherwise, you should request maintainers' approval
+before committing to their packages. The exceptions to this rule are
+trivial changes implied by your own actions, e.g. dependency updates
+as a result of package moves.
+</p>
```
ulm: This is somewhat redundant to https://devmanual.gentoo.org/ebuild-maintenance/maintenance-tasks/index.html#touching-other-developers-ebuilds. Maybe shorten, and add a cross reference?

general-concepts/package-maintainers/text.xml
```diff
@@ -5,6 +5,63 @@
...
+Maintainers do not necessarily have to have commit access. Maintainers
+without it are called proxied maintainers, and they proxy their commits
+via another developer or project. All packages maintained by proxied
+maintainers must explicitly list their proxy in <c>metadata.xml</c>.
+</p>
+
+<p>
+When adding new maintainers or removing old maintainers, please remember
+to reassign bugs. The last maintainer to resign from maintaining
+the package must place a <c>maintainer-needed</c> comment
```
ulm: Be explicit here, i.e. `<!-- maintainer-needed -->`?

general-concepts/package-maintainers/text.xml
```diff
@@ -5,6 +5,63 @@
...
+Unless specified otherwise, you should request maintainers' approval
+before committing to their packages. The exceptions to this rule are
+trivial changes implied by your own actions, e.g. dependency updates
+as a result of package moves.
+</p>
+
+<p>
+In case of maintainer's inactivity, a customary tool of maintainer
+timeout may be used. This timeout can be invoked if none of
+the package's maintainers reply to a patch within 4 weeks of it being
```
ulm: The other section (see URL above) says "a soft limit of 2 to 4 weeks" so this should say the same.

general-concepts/package-maintainers/text.xml
```diff
@@ -5,6 +5,63 @@
...
+one maintainer specified. New maintainers can only be added with their
+consent. In particular, it is not acceptable to add generic projects
+(such as the Python project) as package maintainers without approval
+from their members or against their explicit policy.
+</p>
+
+<p>
+Maintainers do not necessarily have to have commit access. Maintainers
+without it are called proxied maintainers, and they proxy their commits
+via another developer or project. All packages maintained by proxied
```
ulm: Proxied maintainers don't proxy; proxy maintainers do. :-)

This should probably say something like "... and another developer or project pushes their commits as a proxy."

general-concepts/package-maintainers/text.xml
mgorny: Hmm, maybe we should move it all here. I feel like 'maintenance tasks' is really a large set of 'this and that', and not really suitable for finding what you're looking for.

mgorny: I went that way. I've added remaining data to maintainers section, split it into two subsections and removed it from 'maintenance tasks'.

general-concepts/package-maintainers/text.xml
```diff
@@ -0,0 +1,183 @@
...
+one maintainer specified. New maintainers can only be added with their
+consent. In particular, it is not acceptable to add generic projects
+(such as the Python project) as package maintainers without approval
+from their members or against their explicit policy.
+</p>
+
+<p>
+Maintainers do not necessarily have to have commit access. Maintainers
+without it are called proxied maintainers, and another developer pushes
+their commits as a proxy. All packages maintained by proxied maintainers
```
gktrk:
```suggestion
who acts as their proxy. All packages maintained by proxied maintainers
```

general-concepts/package-maintainers/text.xml
```diff
@@ -0,0 +1,183 @@
...
+must explicitly list their proxy developer or project
+in <c>metadata.xml</c>.
+</p>
+
+<p>
+When adding new maintainers or removing old maintainers, please remember
+to reassign bugs. The last maintainer to resign from maintaining
+the package must place a <c>&lt;!-- maintainer-needed --&gt;</c> comment
+in <c>metadata.xml</c>, in order to make it easy to find unmaintained
```
gktrk:
```suggestion
in <c>metadata.xml</c>, in order simplify finding unmaintained
```

mgorny:
Applied all @gktrk 's suggestions. 'Projects' now remain a separate section, and 'Package Maintainers' are added separately to them. I've also rewritten the inactivity section to retain that original advice.

@gktrk, if you have further minor changes, please apply them while merging.

mgorny:
Ping.

general-concepts/package-maintainers/text.xml
```diff
@@ -0,0 +1,93 @@
...
+</p>
+
+<p>
+In case of maintainer inactivity, it is a good idea to consult other
+developers on how to handle the situation, especially if it is
+a critical issue that needs to be handled ASAP. A customary tool
+of maintainer timeout may be used in this case. This timeout can be
+invoked if none of the package's maintainers reply to a patch within
+2<d/>4 weeks of it being attached or linked to a correctly assigned bug,
+or being sent to the maintainer.
```
gktrk: Just want to clarify that this change removes the IRC out of the equation. The previous version allows you to establish inactivity from the point when the developer files a bug, sends an email, or messages the maintainer on IRC. This wording requires the developer to come up with a patch before they can establish inactivity. In other words, I can't anymore say "hey, this is broken. the bug has been open for X amount of time. it's blocking my work. can you please fix it?", wait for 2 weeks, and fix it myself. Yes, if I fix it myself, inevitably I'll have to come up with that patch, but not until the timeout happens according to the previous version of the paragraph.

Are we on the same page with the implications of this change? Other than that, everything looks fine.

gktrk:
@ulm see my comment in the approval before proceeding. as long as we understand the implications of the change, I'm not opposed to it being worded this way.

general-concepts/package-maintainers/text.xml
mgorny: Yes, that's the idea. I want to avoid basically the situation when someone reports a hard issue, waits a month, then attaches a really bad patch and commits it without giving the developer time to review. I mean, I've seen that happen.

I don't mean we should let developers block something forever waiting for the perfect patch. But we shouldn't conflate inactivity with inability to find a solution. It's one thing when developer doesn't review suggested fixes. It's another when he rejects them (and in that case you ought to appeal rather than claim maintainer timeout).

PR #105, opened 2019-09-06
eclass-writing: Make even more explicit, per gentoo-dev

mgorny:
Change 'should' into 'must' for initial reviews.  Make it clear
that per-package eclass exception applies only to updates.

@Whissi, is this clear enough for you?


gktrk:
Merged in https://github.com/gentoo/devmanual.gentoo.org/compare/c87c36a...9f63cc9. Thanks!

I put the ML link in the commit message for reference.

PR #106, opened 2019-09-06
general-concepts/dependencies: Fix ranged dependencies example

Whissi:
The error should be obvious: Text is saying 

> To specify "version **2.x** (not 1.x or 3.x)" of a package

but the example would pin version to 1.2.x.

### DIFF
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -195,7 +195,7 @@ the asterisk postfix. This is most commonly seen in situations like:
 </p>
 
 <codesample lang="ebuild">
-DEPEND="gtk? ( =x11-libs/gtk+-1.2* )"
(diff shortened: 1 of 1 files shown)


ulm:
Isn't that a bad example anyway? One would use slot dependencies for `gtk+`.

Whissi:
Well, I also thought about changing example to something like dev-libs/openssl... but this example will be outdated in 2020 when we hopefully are able to remove EOL openssl version so I kept using gtk+.

Do you know a better example package?

ulm:
Not sure. It used to be common with dev-lang/perl.

gktrk:
Given

```
sys-process/latencytop/latencytop-0.5.ebuild:	gtk? ( =x11-libs/gtk+-2* )
```

exists, I suggest this as is and worry about a better example later. What do you guys think?

PR #107, opened 2019-09-08
general-concepts/licenses: Use sections instead of subsections

mgorny:

// Since this is a trivial change, I've split it out of my original PR.


ulm:
Merged.

PR #110, opened 2019-10-06
Updates to mirror-related information

mgorny:
* discourage `RESTRICT=primaryuri`
* replace files via SRC_URI arrows instead of distfiles-local voodoo
* make it clearer that thirdpartymirrors should be used only if multiple ebuilds reference the same mirror group
* explain when thirdpartymirrors make sense, and when they're just a maintenance burden

### DIFF
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -399,22 +399,25 @@ SRC_URI="https://example.com/files/${PV}.tar.gz -> ${P}.tar.gz"
 
 
 <subsection>
-  <title>Third-party mirrors</title>
-  <body>
-    <p>
-      If the items in <c>SRC_URI</c> are available on multiple
-      third-party mirrors, then you don't have to list each mirror in
-      your ebuild. The <c>profiles/thirdpartymirrors</c> file in the
-      <c>::gentoo</c> repository contains named groups of mirrors,
-      accessible through the <c>mirror://</c> pseudo-protocol.
-    </p>
-    <p>
-      As of EAPI 7, the format of <uri
-      link="https://projects.gentoo.org/pms/7/pms.html#x1-340004.4.2">the
-      thirdpartymirrors file</uri> is described in section 4.4.2 of
-      the Package Manager Specification (PMS). One might define a set
-      of &quot;example&quot; mirrors,
-    </p>
+<title>Third-party mirrors</title>
+<body>
+
+<p>
+If the items in <c>SRC_URI</c> are available on multiple third-party mirrors,
+then you don't have to list each mirror in your ebuild.
+The <c>profiles/thirdpartymirrors</c> file in the <c>::gentoo</c> repository
+contains named groups of mirrors, accessible through the <c>mirror://</c>
+pseudo-protocol.
+</p>
+
+<p>
+As of EAPI 7, the format of <uri
+link="https://projects.gentoo.org/pms/7/pms.html#x1-340004.4.2">the
+thirdpartymirrors file</uri> is described in section 4.4.2 of the Package
+Manager Specification (PMS). One might define a set of &quot;example&quot;
+mirrors,
+</p>
+
 <!--
   The following isn't ebuild code, but lang="ebuild" works, and no
   other language choice would be accurate either.
@@ -422,25 +425,26 @@ SRC_URI="https://example.com/files/${PV}.tar.gz -> ${P}.tar.gz"
 <codesample lang="ebuild">
 example https://download.example.com https://mirror1.example.org/example
 </codesample>
-    <p>
-      that can afterwards be referenced through a <c>mirror://</c>
-      URL:
-    </p>
+
+<p>
+that can afterwards be referenced through a <c>mirror://</c> URL:
+</p>
+
 <codesample lang="ebuild">
 SRC_URI="mirror://example/${PN}/${P}.tar.gz"
 </codesample>
-    <warning>
-      If the sources for a package are neither mirror-restricted nor
-      fetch-restricted, then they will be <uri
-      link="::general-concepts/mirrors">mirrored onto Gentoo
-      infrastructure</uri> automatically; in that case, there is
-      little benefit to using this feature. There is a maintenance
-      burden associated with using a third-party mirror list: it must
-      be updated when new mirrors are born and when old mirrors
-      die. Please consider that when deciding whether or not to use
-      this feature.
-    </warning>
-  </body>
+
+<warning>
+If the sources for a package are neither mirror-restricted nor
+fetch-restricted, then they will be <uri
+link="::general-concepts/mirrors">mirrored onto Gentoo infrastructure</uri>
+automatically; in that case, there is little benefit to using this feature.
+There is a maintenance burden associated with using a third-party mirror list:
+it must be updated when new mirrors are born and when old mirrors die. Please
+consider that when deciding whether or not to use this feature.
+</warning>
+
+</body>
 </subsection>
 </section>
 
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -420,7 +420,7 @@ SRC_URI="https://example.com/files/${PV}.tar.gz -> ${P}.tar.gz"
   other language choice would be accurate either.
 -->
 <codesample lang="ebuild">
-example http://download.example.org/ ftp://ftp.example.org/
+example https://download.example.com https://mirror1.example.org/example
 </codesample>
     <p>
       that can afterwards be referenced through a <c>mirror://</c>
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -435,15 +435,28 @@ that can afterwards be referenced through a <c>mirror://</c> URL:
 SRC_URI="mirror://example/${PN}/${P}.tar.gz"
 </codesample>
 
-<warning>
-If the sources for a package are neither mirror-restricted nor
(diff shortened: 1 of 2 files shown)


ebuild-writing/variables/text.xml
```diff
@@ -399,48 +399,66 @@ SRC_URI="https://example.com/files/${PV}.tar.gz -> ${P}.tar.gz"
...
+mirrors,
+</p>
+
 <!--
   The following isn't ebuild code, but lang="ebuild" works, and no
   other language choice would be accurate either.
 -->
 <codesample lang="ebuild">
-example http://download.example.org/ ftp://ftp.example.org/
+example https://download.example.org/ https://mirror1.example.org/
```
Whissi: How about improving the example like
```
example https://download.example.org/releases https://mirror.example.org/
```
which will indicate how to deal with slashes when mirror URL contains a path.

general-concepts/mirrors/text.xml
```diff
@@ -91,8 +98,14 @@ requirement of the license.
 
 <p>
 When you upload the file to <c>dev.gentoo.org:~/public_html</c>, you must ensure that it, and its
-parent directories, are world-readable.
+parent directories, are world-readable.  An example <c>SRC_URI</c> referencing
+a distfile mirrored this way follows:
 </p>
+
+<codesample lang="ebuild">
+SRC_URI="https://dev.gentoo.org/~myname/${P}.tar.gz"
```
Whissi: I would suggest a better example, like
```
<codesample lang="ebuild">
SRC_URI="https://dev.gentoo.org/~myname/distfiles/${PN}/${P}.tar.gz"
</codesample>
```
Because we don't want to end up re-creating the same problem /distfiles have (single folder containing too many files). It will also help developer to cleanup files because files are logically grouped.

ulm:
> Not sure about it. Current devmanual describes a use case. I remember that before I joined Mozilla project, firefox-bin packages had `RESTRICT=mirror` set just to allow upstream to count downloads (I removed this restriction because upstream stopped counting since they switched to CDN but...).

Well, either the file is redistributable, then counting downloads won't provide much useful information anyway. Or it is not, then the ebuild needs `RESTRICT="mirror"` and not `RESTRICT="primaryuri"`. I just don't see why we should impose such restrictions on our users (think about local mirrors which will reduce their external traffic), when we're allowed to mirror the file.

gktrk:
I use `primaryuri` for the ebuilds in my overlays, so that it doesn't pointlessly probe gentoo mirrors for files that obviously won't be there.

ebuild-writing/variables/text.xml
```diff
@@ -435,15 +435,28 @@ that can afterwards be referenced through a <c>mirror://</c> URL:
...
-If the sources for a package are neither mirror-restricted nor
-fetch-restricted, then they will be <uri
-link="::general-concepts/mirrors">mirrored onto Gentoo infrastructure</uri>
-automatically; in that case, there is little benefit to using this feature.
-There is a maintenance burden associated with using a third-party mirror list:
-it must be updated when new mirrors are born and when old mirrors die. Please
-consider that when deciding whether or not to use this feature.
-</warning>
+<p>
+There are two valid cases for using <c>thirdpartymirrors</c>:
```
orlitzky: I tried to make a change like this before and it got rejected because it goes against what's currently done in ::gentoo. Needless to say, I prefer the stronger language.

ebuild-writing/variables/text.xml
```diff
@@ -399,48 +399,52 @@ SRC_URI="https://example.com/files/${PV}.tar.gz -> ${P}.tar.gz"
 
 
 <subsection>
-  <title>Third-party mirrors</title>
```
orlitzky: Sorry, that was me. Nobody told me to RTFM and format it properly.

orlitzky: > It's still useful in new overlay ebuilds, but your version is a lot better than what's there.

That comment was intended for the primaryuri commit (https://github.com/gentoo/devmanual.gentoo.org/pull/110/commits/25c4c71ff20d81c5af7fbd54a83e9907e8f8fcbd). Oops.

mgorny:
> I use `primaryuri` for the ebuilds in my overlays, so that it doesn't pointlessly probe gentoo mirrors for files that obviously won't be there.

Do we [want to] document use in overlays in devmanual?

ulm:
> > I use `primaryuri` for the ebuilds in my overlays, so that it doesn't pointlessly probe gentoo mirrors for files that obviously won't be there.
> 
> Do we [want to] document use in overlays in devmanual?

Even if we would, I think that `primaryuri` is the wrong token. Overlays should use `mirror` restriction for files that are not on mirrors. (If `SRC_URI` fails, falling back to mirrors makes no sense.)

ebuild-writing/variables/text.xml
mgorny: Well, technically slashes are optional. Not sure if we want them or not. I suppose not having slashes will clearly indicate that they're optional.

mgorny:
@gktrk, addressed.

ebuild-writing/variables/text.xml
```diff
@@ -399,48 +399,66 @@ SRC_URI="https://example.com/files/${PV}.tar.gz -> ${P}.tar.gz"
...
+mirrors,
+</p>
+
 <!--
   The following isn't ebuild code, but lang="ebuild" works, and no
   other language choice would be accurate either.
 -->
 <codesample lang="ebuild">
-example http://download.example.org/ ftp://ftp.example.org/
+example https://download.example.com https://mirror1.example.org/example
```
ulm: Please use the canonical form of URLs, i.e. with a trailing slash.

ebuild-writing/variables/text.xml
```diff
@@ -399,48 +399,66 @@ SRC_URI="https://example.com/files/${PV}.tar.gz -> ${P}.tar.gz"
...
+    dealing with upstreams that distribute their distfiles via a network
+    of mirrors without a primary download location or a bouncer service.
+  </li>
+</ol>
+
+<p>
+In any other case, the primary location must be used instead. The distfiles
+will be <uri link="::general-concepts/mirrors">mirrored onto Gentoo
+infrastructure</uri>; in that case, the benefit to using third-party mirror
+list does not outweigh the burden of maintaining it.
```
gktrk: I'm confused by what is meant in this part. If the purpose of this paragraph is to communicate the existence of the infra mirroring system, this is too quick and vague of a transition to talk about that. For example, the following makes for sense to me:

"""
If no mirroring opportunities are available, sources must be listed by their primary location in <c>SRC_URI</c> instead. These files will automatically be mirrored onto the Gentoo Infrastructure when applicable, which then gets replicated across many community source mirrors across the world. The Infra Project is in charge of maintaining the list of community mirrors. This process is transparent to the developers and no steps are needed to opt-in.
"""

ebuild-writing/variables/text.xml
```diff
@@ -399,48 +399,66 @@ SRC_URI="https://example.com/files/${PV}.tar.gz -> ${P}.tar.gz"
...
-    </warning>
-  </body>
+
+<p>
+There are two valid cases for using <c>thirdpartymirrors</c>:
+</p>
+
+<ol>
+  <li>
+    providing multiple download locations for mirror- or fetch-restricted
```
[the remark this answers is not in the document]
mgorny: Yes, fetch-restricted is wrong here. However, mirror restriction implies *we* can't mirror it. It doesn't mean upstream can't have multiple download locations.

ebuild-writing/variables/text.xml
```diff
@@ -399,48 +399,66 @@ SRC_URI="https://example.com/files/${PV}.tar.gz -> ${P}.tar.gz"
...
-example http://download.example.org/ ftp://ftp.example.org/
+example https://download.example.com https://mirror1.example.org/example
 </codesample>
-    <p>
-      that can afterwards be referenced through a <c>mirror://</c>
-      URL:
-    </p>
+
+<p>
+that can afterwards be referenced through a <c>mirror://</c> URL:
```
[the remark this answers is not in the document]
mgorny: I don't know. I can change it.

ebuild-writing/variables/text.xml
mgorny: No, it's meant to supposed that otherwise you're not supposed to use `mirror://` even if there are 'opportunities' but just list the primary URI and let mirrors take care of mirroring.

mgorny: I've taken part of your sentence and integrated it into the basic idea. Not sure if it doesn't suggest that other files wouldn't be mirrored though.

mgorny:
You know what? I'm done with this. I've spent four hours merging your requests, reformatting, rebasing to apply it to the right commits. You just keep demanding more and more, and now you even want me to change stuff that wasn't part of my original patch.

Devmanual is in shitty state. Nobody wants to contribute. Whoever tries, quickly gets demotivated. Look at it from my perspective. I spend a lot of work to improve stuff. I wait. I hear lot of demands on minor changes that take three times as much of my time than if you did them yourself. Then I wait again. And there are more demands. Then the other developer tells me to do it the other way. Then I realize I can't even merge the requested changes back because they're totally outside my commits.

If you want devmanual to improve, please apply your suggestions yourself. Either do it on top of my commits, or merge and fix them afterwards, or squash them, or whatever. If you don't want to do the work, then just please close it and feel free to close the project since there's really no point in having devmanual developers who don't work on devmanual but only focus on demotivating others from contributing.

PR #112, opened 2019-10-16
general-concepts/mirrors: improve bad upstream distfile rename text

gktrk:
If an upstream updates a distfile in a way that requires an ebuild
revision bump, remind readers to remove the ebuilds associated with
the old distfile. This is necessary because the Manifest entry for the
old tarball has its checksum changed due to upstream and cause
checksum mismatches during merges.

### DIFF
--- a/general-concepts/mirrors/text.xml
+++ b/general-concepts/mirrors/text.xml
@@ -66,8 +66,13 @@ fetch and start distributing the new version.
 </p>
 
 <p>
-Please note that if upstream made any changes affecting the built package,
-you need to also bump the ebuild's revision.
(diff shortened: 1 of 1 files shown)


gktrk:
> Don't you need to do this regardless of whether or not you create a new revision? If -r1, -r2, and -r3 all point to the same tarball and if upstream modifies it in-place, then you can't just fix the -r3 ebuild, right?

That sounds right. I'll rewrite to unconditionally remove old files.

gktrk:
@orlitzky how about this?

gktrk:
@orlitzky this last push should do it, take a look

orlitzky:
Even better.

PR #113, opened 2019-10-23
ebuild-maintenance/git: add sign-off for Certificate of Origin

wimmuskee:
Not sure if https://bugs.gentoo.org/668686 can be closed after this.

### DIFF
--- a/ebuild-maintenance/git/text.xml
+++ b/ebuild-maintenance/git/text.xml
@@ -141,7 +141,11 @@ rebase</c>.</li>
 Whenever repoman is not used to commit, you need to manually verify all
 packages affected by the commit using <c>repoman full</c>. Since repoman
 is not aware of staged changes, please make sure that all files are included
-in the commit.
+in the commit. Also when not using repoman, you must perform a manual sign-off
+to the <uri link="https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin">
+Certificate of Origin</uri> using the <c>-s</c> or <c>--signoff</c> option
+with your git commit commands. Make sure you have read and understand the
+actual Certificate.
 </p>
 
 </body>
@@ -205,8 +209,9 @@ line. From the third line onward should be a detailed, multiline
 explanation of the changes introduced by the commit. At the very end,
 auxiliary information such as the bugs related to the commit,
 contributors, and reviewers should be listed using RFC822/git style
-tags. The length of the lines in the message should be 70-75
-characters at maximum.
+tags. The sign-off agreement will be added there as well when applied
+with the commit action. The length of the lines in the message should
+be 70-75 characters at maximum.
 </p>
 
 <p>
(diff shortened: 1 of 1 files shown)


ebuild-maintenance/git/text.xml
```diff
@@ -202,11 +202,15 @@ way. Additionally, consistency in message format allows for easier
 parsing by external tools. The first line of the commit message should
 contain a brief summary of the changes, followed by an empty new
 line. From the third line onward should be a detailed, multiline
-explanation of the changes introduced by the commit. At the very end,
-auxiliary information such as the bugs related to the commit,
-contributors, and reviewers should be listed using RFC822/git style
-tags. The length of the lines in the message should be 70-75
-characters at maximum.
+explanation of the changes introduced by the commit. Following that,
```
gktrk: Overall LGTM. The order of things don't naturally appear that way though. See the following example commit:

```
commit b48f5e83b285aa086a844e1e9bd1035d49df122a
Author:     Thomas Deutschmann <whissi@whissi.de>
AuthorDate: Fri Sep 6 16:57:52 2019 +0200
Commit:     Göktürk Yüksek <gokturk@gentoo.org>
CommitDate: Wed Oct 16 14:47:48 2019 -0400

    general-concepts/dependencies: Fix ranged dependencies example
    
    Closes: https://github.com/gentoo/devmanual.gentoo.org/pull/106
    Acked-by: Michael Orlitzky <mjo@gentoo.org>
    Signed-off-by: Thomas Deutschmann <whissi@whissi.de>
```

Because of the way git parses the `--signoff` option, it'll append the `Signed-off-by` after `Closes:`. I'd probably merge the signoff sentences with the sentence that starts with "At the very end,", and also perhaps talk about repoman doing this automatically and passing `--signoff` to git if not using repoman.

If you don't want to edit any further, that's fine. I can make those changes post merge.

wimmuskee: Ok, made it a little different. Put most of the sign-off action at the section dealing with doing commits. Feels more correct that way since that is the point you are actually committing to the sign-off.
At that point, the information about the sign-off in the commit message functions more as a verification for the reader. And to finish it up, added the line in the example.

ebuild-maintenance/git/text.xml
```diff
@@ -141,7 +141,10 @@ rebase</c>.</li>
 Whenever repoman is not used to commit, you need to manually verify all
 packages affected by the commit using <c>repoman full</c>. Since repoman
 is not aware of staged changes, please make sure that all files are included
-in the commit.
+in the commit. Also when not using repoman, you must perform a manual sign-off
+to the <uri link="https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin">
```
mgorny: One thing I'd change is to make it more verbose that people are supposed to read what they're signing before signing it.

wimmuskee: you mean something like "make sure you have read the actual certificate", or more something that underlines the legal side of signing a commit

mgorny: The former. We don't want to get into a legal debate.

wimmuskee: added a line

PR #114, opened 2019-10-31
ebuild-writing/file-format: discourage _pre for snapshots

gktrk:
Do not discourage people to use the format _preYYYYMMDD when creating
snapshot ebuilds. This requires the developer to guess the version of
the next release by the upstream. The same snapshot can be treated as
a patch release over an existing version with the _pYYYYMMDD format,
which is more reliable.

### DIFF
--- a/ebuild-writing/file-format/text.xml
+++ b/ebuild-writing/file-format/text.xml
@@ -115,11 +115,14 @@ Overall, this gives us a filename like <c>libfoo-1.2.5b_pre5-r2.ebuild</c>.
 </p>
 
 <p>
-When packaging a snapshot of a source repository, there are two commonly used formats. The first
-treats the snapshot as a patch to the previous version, and so the ebuild version is in the format
-$(last-released-version)_pYYYYMMDD. Alternatively, the snapshot may be treated as a pre-release to
-an upcoming version, usually used when a release is anticipated but not out yet. The format for this
-is $(upcoming-version)_preYYYYMMDD.
+When packaging a snapshot of a source repository, treat the snapshot
+as a patch to the previous version, and so the ebuild version is in
+the format $(last-released-version)_pYYYYMMDD. Treating the snapshot
+as a pre-release to an upcoming version is strongly discouraged, as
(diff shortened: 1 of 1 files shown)


ebuild-writing/file-format/text.xml
```diff
@@ -115,11 +115,14 @@ Overall, this gives us a filename like <c>libfoo-1.2.5b_pre5-r2.ebuild</c>.
 </p>
 
 <p>
-When packaging a snapshot of a source repository, there are two commonly used formats. The first
-treats the snapshot as a patch to the previous version, and so the ebuild version is in the format
-$(last-released-version)_pYYYYMMDD. Alternatively, the snapshot may be treated as a pre-release to
-an upcoming version, usually used when a release is anticipated but not out yet. The format for this
-is $(upcoming-version)_preYYYYMMDD.
+When packaging a snapshot of a source repository, treat the snapshot
+as a patch to the previous version, and so the ebuild version is in
```
orlitzky: The "and so" sounds weird here now that you changed the sentence to an imperative ("treat...") one.

ulm: This isn't an improvement of the existing policy. As mgorny said, the correct procedure is to check the version in the upstream repository, and if there was no release of that version, use the `_pre` variant.

As an example, GNU Emacs uses version 27.0.50 in its `master` branch, but its last release (26.3) was made from the `emacs-26` branch. It would make no sense (and would be confusing) if I were to name a snapshot from the `master` branch 26.3_pYYYYMMDD.

So please drop this patch.

ulm:
This isn't an improvement of the existing policy. As mgorny said, the correct procedure is to check the version in the upstream repository, and if there was no release of that version, use the `_pre` variant.

As an example, GNU Emacs uses version 27.0.50 in its `master` branch, but its last release (26.3) was made from the `emacs-26` branch. It would make no sense (and would be confusing) if I were to name a snapshot from the `master` branch 26.3_pYYYYMMDD.

So please drop this patch.

PR #115, opened 2019-11-03
HOMEPAGE: multi-valued, a bit of cleanup

mgorny:
Better indicate that HOMEPAGE accepts multiple values, and clean up the paragraph while at it. See commit messages for details.

### DIFF
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -185,11 +185,12 @@ The following variables may or must be defined by every ebuild.
   <tr>
     <ti><c>HOMEPAGE</c></ti>
     <ti>
-    Package's homepage. Mandatory (except for virtuals). If you are
-    unable to locate an official one, try to provide a link to
-    <uri link="http://freshmeat.net">freshmeat.net</uri> or a similar
-    package tracking site. If the homepage of the package is unavailable
-    and no new location can be found, please update it to
+    Package's homepage(s). Mandatory (except for virtuals), accepts
+    multiple values. In some contexts, it is customary to provide
+    package tracker and/or code hosting links besides the main homepage
+    (e.g. PyPI link for Python packages, GitHub link when code can't
+    easily be found on homepage) for convenience. If no homepage for
+    the package is available, please set it to
     <c>https://wiki.gentoo.org/wiki/No_homepage</c>.
     Never refer to a variable name in the string; include only raw text.
     </ti>
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -700,6 +700,11 @@ have no direct correlation with the expected string should be avoided.
   </li>
 
   <li>
+    using <c>${HOMEPAGE}</c> in <c>SRC_URI</c> — it breaks when homepage
+    changes or additional entry is added,
+  </li>
+
+  <li>
     using <c>${PN}</c> (or other unnecessary helper variables)
     excessively in <c>src_install()</c> — it impairs readability for
     little benefit and causes a lot of trouble when the package needs to
--- a/quickstart/text.xml
+++ b/quickstart/text.xml
@@ -40,7 +40,7 @@ can see real ebuilds in the main tree).
 EAPI=6
 
 DESCRIPTION="Exuberant ctags generates tags files for quick source navigation"
-HOMEPAGE="http://ctags.sourceforge.net"
+HOMEPAGE="https://ctags.io/ https://github.com/universal-ctags/ctags"
 SRC_URI="mirror://sourceforge/ctags/${P}.tar.gz"
 
 LICENSE="GPL-2+"


ebuild-writing/variables/text.xml
```diff
@@ -185,11 +185,12 @@ The following variables may or must be defined by every ebuild.
...
     <ti>
-    Package's homepage. Mandatory (except for virtuals). If you are
-    unable to locate an official one, try to provide a link to
-    <uri link="http://freshmeat.net">freshmeat.net</uri> or a similar
-    package tracking site. If the homepage of the package is unavailable
-    and no new location can be found, please update it to
+    Package's homepage(s). Mandatory (except for virtuals), accepts
+    multiple values. In some contexts, it is customary to provide
+    package tracker and/or code hosting links besides the main homepage
+    (e.g. PyPI link for Python packages, GitHub link when code can't
```
sbraz: I thought we weren't adding PyPI links in HOMEPAGE, but rather leaving them in metadata's upstream section.

ebuild-writing/variables/text.xml
```diff
@@ -698,6 +699,11 @@ have no direct correlation with the expected string should be avoided.
     (e.g. while reviewing via gitweb),
   </li>
 
+  <li>
+    using <c>${HOMEPAGE}</c> in <c>SRC_URI</c> — it breaks when homepage
+    changes or additional entry is added,
```
sbraz: Maybe "additional entries are added"?

gktrk:
> `HOMEPAGE` even supports USE conditionals, although I'm not sure we should tell people to do it.
> 
> You already know that I disagree about the `HOMEPAGE` in `SRC_URI` issue. The following is incredibly common:
> 
> ```shell
> HOMEPAGE="https://github.com/${PN}/${PN}"
> SRC_URI="${HOMEPAGE}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
> ```
> 
> and I don't see the point of manually inserting all of those values in there so long as our package format is a shell script. Yes, it will break if `HOMEPAGE` changes, but that's true of every shell variable -- you have to check the "call sites" when you change a variable's value. And those two variables are right next to each other, which makes it relatively easy to do so. Finally, the resulting breakage is something that repoman will catch, so no bad things happen if you forget.
> 
> I think shell script is a bad package format, but it's what we've got. Disallowing this one specific string substitution when e.g. using DEPEND inside RDEPEND is far more dangerous just feels wrong to me.

I would argue that HOMEPAGE is usually parsed by humans, unlike DEPEND or RDEPEND. I don't know a case where the package manager actually consumes HOMEPAGE beyond printing stuff on the console. Assuming this is the case, I would support discouraging variables in HOMEPAGE as well. It just makes copy paste so damn hard, which is pretty much the only use of it AFAIK.

orlitzky:
Everything inside an ebuild file is meant to be parsed by a human: it's source code. It's not like you can copy-paste `src_prepare()` out into a separate file and run it through bash to see if it works.

But the fact that it's parsed by a human is precisely why variable substitutions are helpful. The reason we do things like putting `DEPEND` inside `RDEPEND` is because sometimes, the runtime dependencies are a superset of the build dependencies. And in cases like that, it's easier to see that

```bash
DEPEND="dev-libs/libltdl:0
	virtual/mailx
	perl? ( dev-lang/perl:= )
	web? (
		app-arch/unzip
		media-libs/gd[jpeg,png]
		lighttpd? ( www-servers/lighttpd[php] )
		apache2? (
			|| (
				>=www-servers/apache-2.4[${MOD_ALIAS},apache2_modules_cgi]
				>=www-servers/apache-2.4[${MOD_ALIAS},apache2_modules_cgid]
				>=www-servers/apache-2.4[${MOD_ALIAS},apache2_modules_fcgid] )
			|| (
				dev-lang/php:*[apache2]
				dev-lang/php:*[cgi]
				dev-lang/php:*[fpm] )
		)
	)"

RDEPEND="${DEPEND}
	vim-syntax? ( app-vim/nagios-syntax )"
```

is correct than the alternative. Quick, did I cut out any build dependencies below for the sake of illustration?

```bash
DEPEND="dev-libs/libltdl:0
	virtual/mailx
	perl? ( dev-lang/perl:= )
	web? (
		app-arch/unzip
		media-libs/gd[jpeg,png]
		lighttpd? ( www-servers/lighttpd[php] )
		apache2? (
			|| (
				>=www-servers/apache-2.4[${MOD_ALIAS},apache2_modules_cgi]
				>=www-servers/apache-2.4[${MOD_ALIAS},apache2_modules_cgid]
				>=www-servers/apache-2.4[${MOD_ALIAS},apache2_modules_fcgid] )
			|| (
				dev-lang/php:*[apache2]
				dev-lang/php:*[cgi]
				dev-lang/php:*[fpm] )
		)
	)"

RDEPEND="dev-libs/libltdl:0
	virtual/mailx
	perl? ( dev-lang/perl:= )
	web? (
		app-arch/unzip
		media-libs/gd[jpeg,png]
		lighttpd? ( www-servers/lighttpd[php] )
		apache2? (
			|| (
				>=www-servers/apache-2.4[${MOD_ALIAS},apache2_modules_cgi]
				>=www-servers/apache-2.4[${MOD_ALIAS},apache2_modules_cgid]
				>=www-servers/apache-2.4[${MOD_ALIAS},apache2_modules_fcgid] )
			|| (
				dev-lang/php:*[apache2]
				dev-lang/php:*[cgi]
				dev-lang/php:*[fpm] )
		)
	)
	vim-syntax? ( app-vim/nagios-syntax )"
```

It's not obvious if the two agree. It's less error-prone (and just plain easier to read) if you say what you mean rather than trying to reimplement portage inside the developer.

`HOMEPAGE` is the same way. Semantically, github.com-slash-the-name-of-the-package is the homepage for things that live only on github, and you should be able to say that rather than having to constantly stare are three identical URLs to see if they're the same.

For packages that have already been committed, you should be using the package manager to get the homepage. Michał is mainly interested in this because he winds up reviewing code that he didn't write, at a glance, through github. I don't want to make that any harder so I'm reluctant to start (yet another) argument about it, but I don't think that use case should inform tree-wide policy for the rest of us.

mgorny:
Please tell me that you've never ever used the URL from ebuild seen via editor/grep/whatever and always used some kind of API frontend to process ebuilds.

orlitzky:
For my own packages, I test the homepage once through the package manager to make sure that portage spits out what I think it will, and then I forget about it. For other peoples' packages, `eix foo` returns an answer immediately -- but that only works for things in the tree.

gktrk:
> Please tell me that you've never ever used the URL from ebuild seen via editor/grep/whatever and always used some kind of API frontend to process ebuilds.

That's precisely my reasoning too. This was especially helpful when reviewing proxy-maintained packages. While I'm reviewing an ebuild, I wanna be able to quickly check if they got the home page right. Or I want to navigate to the homepage to get to their VCS repository. It sucks having to copy the URL and replace by hand the variables used in there.

gktrk:
> For my own packages, I test the homepage once through the package manager to make sure that portage spits out what I think it will, and then I forget about it. For other peoples' packages, `eix foo` returns an answer immediately -- but that only works for things in the tree.

Exactly, eix doesn't work when you are reviewing somebody else's work.

orlitzky:
But there are 36,000 ebuilds in the tree that will never be reviewed on github again, and you're making life harder for the people who maintain them for this one use case. (They shouldn't be reviewed on github at all, social contract and whatnot.)

"Eyeballing" the `HOMEPAGE` is only good for cursory reviews in the first place. Proxy maintainers don't just make mistakes in `HOMEPAGE`, they do it in every variable. And the only way to see the result of expanding every other variable is to use the package manager. So at some point you're going to have to wget the patch, apply it, and run repoman on it at the very least. When you do that, it's easy to get portage to spit out the value of `HOMEPAGE` even if you don't use eix. For example, I figured this out in about 30 seconds:

```bash
$ ebuild foo.ebuild setup --debug | grep HOMEPAGE
```

Just like you have to do with every other variable. And while this isn't the place to complain about it -- so long as this isn't a real QA policy, commits like these are just a waste of both of our time:


```patch
commit 3a39148aed0a13a8c97f098f58bbc96c0e68817f
Author: Michał Górny <mgorny@gentoo.org>
Date:   Thu Nov 7 17:15:11 2019 +0100

    app-antivirus/clamav-unofficial-sigs: [QA] Fix HOMEPAGE & SRC_URI

diff --git a/app-antivirus/clamav-unofficial-sigs/clamav-unofficial-sigs-6.0.1-r2.ebuild b/app-antivirus/clamav-unofficial-sigs/clamav-unofficial-sigs-6.0.1-r2.ebuild
index eb1fdfbcc04..694d3dc7418 100644
--- a/app-antivirus/clamav-unofficial-sigs/clamav-unofficial-sigs-6.0.1-r2.ebuild
+++ b/app-antivirus/clamav-unofficial-sigs/clamav-unofficial-sigs-6.0.1-r2.ebuild
@@ -6,8 +6,8 @@ EAPI=7
 inherit user systemd

 DESCRIPTION="Download and install third-party clamav signatures"
-HOMEPAGE="https://github.com/extremeshok/${PN}"
-SRC_URI="${HOMEPAGE}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+HOMEPAGE="https://github.com/extremeshok/clamav-unofficial-sigs"
+SRC_URI="https://github.com/extremeshok/clamav-unofficial-sigs/archive/${PV}.tar.gz -> ${P}.tar.gz"

 LICENSE="BSD"
 SLOT="0"
```

How does fucking with my package in ::gentoo help you review things on github?

mgorny:
> But there are 36,000 ebuilds in the tree that will never be reviewed on github again, and you're making life harder for the people who maintain them for this one use case. (They shouldn't be reviewed on github at all, social contract and whatnot.)

What makes you think they will not? Why are you limiting everyone's usage to your own usage? I frequently click HOMEPAGE links when I have the ebuild open. I know I can use super complex tools to achieve the same result. This doesn't mean it's the right thing to do.

> "Eyeballing" the `HOMEPAGE` is only good for cursory reviews in the first place. Proxy maintainers don't just make mistakes in `HOMEPAGE`, they do it in every variable. And the only way to see the result of expanding every other variable is to use the package manager. So at some point you're going to have to wget the patch, apply it, and run repoman on it at the very least. When you do that, it's easy to get portage to spit out the value of `HOMEPAGE` even if you don't use eix. For example, I figured this out in about 30 seconds:
> 
> ```shell
> $ ebuild foo.ebuild setup --debug | grep HOMEPAGE
> ```

That's great. You just wasted 30 seconds on something I could do in 2 seconds by clicking the link. Plus, you're entirely missing the corner cases such as the ebuild being already unpacked somewhere, causing ebuild to fail with permission errors.

> 
> Just like you have to do with every other variable. And while this isn't the place to complain about it -- so long as this isn't a real QA policy, commits like these are just a waste of both of our time:

Is it an imaginary policy? Because it's been what I was taught back when I joined Gentoo, it's what's enforced by gentoo-syntax packages. It's as close to a de-facto policy as I see it. Just because some developers believe they're special and they get to do bad stuff for the fun of it doesn't make it 'not a real policy'. Surely, I can ask QA to do another vote just to prove the point because we obviously don't have anything better to do with our lives.

> How does fucking with my package in ::gentoo help you review things on github?

That is not the language that encourages healthy collaboration.

orlitzky:
> What makes you think they will not? Why are you limiting everyone's usage to your own usage?

I'm not. You're the one proposing a patch to make everyone fit *your* workflow. I'm saying: leave me alone, it's better the way it is, but if you want to waste your own time copy/pasting long strings from short variables into your own ebuilds, then go right ahead. If you want to make it a proxy-maint policy, go right ahead.

> That's great. You just wasted 30 seconds on something I could do in 2 seconds by clicking the link. 

Next time it will also take me two seconds. Twenty-eight of them were spent reading a man page.

 > Surely, I can ask QA to do another vote just to prove the point because we obviously don't have anything better to do with our lives.

Nothing is wrong here, these packages are all fine; and yes, clearly you don't have anything better to do. If you want me to go out of my way to make my code look worse and be harder to maintain, then yeah, make it a real policy.

> That is not the language that encourages healthy collaboration.

If you're interested in healthy collaboration, don't edit peoples' packages behind their backs. Otherwise, get called names =P

gktrk:
@orlitzky your example of including `DEPEND` in `RDEPEND` is false equivalency though. Including DEPEND in RDEPEND or replicating its contents directly in RDEPEND expand to the same string, and the package manager will parse either just fine. But the developer isn't simply trying to list package for RDEPEND right, it's an expression of the idea that packages in DEPEND are also needed in RDEPEND. It contains more information that is useful to the human reading it. It also eases maintenance because you don't have to keep two related lists in sync.

The same cannot be said for HOMEPAGE. First of all, it isn't processed by the package manager in a way that affects the merge process. So we'd have to ask ourselves the intent of a developer who uses ${PN} in HOMEPAGE, as opposed to no variables. Does it express more like the case in DEPEND? Is there any meaningful information to be gained by knowing that part of the HOMEPAGE is made of the package's name? With DEPEND in RDEPEND, it is telling you that DEPEND is a subset of RDEPEND, and you can use that information in useful ways during development. I see no equivalent functionality for HOMEPAGE containing variables. HOMEPAGE is a string that happens to contain a substring that matches the package name, not a function of it.

orlitzky:
@gktrk you're anthropomorphizing bash. Like I said in my first comment: shell script is a stupid package format, but it's what we've got. Putting `DEPEND` inside `RDEPEND` is string concatenation, nothing more. Any other meaning you assign to it is in your head. `HOMEPAGE` is no different in that regard.

Sometimes it's convenient to use variables when it makes your life easier. Putting `DEPEND` into `RDEPEND` sure makes it easier to see what's going on, at the cost of having to stop and think when you make changes to `DEPEND`. And yeah, it makes it easier to keep them in sync. But how is `HOMEPAGE` any different? Every github package has the same format:

```bash
HOMEPAGE="https://github.com/${GITHUB_USER}/${PN}"
SRC_URI="${HOMEPAGE}/archive/${PV}.tar.gz -> ${P}.tar.gz"
```

This is obviously correct, under 80 characters, doesn't need to be kept in sync, and works for hundreds of packages. After Michał's edit, we get,

```bash
HOMEPAGE="https://github.com/extremeshok/clamav-unofficial-sigs"
SRC_URI="https://github.com/extremeshok/clamav-unofficial-sigs/archive/${PV}.tar.gz -> ${P}.tar.gz"
```

which works for one package, causes word-wrap, and can easily fall out of sync. The first one is easier to maintain, if we just admit that out package format is shell script and move on with our lives. Adding a million tiny annoying rules has a real cost in complexity, which I'm sure you know from curating the decades-old bad advice that the devmanual still suggests in many cases.

If you want to talk about the semantics of using `${PN}` in `HOMEPAGE`, then, even though I think it's a red herring, keep in mind that Github was born as a rails application. If you're familiar with rails, they do semantic routing, and I'd bet you anything that a path like

```bash
HOMEPAGE="https://github.com/extremeshok/clamav-unofficial-sigs"
```

literally gets mapped on their end via a string like

```ruby
get '/:user/:repo'
```

to a user/repo object in the database. Using `${PN}` in that path component is about as close as you can get semantically.

mgorny:
And now imagine you need to do a pkgmove and you've just quadruplicated the work because you thought you're going to save effort by putting more variables in something you don't touch at all during normal bumps.

I suppose you also love things like:

    dobin ${PN}-config
    dobin foo${PN/-/_}bar

ulm:
Why are we discussing usage of variables in `HOMEPAGE` here? Current policy is "Never refer to a variable name in the string; include only raw text", and the PR doesn't touch this.

Besides, I find it rather practical that I can access the homepage if I have the ebuild open in Emacs, simply by pressing `C-c u .` (`browse-url-at-point`) and without refering to external utilities. Also `eix` isn't a proper tool for this, because it will print the homepage of the package (newest ebuild in tree, I guess), and it will neither accept an atom nor an ebuild filename as argument.

As for using `${HOMEPAGE}` for assignment of other variables, I don't have a strong opinion.

orlitzky:
> And now imagine you need to do a pkgmove and you've just quadruplicated the work because you thought you're going to save effort by putting more variables in something you don't touch at all during normal bumps.

If you're lucky and the package was forked to another github maintainer, then in

```bash
GITHUB_USER=orlitzky
HOMEPAGE="https://github.com/${GITHUB_USER}/${PN}"
SRC_URI="${HOMEPAGE}/archive/${PV}.tar.gz -> ${P}.tar.gz"
```

all you have to change is the `GITHUB_USER` variable. Then `SRC_URI` gets set correctly automatically. If you're unlucky and the package is moved somewhere else entirely, then you need to redo both `HOMEPAGE` and `SRC_URI`.

If you copy/paste those variables in there, on the other hand, you **always** need to change **both** lines. Using variables never makes your life harder as far as I can tell.


> 
> I suppose you also love things like:
> 
> ```
> dobin ${PN}-config
> dobin foo${PN/-/_}bar
> ```

It depends. If there's a build system and if the build system uses the name of the package in those contexts, then sure. Otherwise, it's pointless. But either way I wouldn't try to ban the use of bash in bash to prevent people from doing it.

orlitzky:
> Why are we discussing usage of variables in `HOMEPAGE` here? Current policy is "Never refer to a variable name in the string; include only raw text", and the PR doesn't touch this.

Because it's not a policy, and I think it would be stupid to make it one.


> Besides, I find it rather practical that I can access the homepage if I have the ebuild open in Emacs, simply by pressing `C-c u .` (`browse-url-at-point`) and without refering to external utilities. Also `eix` isn't a proper tool for this, because it will print the homepage of the package (newest ebuild in tree, I guess), and it will neither accept an atom nor an ebuild filename as argument.

I get that it's convenient sometimes, and I'd often like to open `SRC_URI` in my browser, too. But that's not reality. All metadata variables except `EAPI` are intended for the package manager. If you want it to work, change the PMS to disallow dependency specifications in `HOMEPAGE`, and disallow variable substitutions as well while you're in there. At least then this won't be unreliable. But for now, just like with `SRC_URI`, it's still easiest and most reliable to get the `HOMEPAGE` with tools. Trying to legislate against reality is more trouble than it's worth.

orlitzky:
Let me offer something constructive instead of just saying that I think this idea is stupid over and over again. The metadata variables `DESCRIPTION` and `HOMEPAGE` rarely ever change, and can't break the package when they do. `HOMEPAGE` doesn't need dependency specifications as far as I can tell, even though the PMS allows it.

[GLEP5](https://www.gentoo.org/glep/glep-0005.html) already proposed moving these into metadata.xml, where there are no variables. This would make you (and me!) happy, since it can be cut/pasted. It would make me happy, because I wouldn't need to copy around the same `HOMEPAGE` a thousand times, and -- if I don't have to look at it -- maintainability isn't an issue.

If we just dump these two variables into `metadata.xml` in EAPI=8, everyone is happy.

ulm:
> > Why are we discussing usage of variables in `HOMEPAGE` here? Current policy is "Never refer to a variable name in the string; include only raw text", and the PR doesn't touch this.
> 
> Because it's not a policy, and I think it would be stupid to make it one.

Of course it is a long established policy, which was mentioned in the ebuild howto in 2004 already:
https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo/xml/htdocs/proj/en/devrel/handbook/hb-guide-ebuild.xml?hideattic=0&revision=1.10&view=markup#l540

orlitzky:
> 
> Of course it is a long established policy, which was mentioned in the ebuild howto in 2004 already:

Being mentioned in some fifteen-year-old webpage (that predates the QA team itself, so it couldn't have been a QA policy) doesn't make it policy.

ulm:
> Being mentioned in some fifteen-year-old webpage (that predates the QA team itself, so it couldn't have been a QA policy) doesn't make it policy.

Well, it was in the ebuild howto since 2004, and it was copied to the devmanual in 2007, so IMHO we can at least call it established best practice.

And of course the QA project already existed back then, it is listed in GLEP 4 (posted in 2003), and QA is in CC of bug 73272 where this was discussed.

orlitzky:
> > Being mentioned in some fifteen-year-old webpage (that predates the QA team itself, so it couldn't have been a QA policy) doesn't make it policy.
> 
> Well, it was in the ebuild howto since 2004, and it was copied to the devmanual in 2007, so IMHO we can at least call it established best practice.

And the only reason I'm still commenting here is because someone shouldn't be able to add a bad idea to a webpage with no oversight and then cite it fifteen years later as if it were gospel.

 
> And of course the QA project already existed back then, it is listed in GLEP 4 (posted in 2003), and QA is in CC of bug 73272 where this was discussed.

Ok, you've got me there. I was going by GLEP48 where QA was given the ability to make and enforce policies.

gktrk:
> > > Being mentioned in some fifteen-year-old webpage (that predates the QA team itself, so it couldn't have been a QA policy) doesn't make it policy.
> > 
> > 
> > Well, it was in the ebuild howto since 2004, and it was copied to the devmanual in 2007, so IMHO we can at least call it established best practice.
> 
> And the only reason I'm still commenting here is because someone shouldn't be able to add a bad idea to a webpage with no oversight and then cite it fifteen years later as if it were gospel.
> 

As ulm pointed out, this appears to be discussed in https://bugs.gentoo.org/73272. It is not one person adding an idea to a page with no oversight. QA is on CC for that bug too.

> > And of course the QA project already existed back then, it is listed in GLEP 4 (posted in 2003), and QA is in CC of bug 73272 where this was discussed.
> 
> Ok, you've got me there. I was going by GLEP48 where QA was given the ability to make and enforce policies.

Given the evidence to suggest that this is at least established best practice and that multiple devs put thought into it back in 2004, I'm gonna go ahead and merge this. Of course there can be opposing opinions but I don't wanna hold this up until it comes to a resolution on Github. If the usefulness of the practice is being challenged, gentoo-dev is a better avenue for discussion on the topic at this point. I don't see the downside of including this patch at the moment.

orlitzky:
I can't stop anyone from adding a non-policy from 2004 to the devmanual, but if you really want to make a change that affects every single ebuild in the tree, then the onus should be on you to take it to -dev (or just get QA to vote on it).

If it's a QA policy, I'll think the people who voted for it are stupid, but I'll shut up and follow it =P

mgorny:
Michael, do you really think the best way to work on Gentoo is to call everyone who doesn't agree with you stupid?

orlitzky:
@mgorny

The "=P" should indicate that I am not entirely serious when I say that.

gktrk:
> I can't stop anyone from adding a non-policy from 2004 to the devmanual, but if you really want to make a change that affects every single ebuild in the tree, then the onus should be on you to take it to -dev (or just get QA to vote on it).
> 
> If it's a QA policy, I'll think the people who voted for it are stupid, but I'll shut up and follow it =P

This was kind of my logic actually. Adding it to the tree doesn't give to or take from the QA any power to enforce this. They have their own policies and in devmanual we try to reflect those in combination with the PMS and the community's best practices. If you take it to gentoo-dev and there is support for it, I'll change devmanual to reflect that too then.

PR #116, opened 2019-11-03
ebuild-writing/users-and-groups: GLEP 81 user data guidelines.

orlitzky:
GLEP 81 significantly changes the way that user management is handled,
and reveals some subtle issues in existing packages that have remained
hidden until now. Many of these issues can be avoided (in GLEP 81, but
also in general) by exercising some discipline when choosing the data
for new users and groups:

  * User and group names
  * Default shell
  * Home directory
  * Home directory ownership and permissions

This commit adds a few new sections to the "Users and Groups" chapter
that explain the current best practices, and give suggestions for
choosing the user and group data listed above. The rationale for these
choices is explained, and some notable exceptions are mentioned.

Closes: https://bugs.gentoo.org/693964
Signed-off-by: Michael Orlitzky <mjo@gentoo.org>

### DIFF
--- a/ebuild-writing/users-and-groups/text.xml
+++ b/ebuild-writing/users-and-groups/text.xml
@@ -103,6 +103,175 @@ The defaults are the user and primary group for the owner, and 0755 for the
 permissions.
 </p>
 
+<important>
+Whenever possible, the default shell and home directory should be
+used. The rationale for this is explained below.
+</important>
+
+<p>
+You should embark upon a GLEP81 migration like an EAPI
+update. Rather than simply copy your user's settings into an
+<c>acct-user</c> package, you should take the opportunity to
+re-evaluate your user's name, shell, home directory, and its
+permissions. Our GLEP 81 implementation will reveal many user
+management issues that were allowed to fester in the past.
+</p>
+
+<subsection>
+<title>Choosing a shell</title>
+<body>
+<p>
+In most cases, the default shell (that is, no shell) should be
+used. Services can still be started as a user who has no shell, and
+daemons are able to drop privileges to a user who has no shell. If
+necessary, the administrator can override a user's default shell with
+<c>su -s &lt;shell&gt; &lt;username&gt;</c>. This is sufficient for
+testing, management of SSH credentials, and for initial configuration
+in an ebuild's <c>pkg_config</c> phase.
+</p>
+<p>
+An obvious exception to this rule is if a human being will need to log
+into the account interactively, as is the case with the <c>root</c>
+user. Other exceptions certainly exist, but should be evaluated
+case-by-case. In other words, if you haven't checked, don't set your
+user's shell to <c>/bin/bash</c> because you think he <e>might</e>
+need it.
+</p>
+<p>
+The goal here is twofold. First, the principle of least privilege says
+that if a user doesn't need a real shell, he shouldn't have one. And
+along those same lines, not having a shell gives the system
+administrator some peace of mind: he doesn't have to be as concerned
+with whether or not that user has a password (and how strong it is),
+or whether or not its filesystem permissions are all set correctly, or
+whether or not it can log in via SSH, et cetera.
+</p>
+</body>
+</subsection>
+
+<subsection>
+<title>Choosing a home directory</title>
+<body>
+<p>
+In most cases, the default home directory (that is, no home directory)
+should be used. GLEP81 changed two aspects of user management with
+respect to home directories:
+</p>
+
+<ol>
+  <li>
+    Creating a user can now modify the permissions on an existing
+    directory. Should the need arise, this is necessary for a new
+    version of an <c>acct-user</c> package to be able to fix the
+    ownership and permissions of its home directory.
+  </li>
+  <li>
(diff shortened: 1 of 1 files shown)


ebuild-writing/users-and-groups/text.xml
```diff
@@ -103,6 +103,175 @@ The defaults are the user and primary group for the owner, and 0755 for the
...
+permissions. Our GLEP 81 implementation will reveal many user
+management issues that were allowed to fester in the past.
+</p>
+
+<subsection>
+<title>Choosing a shell</title>
+<body>
+<p>
+In most cases, the default shell (that is, no shell) should be
+used. Services can still be started as a user who has no shell, and
```
gktrk:
```suggestion
used. Services can still be started as a user that has no shell, and
```

I'm almost tempted to say that you were the person who quoted either "The Elements of Style" or Knuth about using `that` vs `who` :)

Alternatively, this would also work I think:

```suggestion
used. Services can still be started as a user with no shell, and
```

ebuild-writing/users-and-groups/text.xml
```diff
@@ -103,6 +103,175 @@ The defaults are the user and primary group for the owner, and 0755 for the
...
+daemons are able to drop privileges to a user who has no shell. If
+necessary, the administrator can override a user's default shell with
+<c>su -s &lt;shell&gt; &lt;username&gt;</c>. This is sufficient for
+testing, management of SSH credentials, and for initial configuration
+in an ebuild's <c>pkg_config</c> phase.
+</p>
+<p>
+An obvious exception to this rule is if a human being will need to log
```
gktrk: I wonder if we should compress this paragraph to `Don't blindly set your user's shell to <c>/bin/bash</c> unless there is a good reason to do so.` I can see this from a sysadmin's perspective but a package manager may not have the luxury to tell if the account will be logged in by a human being interactively.

ebuild-writing/users-and-groups/text.xml
```diff
@@ -103,6 +103,175 @@ The defaults are the user and primary group for the owner, and 0755 for the
...
+into the account interactively, as is the case with the <c>root</c>
+user. Other exceptions certainly exist, but should be evaluated
+case-by-case. In other words, if you haven't checked, don't set your
+user's shell to <c>/bin/bash</c> because you think he <e>might</e>
+need it.
+</p>
+<p>
+The goal here is twofold. First, the principle of least privilege says
```
gktrk: Similarly, I think this paragraph can be compressed into its second sentence. The details of how a sysadmin would worry about such things is not necessarily relevant to the package maintainer. The recommendation of following the principle of least privilege should be sufficient.

ebuild-writing/users-and-groups/text.xml
```diff
@@ -103,6 +103,175 @@ The defaults are the user and primary group for the owner, and 0755 for the
...
+<body>
+<p>
+In most cases, the default home directory (that is, no home directory)
+should be used. GLEP81 changed two aspects of user management with
+respect to home directories:
+</p>
+
+<ol>
+  <li>
+    Creating a user can now modify the permissions on an existing
```
gktrk: Are these points better off in a separate subsection for upgrading to GLEP 81? I assume after time time T, all packages will upgrade to GLEP 81 and the way this text reads will not make as much sense. So we could write as if there are no non-GLEP-81 ebuilds, move the upgrade stuff to a separate subsection and kill it later.

orlitzky:
Since this is critically important information that's now been missing from the devmanual for 4 months, and requires me to comment on every single incorrect PR posted to the -dev list, can you just shorten whatever you want to shorten during the merge process? The exceptions and especially rationale are important to keep people from just ignoring the guidelines thinking that they're stupid. I'll put the full version on my website.

orlitzky:
And in the process, please drop ACCT_USER_SHELL,  ACCT_USER_HOME, ACCT_USER_HOME_OWNER, and ACCT_USER_HOME_PERMS from the template ebuild.

orlitzky:
I'll leave the bug open, but I'm taking this off my to-do list after four months:

http://michael.orlitzky.com/articles/gentoo_glep81_user_package_guidelines.xhtml

gktrk:
I consistently fail to allocate time to shorten this, so I merged it as-is. I can't justify holding it up any longer.

ulm:
> I consistently fail to allocate time to shorten this, so I merged it as-is. I can't justify holding it up any longer.

First commit after I've fixed the XML structure and it breaks it again. :( This doesn't validate against the DTD, because is has "rogue" paragraphs outside of any subsection body.

Fixed in commit 7a723385b93d3b864f17a9f14b1de73b734794bd.

orlitzky:
> I consistently fail to allocate time to shorten this, so I merged it as-is. I can't justify holding it up any longer.

Thank you.



> > I consistently fail to allocate time to shorten this, so I merged it as-is. I can't justify holding it up any longer.
> 
> First commit after I've fixed the XML structure and it breaks it again. :( This doesn't validate against the DTD, because is has "rogue" paragraphs outside of any subsection body.

It was valid when I wrote it =P

PR #117, opened 2019-11-08
Pkg pretend improvement

gktrk:
The ebuild phase functions page still incorrectly starts the sequence from `pkg_setup` as opposed to `pkg_pretend`. Fix the figure and the text accompanying it.

### DIFF
--- a/ebuild-writing/functions/diagram.svg
+++ b/ebuild-writing/functions/diagram.svg
@@ -1,318 +1,416 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!-- Created with Inkscape (http://www.inkscape.org/) -->
 <svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
-   version="1.0"
-   width="1100"
-   height="80"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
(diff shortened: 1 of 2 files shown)


ebuild-writing/functions/diagram.svg
```diff
@@ -1,318 +1,417 @@
...
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
-   version="1.0"
-   width="1100"
-   height="80"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   inkscape:export-ydpi="96"
+   inkscape:export-xdpi="96"
+   inkscape:export-filename="/home/gokturk/stuff/devmanual/ebuild-writing/functions/diagram.png"
```
mgorny: Are you sure you want to leak this?

gktrk: I'd be more worried about polluting it than leaking. Thanks for noticing though. Force pushed the fix.

gktrk:
For reference, this is how the new figure looks like: https://dev.gentoo.org/~gokturk/devmanual/pr177/ebuild-writing/functions/diagram.png

orlitzky:
LGTM, but if you're soliciting extra improvements: I find the visual presentation of https://projects.gentoo.org/pms/7/pms.html#x1-1030009.2 to be almost perfect. The paragraph we have is hard to read compared to the nice bullet list. We're also missing the call order for **de**installations. Either way it would be nice to link to the PMS as the authoritative source.

gktrk:
> LGTM, but if you're soliciting extra improvements: I find the visual presentation of https://projects.gentoo.org/pms/7/pms.html#x1-1030009.2 to be almost perfect. The paragraph we have is hard to read compared to the nice bullet list. We're also missing the call order for **de**installations. Either way it would be nice to link to the PMS as the authoritative source.

I don't really have an idea about how to nicely incorporate that. Another option is using two separate colors of arrows: one of installation and one for removal.

Can we merge this as-is for now because of how broken it is?

PR #118, opened 2019-12-04
Fix XML structure

ulm:
This series of patches is a first attempt to fix the structure of the devmanual. I mainly went by what is actually supported by `devbook.xsl`, and I've changed or dropped those elements and attributes that aren't.

I've also written a preliminary [devbook.dtd](https://gist.github.com/ulm/871338b513483ce0ff45e12295aacfb1) and all files of the document validate with it:
`xmllint --noout --dtdvalid devbook.dtd text.xml`

As is to be expected for a project that has been developed since 10+ years without any formal verification, the resulting rules aren't very strict. For example, there can be formatting elements inside `<title>` when there really shouldn't be, and almost everything is allowed below `<li>` or `<ti>`. We might think about tightening the rules in a second round of updates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,12 @@
 # These "find" commands match text.xml and *.svg files, respectively,
 # but only after excluding the .git directory from the search for
 # performance and overall sanity reasons.
-HTMLS := $(subst text.xml,index.html,\
-  $(shell find ./ -name .git -prune -o \( -type f -name 'text.xml' -print \)))
-IMAGES := $(patsubst %.svg,%.png,\
-  $(shell find ./ -name .git -prune -o \( -type f -name '*.svg' -print \)))
+XMLS := $(shell find . -name .git -prune -o -type f -name 'text.xml' -print)
+SVGS := $(shell find . -name .git -prune -o -type f -name '*.svg' -print)
+HTMLS := $(subst text.xml,index.html,$(XMLS))
+IMAGES := $(patsubst %.svg,%.png,$(SVGS))
 
-all: prereq $(HTMLS) $(IMAGES)
+all: prereq validate $(HTMLS) $(IMAGES)
 
 prereq:
 	@type convert >/dev/null 2>&1 || \
@@ -14,6 +14,9 @@ prereq:
           exit 1; }
 	@type xsltproc >/dev/null 2>&1 || \
 	{ echo "dev-libs/libxslt is required" >&2;\
+	  exit 1; }
+	@type xmllint >/dev/null 2>&1 || \
+	{ echo "dev-libs/libxml2 is required" >&2;\
 	  exit 1; }
 
 %.png : %.svg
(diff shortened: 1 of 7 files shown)


orlitzky:
Hijacking: while you're doing this, maybe you want to apply these fixes that work towards the same goal?

https://626032.bugs.gentoo.org/attachment.cgi?id=574554
https://626032.bugs.gentoo.org/attachment.cgi?id=574558
https://626032.bugs.gentoo.org/attachment.cgi?id=574560

These fix structural problems that prevented me from running html-tidy on the result in bug 626032.

ulm:
> Hijacking: while you're doing this, maybe you want to apply these fixes that work towards the same goal?
> 
> https://626032.bugs.gentoo.org/attachment.cgi?id=574554

Patch looks good, but can we please keep it separate from this PR?

> https://626032.bugs.gentoo.org/attachment.cgi?id=574558

I've fixed that one differently in commit 939c9bff1acbc73b16e869f24d7cb59e28e27e24 by moving the subsection into the preceding section. Arguably, "Simultaneous stabilization on all architectures" is a subtopic of "Moving from ~arch to arch".

> https://626032.bugs.gentoo.org/attachment.cgi?id=574560

I'm not happy about this one, because there is no reason why `<uri>` should behave differently from GuideXML. So really the underlying breakage should be fixed in `devbook.xsl`.

ulm:
Merged to master, except for the DTD and the Makefile update.

Makefile
```diff
@@ -1,12 +1,12 @@
...
 # but only after excluding the .git directory from the search for
 # performance and overall sanity reasons.
-HTMLS := $(subst text.xml,index.html,\
-  $(shell find ./ -name .git -prune -o \( -type f -name 'text.xml' -print \)))
-IMAGES := $(patsubst %.svg,%.png,\
-  $(shell find ./ -name .git -prune -o \( -type f -name '*.svg' -print \)))
+XMLS := $(shell find . -name .git -prune -o -type f -name 'text.xml' -print)
+SVGS := $(shell find . -name .git -prune -o -type f -name '*.svg' -print)
+HTMLS := $(subst text.xml,index.html,$(XMLS))
+PNGS := $(patsubst %.svg,%.png,$(SVGS))
```
gktrk: I'm skeptical about the change from `IMAGES` to `PNGS`. Can we do:

```suggestion
PNGS := $(patsubst %.svg,%.png,$(SVGS))
IMAGES := $(PNGS)
```

ulm: I have reverted the variable name to IMAGES.

orlitzky: Bikeshed only: I thought `PNGS` was fine, because the "png" extension is mentioned explicitly and it's consistent with `XMLS` and `HTMLS`. The place where you would want `IMAGES` is in the default target, because you want to build all images there. Something like

```
IMAGES = $(PNGS) $(JPGS)
all: ... $(IMAGES)
```

But obviously not a big deal.

PR #119, opened 2019-12-06
devbook.xsl: Fix handling of <uri> without link attribute.

ulm:

### DIFF
--- a/devbook.xsl
+++ b/devbook.xsl
@@ -336,8 +336,15 @@
           </xsl:otherwise>
         </xsl:choose>
       </xsl:when>
-      <xsl:otherwise>
+      <xsl:when test="@link">
         <a class="{$class}" href="{@link}"><xsl:value-of select="."/></a>
+      </xsl:when>
+      <xsl:when test="contains(., '://')">
+        <a class="{$class}" href="{.}"><xsl:value-of select="."/></a>
+      </xsl:when>
+      <xsl:otherwise>
+        <xsl:message>Error: No link target (<xsl:value-of select="."/>)</xsl:message>
+        <a class="{$class}"><xsl:value-of select="."/></a>
       </xsl:otherwise>
     </xsl:choose>
   </xsl:template>


ulm:
This should fix the `<uri>https://forums.gentoo.org/</uri>` example in the appendix.

orlitzky:
I just did the same thing on [bug 702180](https://bugs.gentoo.org/702180) =P

Either version is fine. The main difference is that mine bails out early if the link is missing. This avoids doing a bunch of tests on an attribute that isn't there, and let's me leave the existing code alone. But the trade-off is that I increment the code complexity (i.e. indentation) level by one.

Mine also accepts any text, whereas yours will reject anything without `://` (do you want `terminate="yes"` on that xsl:message?) The added check was a good idea.

ulm:
The idea was that forgetting the link attribute in `<uri>Gentoo Forums</uri>` is not completely unlikely as a mistake, so I check for `://` which shouldn't normally occur in plain text.

ulm:
> xsltproc devbook.xsl appendices/contributing/devbook-guide/text.xml > appendices/contributing/devbook-guide/index.html
> Error: No link target (forums.gentoo.org)

I cannot reproduce this. `appendices/contributing/devbook-guide` is processed without error here.

>     2. Some URIs without a `://` are valid, like relative path references. You could always override this by specifying `link=`, but it's a little annoying to reject something that's technically valid.

Yes, but I cannot imagine an example where anything else than an external link would be specified in a link text (and this would require more intrusive changes). So, let's discuss it again when we'll have the first instance of it.

orlitzky:
> > xsltproc devbook.xsl appendices/contributing/devbook-guide/text.xml > appendices/contributing/devbook-guide/index.html
> > Error: No link target (forums.gentoo.org)
> 
> I cannot reproduce this. `appendices/contributing/devbook-guide` is processed without error here.
> 

That file's fine, I edited it to delete the `https://` so that I could test the error message. But, when there's a mistake, `make` should die there instead of proceeding as if everything's fine.

ulm:
> But, when there's a mistake, `make` should die there instead of proceeding as if everything's fine.

I haven't added the `terminate` attribute, because that would make remote generation of the document fail (which only infra can see, the devmanual team isn't noticed of it). Also locally, the error message can be seen, and it's the only output from `xsltproc` which otherwise works silently. So it's not like we're being drowned in output.

PR #120, opened 2019-12-11
Improve search

gktrk:
This is an attempt to merge #93 and provide improvements on top. Will merge once the infra deploys minified lunr.js. The function `stringify_node()` in `search_index.py` needs some extra attention regarding whitespace handling. Will do that and commentify the code before the merge.


gktrk:
The biggest downside is that `documents.js` is 414K now. We might wanna look into conditionally fetching it somewhow (i.e. fetching it on first search attempt), and possibly compressing it.

gktrk:
> The biggest downside is that `documents.js` is 414K now. We might wanna look into conditionally fetching it somewhow (i.e. fetching it on first search attempt), and possibly compressing it.

I pushed code that fetches `documents.js` on the first click to the search text area (not the button). This should eliminate the search overhead for people who never utilize it.

I also did some further performance analysis.

If we ship `documents.js` and build the index in the  browser, here's the overhead breakdown:
```
{
  GET documents.js: 501.92 ms
  buildIndex(): 309.40 ms
} 811.32 ms
```

If we create a prebuilt index on the infra side (`documents.json`) and ship that along with `documents.js` (we still need to index into this to match refs to results):
```
{
  {
    GET documents.json: 582.72 ms
    GET documents.js: 622.11 ms
  }
  e.Index.load(): 28.80 ms
} 650.91 ms
```

Note that the two GET requests take place in parallel.

The first time the website is run, the first method is about 25% slower. In a more realistic scenario, both `documents.json` and `documents.js` will be hits in the browser cache. In that case, we only compare the overhead of `buildIndex()`to `e.index.load()`, which is 10x slower. I think there is a good argument for shipping the prebuilt index. Anybody interested in reproducing these results can do so via https://dev.gentoo.org/~gokturk/devmanual/improve-search/index.html.

Makefile
```diff
@@ -13,12 +13,18 @@ prereq:
...
 	  exit 1; }
 	@type xmllint >/dev/null 2>&1 || \
 	{ echo "dev-libs/libxml2 is required" >&2;\
 	  exit 1; }
 
+# Since search_index.py rebuilds the index from scratch instead of
+# updating it, we pass it the names of ALL prerequisites ($^) and not
+# just the names of the ones that are new ($?).
+documents.js: $(XMLS)
+	./search_index.py $^ > documents.js
```
robbat2: you must write to a temporary file and rename into the final name atomically.

robbat2: add this to the `all` target as well.

search_index.py
```diff
@@ -0,0 +1,61 @@
...
+import sys
+import xml.etree.ElementTree as ET
+
+files = sys.argv[1:]
+documents = []
+url_root = 'https://devmanual.gentoo.org/'
+
+def stringify_node(parent):
+    if parent.text:
+        text = parent.text.strip().replace('\n', ' ')
```
robbat2: shouldn't this be the other way around maybe?

gktrk: Let me re-examine the behavior of ElementTree on empty tags (such as `<d/>`) and get back to you.

gktrk:
@ulm I'm ready to merge this. The javascript aside, do you have any general remarks?

ulm:
Looks like the semaphoreci check fails now? Why?

document.js should be added to the `make clean` target.

gktrk:
> Looks like the semaphoreci check fails now? Why?
> 

The CI is invoking python2 by default which doesn't support the function annotation syntax. I've forced it to use python3. It should be ok now.

> document.js should be added to the `make clean` target.

It's already in there, no?

```
--- a/Makefile
+++ b/Makefile
@@ -48,6 +49,6 @@ validate: prereq
          && echo "xmllint validation successful"
 
 clean:
-       rm -f $(HTMLS) $(IMAGES) documents.js
+       rm -f $(HTMLS) $(IMAGES) _documents.js documents.js
 
 .PHONY: all prereq validate clean

```

.gitignore
```diff
@@ -1,3 +1,4 @@
 *.html
 *.png
+documents.js
```
robbat2: ignore your tempfile as well.

gktrk: Done

PR #128, opened 2019-12-25
remove needless pam section

Zlogene:
Signed-off-by: Mikle Kolyada <zlogene@gentoo.org>

### DIFF
--- a/tasks-reference/pam/text.xml
+++ b/tasks-reference/pam/text.xml
@@ -16,14 +16,6 @@ With PAM, a program just needs to require authentication for a given login class
 (defined in a <c>pam.d</c> file), and PAM framework will take care of calling the
 modules which will provide authentication.
 </p>
-
-<p>
-There are different PAM implementations. Gentoo Linux, by default, uses the
-Linux-PAM implementation which is installed via <c>sys-libs/pam</c>; FreeBSD and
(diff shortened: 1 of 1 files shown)


tasks-reference/pam/text.xml
```diff
@@ -99,16 +91,6 @@ The statement is composed of 3 or 4 tokens:
...
-
-<p>
-As the number and the type of modules shipped with the implementation depends on
-the implementations themselves (Linux-PAM provides a full working set of
-modules, OpenPAM doesn't provide modules at all, and it's the operating system
-which provides them, as FreeBSD or NetBSD do), there are just a few modules
-which can be used directly in <c>pamd</c> files without the risk of providing a
-non-working configuration file:
-</p>
-
```
gktrk: You are removing the paragraph leading to a list. Should this now be merged with the `<ul/>` above instead?

tasks-reference/pam/text.xml
```diff
@@ -163,25 +145,15 @@ completely non-portable. It is not used in all the implementations of Linux-PAM
```
gktrk: The paragraph above talks about `other PAM implementations`. This should also be updated, no?

gktrk:
@gentoo-bot bad bot!

PR #129, opened 2019-12-26
Unsupported elements

ulm:


ulm:
> Assuming there's nothing in `devbook.xsl` that made use of these tags, LGTM.

None of them are processed by devbook.xsl, i.e. these tags are simply ignored.

gktrk:
We never rendered `<path/>` differently. I thought about different ways to display it but I can't think of any. I don't see any future use of knowing what is a path and what isn't. Let's get rid of them.

PR #130, opened 2019-12-26
More conventional syntax of definition lists

ulm:
Definition lists required only `p` elements as children of `dd`. Change to a more conventional syntax where each `dt` element can be followed by one **or more** `dd` elements.


ulm:
> Does it affect the way multiple `<p/>` are displayed in a `<dd/>` at all?

Slightly. The margin above the first paragraph and after the last paragraph remains the same, while the vertical space between adjacent paragraphs (i.e., adjacent `<dd/>`) will slightly decrease. Which looks better, IMHO.

PR #131, opened 2019-12-29
general-concepts/dependencies: Update for EAPI 7, describe circular deps

mgorny:
Multiple improvements to the dependencies chapter:
1. Document `BDEPEND` (with minimal CBUILD/CHOST explanation).
2. Make `PDEPEND` less weird.
3. Remove section on `:` operator, it's not useful for devs.
4. Replace EAPI-conditional implicit RDEPEND explanation with just a minimal warning (we no longer support those old EAPIs).
5. Provide a more detailed section on ciicular deps and how to resolve them.

### DIFF
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -622,5 +622,54 @@ DEPEND="test? ( dev-util/foo )"
 
 </body>
 </section>
+
+<section>
+<title>Circular Dependencies</title>
+<body>
+
+<p>
+Circular dependencies occur if one or more of package's (possibly indirect)
+dependencies depend on the package itself. This creates a dependency cycle where
+each of the packages must technically be installed before the other.
+For example, if package A depends on B, B depends on C and C depends on A, then
+the package manager cannot install A before C, and C before A.
+</p>
+
+<p>
+There are three kinds of circular dependencies:
+</p>
+
+<ol>
+  <li>
+    Circular dependencies that occur if only one of the packages strictly needs
+    to be installed before the other. For example, <c>dev-python/certifi</c>
+    strictly requires <c>dev-python/setuptools</c> to build but the latter
+    package requires the former for some runtime functionality. As a result,
+    <c>dev-python/certifi</c> can be installed later than the other package.
+    <c>PDEPEND</c> is used to express this and automatically resolve
+    the circular dependency.
+  </li>
+
+  <li>
+    Circular dependencies that occur if the cycle applies only to some
+    combination of USE flags on one of the packages. For example, running tests
+    in <c>dev-python/setuptools</c> requires a number of packages which require
+    <c>dev-python/setuptools</c> to be installed first. This kind of circular
+    dependency can be resolved by the user by adjusting USE flags on one
+    of the packages, e.g. by disabling tests on <c>dev-python/setuptools</c>,
+    and reenabling them once the dependency is initially installed.
+  </li>
+
+  <li>
+    Circular dependencies that cannot be resolved using the regular means.
+    For example, <c>dev-util/cmake</c> used to depend
+    on <c>dev-libs/jsoncpp</c>, while the latter package used the former
+    to build. Resolving this kind of dependency usually requires bundling one
+    of the dependencies conditionally, or providing an alternate bootstrap path.
+  </li>
+</ol>
+
+</body>
+</section>
 </chapter>
 </guide>
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -12,14 +12,58 @@ provided by <c>emerge</c>.
 </body>
 
 <section>
+<title>CHOST vs CBUILD</title>
+<body>
+
+<p>
+In order to avoid ambiguity, we use the following terms to indicate different
+systems when cross-compiling:
+</p>
+
+<dl>
+  <dt>CBUILD</dt>
+  <dd>
+    The system on which the build is performed.  Dependencies that apply
+    to the CBUILD system can be executed during build time.  When
+    cross-compiling, they are not installed into the system being built.
+  </dd>
+
+  <dt>CHOST</dt>
+  <dd>
+    The system on which the package is going to be executed.  When
+    cross-compiling, dependencies applying to CHOST can not be executed.
+  </dd>
+</dl>
+
+<p>
+When not cross-compiling, CBUILD and CHOST have the same value and both classes
+of dependencies are merged.
+</p>
+
+</body>
+</section>
+
+<section>
 <title>Build Dependencies</title>
 <body>
 
 <p>
-The <c>DEPEND</c> ebuild variable should specify any dependencies which are
-required to unpack, patch, compile or install the package (but see
+Build dependencies are used to specify any dependencies that are required
+to unpack, patch, compile, test or install the package (but see
 <uri link="::general-concepts/dependencies#Implicit System Dependency"/> for
 exemptions).
+</p>
+
+<p>
+Starting with EAPI 7, build dependencies are split into two variables:
+<c>BDEPEND</c> and <c>DEPEND</c>. <c>BDEPEND</c> specifies dependencies
+applicable to CBUILD, i.e. programs that need to be executed during the build,
+e.g. <c>virtual/pkgconfig</c>. <c>DEPEND</c> specifies dependencies for CHOST,
+i.e. packages that need to be found on built system, e.g. libraries and headers.
+</p>
+
+<p>
+In earlier EAPIs, all build dependencies are placed in <c>DEPEND</c>.
 </p>
 
 </body>
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -94,16 +94,14 @@ Items which are in <c>RDEPEND</c> but not <c>DEPEND</c> could <e>in theory</e> b
 </section>
 
 <section>
-<title>Post-Merge Dependencies</title>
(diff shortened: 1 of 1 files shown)


general-concepts/dependencies/text.xml
```diff
@@ -32,10 +76,9 @@ exemptions).
 <p>
 The <c>RDEPEND</c> ebuild variable should specify any dependencies which are
 required at runtime. This includes libraries (when dynamically linked), any data
-packages and (for interpreted languages) the relevant interpreter. In EAPI=3 or
-older, if this variable is not specified it defaults to the value of
-<c>DEPEND</c>, however the implicit usage is frowned upon. In EAPI=4, the
-implicit behaviour was removed and the assignment is always explicit.
+packages and (for interpreted languages) the relevant interpreter. While
+historically this variable could have implicit value, nowadays it must be
```
sbraz: "an implicit value"?
Does it mean that we need to always specify `DEPEND=""` now?

floppym: @sbraz `RDEPEND` is empty by default in modern EAPIs. In older EAPIs, it was implicitly set to the value of `DEPEND` if not otherwise set in the ebuild.

general-concepts/dependencies/text.xml
```diff
@@ -608,6 +619,54 @@ RESTRICT="!test? ( test )"
...
+the package manager cannot install A before C, and C before A.
+</p>
+
+<p>
+There are three kinds of circular dependencies:
+</p>
+
+<ol>
+  <li>
+    Weak circular dependencies that occur if only of the packages strictly need
```
floppym: I think the word "one" is missing in this line.

general-concepts/dependencies/text.xml
mgorny: Made it clearer that this applies only if it's not empty.

general-concepts/dependencies/text.xml
mgorny: Fixed, thanks.

general-concepts/dependencies/text.xml
```diff
@@ -608,6 +619,54 @@ RESTRICT="!test? ( test )"
...
+</p>
+
+<p>
+There are three kinds of circular dependencies:
+</p>
+
+<ol>
+  <li>
+    Weak circular dependencies that occur if only one of the packages strictly
+    needs to be installed before the other. For example, the Python certifi
```
ulm: Maybe use `<c/>` for typing the package name?

general-concepts/dependencies/text.xml
```diff
@@ -608,6 +619,54 @@ RESTRICT="!test? ( test )"
...
+    which require setuptools to be installed first. This kind of circular
+    dependency can be resolved by the user by adjusting USE flags on one
+    of the packages, e.g. by disabling tests on setuptools, and reenabling them
+    once the package is initially installed.
+  </li>
+
+  <li>
+    Strong circular dependencies that occur if cycle cannot be resolved using
+    the regular means. For example, CMake used to depend on jsoncpp, while
+    jsoncpp required CMake to build. Resolving this kind of dependency usually
```
ulm: Again, I'd write `<c>cmake</c>` etc. here. Also, this is about the Gentoo package, whose name is all lowercase (so no `CMake`).

general-concepts/dependencies/text.xml
```diff
@@ -543,36 +584,6 @@ presence or absence.
...
-</codesample>
-
-<p>
-It is useful to recognise the legacy syntax and to know that it is no longer
-valid.
-</p>
-
-</body>
-</section>
-
```
ulm: Another candidate to move to above mentioned appendix.

general-concepts/dependencies/text.xml
```diff
@@ -32,10 +76,9 @@ exemptions).
 <p>
 The <c>RDEPEND</c> ebuild variable should specify any dependencies which are
 required at runtime. This includes libraries (when dynamically linked), any data
-packages and (for interpreted languages) the relevant interpreter. In EAPI=3 or
-older, if this variable is not specified it defaults to the value of
-<c>DEPEND</c>, however the implicit usage is frowned upon. In EAPI=4, the
-implicit behaviour was removed and the assignment is always explicit.
+packages and (for interpreted languages) the relevant interpreter. While
+in historical EAPIs this variable defaulted to the same value as <c>DEPEND</c>,
+nowadays it must always be specified explicitly if not empty.
```
ulm: I'd rather remove the last sentence altogether.
Or maybe move to an appendix like "historical curiosities".

general-concepts/dependencies/text.xml
mgorny: I was actually focusing on more generic approach (i.e. you have the same problem if you build straight from sources).

general-concepts/dependencies/text.xml
mgorny: Not sure if it's worth keeping in devmanual. We already have it in PMS.

mgorny:
Removed implicit RDEPEND and switched to using Gentoo package names.

general-concepts/dependencies/text.xml
gktrk: Agreed with @mgorny. It's already in the PMS.

mgorny:
It's not established, I've just invented it. If you're aware of any established fitting terminology, I'd be happy to use it instead.

gktrk:
> It's not established, I've just invented it. If you're aware of any established fitting terminology, I'd be happy to use it instead.

My issue is with having "weak", "non-weak", and "strong". If I understand it correctly, "weak" is a runtime-only circular dependency. That's why we can move them to PDEPEND, as it doesn't prevent the package from being merged. Then you have the USE-conditional strong dependency. They can appear in various *DEPEND, but only conditionally, which allows one to break it by merging with the USE flag disabled first, and merging it again. Lastly, you have strict/strong circular dependencies. They truly require each other during both build time and runtime. So what do you think about this:

 * Runtime-only circular dependency
 * USE-conditional circular dependency
 * Strict (or strong) circular dependency

mgorny:
Actually, the first one can also involve build-time (+run-time) <-> run-time cycle. The main point being that one of them is PDEPEND-able. Then, sometimes even RDEPEND<->RDEPEND can't be resolved this way (even if Portage arbitrarily tries it and breaks postinst).

general-concepts/dependencies/text.xml
ulm: I have actually dropped this section with EAPI cleanup in #145.

general-concepts/dependencies/text.xml
```diff
@@ -11,17 +11,61 @@ provided by <c>emerge</c>.
...
+
+<p>
+In order to avoid ambiguity, we use the following terms to indicate different
+systems when cross-compiling:
+</p>
+
+<dl>
+  <dt>CBUILD</dt>
+  <dd>
+    The system on which the build is performed.  Dependencies which apply
```
ulm:
```suggestion
    The system on which the build is performed.  Dependencies that apply
```

https://www.quickanddirtytips.com/education/grammar/which-versus-that

general-concepts/dependencies/text.xml
```diff
@@ -11,17 +11,61 @@ provided by <c>emerge</c>.
...
+    to CBUILD system can be executed during build time.  When cross-compiling,
```
ulm:
```suggestion
    to the CBUILD system can be executed during build time.  When cross-compiling,
```

general-concepts/dependencies/text.xml
```diff
@@ -11,17 +11,61 @@ provided by <c>emerge</c>.
...
+</section>
+
 <section>
 <title>Build Dependencies</title>
 <body>
 
 <p>
-The <c>DEPEND</c> ebuild variable should specify any dependencies which are
-required to unpack, patch, compile or install the package (but see
+Build dependencies are used to specify any dependencies which are
```
ulm:
```suggestion
Build dependencies are used to specify any dependencies that are
```

general-concepts/dependencies/text.xml
```diff
@@ -606,6 +615,57 @@ RESTRICT="!test? ( test )"
...
+  </li>
+
+  <li>
+    Strong circular dependencies that occur if cycle cannot be resolved using
+    the regular means. For example, <c>dev-util/cmake</c> used to depend
+    on <c>dev-libs/jsoncpp</c>, while <c>dev-libs/jsoncpp</c> required
+    <c>dev-util/cmake</c> to build. Resolving this kind of dependency usually
+    requires bundling one of the dependencies conditionally, or providing
+    an alternate bootstrap path.
+  </li>
```
ulm: I agree with gokturk that I also dislike this ternary terminology "weak" vs "non-weak" vs "strong". One would expect that "non-weak" is the opposite of "weak" so there cannot be a third category.

Do we even need names for them? Unless they are referenced from another part of the document, this doesn't seem strictly necessary.

gktrk: The terminology doesn't exist. I don't wanna hold up the PR just for that but I also don't wanna commit to a particular terminology. If we go with "The first kind ...", "The second kind ...", "The third kind ...", I'd be fine with that. Terminology bikeshedding shouldn't hold this up.

mgorny: I think the sentences still make sense even without extra terms like 'first kind'. I'll rewrite them a bit and push the update.

mgorny:
I think I've applied all suggestions, and reworded the text on circular deps. I think it still can be improved, so I'm open to suggestions how to word it better.

PR #132, opened 2019-12-30
Package and slot moves: move and improve

mgorny:
Package moves are rather rarely used, and quite powerful. Let's move them into a dedicated chapter rather than hiding in middle of maintenance tasks. While at it, also:
* add warnings explaining implications of moves better
* cover package renames as well as cross-category moves
* improve the readability of package move section, and remove (inconsistent) duplication in example


ebuild-maintenance/package-and-slot-moves/text.xml
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0"?>
+<guide self="ebuild-maintenance/maintenance-tasks/">
+<chapter>
+<title>Package and slot moves</title>
+
+<body>
+<p>
+This chapter describes the use of package and slot moves. The package update
+mechanism is a powerful too, and needs to be used with caution. In particular,
```
gktrk:
```suggestion
mechanism is a powerful tool, and needs to be used with caution. In particular,
```

ebuild-maintenance/package-and-slot-moves/text.xml
```diff
@@ -0,0 +1,116 @@
...
+the following must be noted:
```
gktrk:
```suggestion
the followings must be noted:
```

ebuild-maintenance/package-and-slot-moves/text.xml
```diff
@@ -0,0 +1,116 @@
...
+
+<ol>
+  <li>
+    Updates are <e>not one-shot</e> operations and they are not stateful. All
+    updates can be reapplied multiple times to the same system, and all old
+    updates are applied to fresh Gentoo installations.
+  </li>
+
+  <li>
+    Once an update entry is created, the old package name (slot) cannot be
```
gktrk:
```suggestion
    Once an update entry is created, the old package name (or slot) cannot be
```

ebuild-maintenance/package-and-slot-moves/text.xml
```diff
@@ -0,0 +1,116 @@
...
+    reused. Attempting to reuse it will cause updates to apply again,
+    to the reused name. This also means that updates cannot be undone.
+  </li>
+
+  <li>
+    Updates can only perform 1:1 moves. They cannot be used to merge packages.
```
gktrk:
```suggestion
    Updates can only perform one-to-one moves. They cannot be used to merge packages.
```

ebuild-maintenance/package-and-slot-moves/text.xml
```diff
@@ -0,0 +1,116 @@
...
+<section>
+<title>Moving or renaming a package</title>
+<body>
+
+<p>
+Moving or renaming a package requires several operations. Firstly, verify that
+the ebuilds will continue to work correctly after the move. If the category
+changes, you must verify all <c>${CATEGORY}</c> uses. If the package name
+changes, you must verify <c>${PN}</c>, <c>${P}</c>, etc. Whenever the old value
+is necessary, inline it. Commit the changes separately before the move.
```
gktrk: What do  you mean by inlining here? Substituting the variable that represents the old package cat/name with the verbatim text, so that it doesn't get affected by the move? Also what's the rationale for creating a separate commit for those before the move?

ulm:
> We should either move `Git for Gentoo Developers` somewhere else and merge `Maintenance Tasks` into its parent, or split every section in it into its own chapter.

I'd prefer the first. This would also fix the internal links to the chapter, which are broken since `Maintenance Tasks` was split out in commit d2e77eb8f1abe200a4c0ede569fa169c08205545. (It's on my to-do list to fix them, but merging the section into its parent would save me that work.)

mgorny:
> We should either move `Git for Gentoo Developers` somewhere else and merge `Maintenance Tasks` into its parent, or split every section in it into its own chapter.

I was actually going for the latter. A big 'maintenance tasks' section is basically 'totally random stuff, if you're looking for something it might or might not be there'.

ebuild-maintenance/package-and-slot-moves/text.xml
mgorny: Wait, 'following' has a plural? oO

ebuild-maintenance/package-and-slot-moves/text.xml
mgorny: Yes. The rationale is to clearly split the move into smaller changes, and make the diff more readable. Doing a move in one commit is like using a merge commit to alter some files.

ulm:
> I was actually going for the latter. A big 'maintenance tasks' section is basically 'totally random stuff, if you're looking for something it might or might not be there'.

WFM, though it looks like we keep moving things around. (We had `ebuild-writing/ebuild-maintenance/` first, then `ebuild-maintenance/`, then `ebuild-maintenance/maintenance-tasks/`, and now it is to be moved again.)

ebuild-maintenance/package-and-slot-moves/text.xml
```diff
@@ -0,0 +1,116 @@
+<?xml version="1.0"?>
+<guide self="ebuild-maintenance/maintenance-tasks/">
```
ulm: This must be updated.

ebuild-maintenance/package-and-slot-moves/text.xml
ulm: > Wait, 'following' has a plural? oO

My dictionary says that it has a plural only for the meaning "group of followers", which doesn't apply here.

gktrk:
It's a mistake on my part. It shouldn't be plural.

>
> *@ulm* commented on this pull request.

> 
> In ebuild-maintenance/package-and-slot-moves/text.xml <https://github.com/gentoo/devmanual/pull/132#discussion_r369878404>:

> > @@ -0,0 +1,116 @@
+<?xml version="1.0"?>
+<guide self="ebuild-maintenance/maintenance-tasks/">
+<chapter>
+<title>Package and slot moves</title>
+
+<body>
+<p>
+This chapter describes the use of package and slot moves. The package update
+mechanism is a powerful too, and needs to be used with caution. In particular,
+the following must be noted:

ulm:
Merged, with all suggestions applied.

PR #133, opened 2020-01-03
ebuild-maintenance/git: merge different references to #BUG-ID

wimmuskee:
Assuming the #BUG-ID references are not preferred by default, move their explanations to the end of the article.

### DIFF
--- a/ebuild-maintenance/git/text.xml
+++ b/ebuild-maintenance/git/text.xml
@@ -233,16 +233,16 @@ For packages where <c>${CATEGORY}/${PN}:</c> is long, the line length
 limit can be exceeded, if absolutely necessary, to ensure a more
 useful summary line. If a commit affects multiple directories, prepend
 the message with which reflects the intention of the change best. If
-there are any bugs on Gentoo Bugzilla associated with the commit, id
+there are any bugs on Gentoo Bugzilla associated with the commit, the id
 of the bug can be appended to the summary line using the format
-<c>#BUG-ID</c>. If you are modifying keywords, clearly state what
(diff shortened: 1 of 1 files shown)


ebuild-maintenance/git/text.xml
```diff
@@ -286,11 +275,19 @@ as --force) for the commit.</li>
 </ul>
 
 <p>
-Additionally, some developers prefer referencing bugs on the summary
-line using the <c>#nnnnnn</c> form. Developers are free to use either
-form, or both simultaneously to combine their advantages.
+Additionally, some developers prefer referencing Gentoo Bugzilla bugs on
+the summary line using the <c>#BUG-ID</c> form. Developers are free to use
```
ulm: IMHO the `#nnnnnn` made it immediately clear that a numerical ID is meant, whereas one may wonder what a "BUG-ID" is. Please leave `#nnnnnn` in place.

ebuild-maintenance/git/text.xml
```diff
@@ -286,11 +275,19 @@ as --force) for the commit.</li>
 </ul>
 
 <p>
-Additionally, some developers prefer referencing bugs on the summary
```
gktrk: We wrote this paragraph when there was a controversy about the summary line vs. the Bug tag. This can go away now.

ebuild-maintenance/git/text.xml
wimmuskee: I've updated the _#BUG-ID_ references to _#nnnnnn_.

ebuild-maintenance/git/text.xml
wimmuskee: done

PR #134, opened 2020-01-03
ID attributes should contain only valid characters.

ulm:
Keep only characters that HTML 4 allows in ID attributes (but exclude the colon as well). Note that HTML 5 is more sloppy and would permit characters outside of the ASCII range, but these won't be valid in URI fragment identifiers (RFC 3986).

### DIFF
--- a/devbook.xsl
+++ b/devbook.xsl
@@ -278,8 +278,8 @@
 
   <xsl:template name="convert-to-anchor">
     <xsl:param name="data"/>
-    <xsl:variable name="lcletters">abcdefghijklmnopqrstuvwxyz--</xsl:variable>
-    <xsl:variable name="ucletters">ABCDEFGHIJKLMNOPQRSTUVWXYZ<xsl:text> </xsl:text>,</xsl:variable>
+    <xsl:variable name="lcletters">abcdefghijklmnopqrstuvwxyz-</xsl:variable>
+    <xsl:variable name="ucletters">ABCDEFGHIJKLMNOPQRSTUVWXYZ<xsl:text> </xsl:text></xsl:variable>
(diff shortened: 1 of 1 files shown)


ulm:
> Can we get away with a literal em dash instead of the escape sequence? I think copy/pasting a `—` in there falls between `<d/>` and `&#x2014;` on the usability spectrum.

I have escaped it on purpose, because different unicode dashes like `-` (hyphen-minus), `—` (em dash), `–` (en dash), `‒` (figure dash), `−` (minus sign), `―` (horizontal bar), or `⸺` (two-em dash) are practically indistinguishable when editing the source code using a teletype font. With `&#x2014;` it is immediately clear what is meant.

gktrk:
`ID and NAME tokens must begin with a letter ([A-Za-z]) and may be followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons (":"), and periods (".").`

Why can't we leave `--` alone or convert it to `-`? It seems a little extreme to insert an escape sequence there.

ulm:
> `ID and NAME tokens must begin with a letter ([A-Za-z])

Right, that's still not guaranteed. Can we postpone this until we have the first section title that doesn't begin with a letter? (Also, I don't know _how_ to solve it. Maybe prefix the anchor with `x` if its first character isn't a letter?)

> and may be followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons (":"), and periods (".").`
> 
> Why can't we leave `--` alone or convert it to `-`? It seems a little extreme to insert an escape sequence there.

I've dropped that part for now.

orlitzky:
Prepending "x" is revolting =)

If the plan to address duplicate IDs is "don't do that," and if we can simply fix them after `make tidy` catches them, then we should adopt the same policy for IDs that begin with an invalid character.

PR #135, opened 2020-01-04
Make (sub-)section titles unique within each chapter.

ulm:
Section, subsection, etc. titles are used to construct ID attributes,
which must be unique within each chapter.

One approach to disambiguate these identifiers would be to use the
complete section hierarchy for constructing them (see bug 626032).
However, that would break external references (and for example, links
from bugzilla or from mailing list archives cannot be updated).

Use a less intrusive approach instead and make the titles of the few
ambiguous subsections unique, or convert them to a list.

Closes: https://bugs.gentoo.org/626032

### DIFF
--- a/ebuild-writing/eapi/text.xml
+++ b/ebuild-writing/eapi/text.xml
@@ -137,7 +137,7 @@ IUSE="foo +bar"
 <title>EAPI=2</title>
 
 <subsection>
-<title>Helpers</title>
+<title>EAPI 2 Helpers</title>
 <body>
 <ul>
 	<li>
@@ -158,7 +158,7 @@ doman foo.lang.1
 </subsection>
 
 <subsection>
-<title>Metadata</title>
+<title>EAPI 2 Metadata</title>
 <body>
 <ul>
 	<li>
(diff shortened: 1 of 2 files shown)


orlitzky:
My objection to this approach was that it imposes invisible constraints on the _content_ of the devmanual in order to satisfy a mysterious technical metadata constraint in another file.

But nevertheless, it is a simpler fix, and you can check that it works with "make tidy" now.

I can live with it.

ulm:
> My objection to this approach was that it imposes invisible constraints on the _content_ of the devmanual in order to satisfy a mysterious technical metadata constraint in another file.
> 
> But nevertheless, it is a simpler fix, and you can check that it works with "make tidy" now.
> 
> I can live with it.

Another interesting approach is used by Mediawiki, which uses the (sub-)section title for the anchor, but disambiguates it by appending numbers to all subsequent anchors if there are duplicates. For example, if there are two sections "foo bar", the first anchor will be `#foo_bar` and the second one will be `#foo_bar_2`.

Unfortunately, my XSLT skills aren't sufficient to implement something similar (at least not in the amount of time I'm willing to spend on it). Plus, that scheme would have the problem that inserting a section with a duplicate name at the beginning would make the anchors for the following ones change, i.e., external links could suddenly point at the wrong target.

PR #136, opened 2020-01-05
Repoman-related cleanup + Manifests

mgorny:
Started as cleanup of repoman mentions but also found really stale Manifest mentions.

Specifically:
* <del>Remove duplicate mentions of `repoman commit`. I want to confine them to a single chapter, then replace altogether.</del>
* Remove a lot of stale information about thick Manifests.
* Mention that manifests can be regened via `repoman manifest` (over `ebuild ... manifest`).

### DIFF
--- a/general-concepts/manifest/text.xml
+++ b/general-concepts/manifest/text.xml
@@ -15,7 +15,9 @@ by the package. This is used to verify integrity upon fetching them.
 </p>
 
 <p>
-To generate the <c>Manifest</c>, use <c>ebuild foo.ebuild manifest</c>.
+To generate the <c>Manifest</c>, use <c>ebuild foo.ebuild manifest</c>
+or <c>repoman manifest</c>. You may want to set <c>GENTOO_MIRRORS=</c> while
+calling it to fetch distfiles from their original locations immediately.
 </p>
 </body>
 </section>
--- a/general-concepts/manifest/text.xml
+++ b/general-concepts/manifest/text.xml
@@ -7,68 +7,15 @@
 <title>Generating the Manifest</title>
 <body>
 <p>
-In the tree, every package has a <c>Manifest</c> file. This file lives in the same
-directory as the ebuilds for the package. The <c>Manifest</c> file contains digests
-(currently RMD160, SHA1, SHA256, SHA512 and WHIRLPOOL) and file size data for every
-file in the directory and any subdirectories. This is used to verify integrity.
-The <c>Manifest</c> may also be digitally signed.
+In the tree, every package has a <c>Manifest</c> file. This file lives
+in the same directory as the ebuilds for the package. The <c>Manifest</c> file
+contains digests (the current list can be found in <c>metadata/layout.conf</c>
+as <c>manifest-hashes</c>) and file size data for every distfile used
+by the package. This is used to verify integrity upon fetching them.
 </p>
 
 <p>
-To generate the <c>Manifest</c>, use <c>ebuild foo.ebuild manifest</c>. When
-committing, the <c>Manifest</c> file must be regenerated to handle any
-changes <d/> <c>repoman</c> will do this automatically.
-</p>
-</body>
-</section>
-
-<section>
-<title>Signing the Manifest using your GPG key</title>
-<body>
-<p>
-Requirements:
-</p>
-
-<ul>
-  <li>&gt;=sys-apps/portage-2.0.51_pre10</li>
-  <li>&gt;=app-crypt/gnupg-1.2.4</li>
-</ul>
-
-<p>
-Key Setup:
-</p>
-
-<ul>
-  <li>
-    <uri link="https://www.gentoo.org/doc/en/gnupg-user.xml#doc_chap2">Create</uri>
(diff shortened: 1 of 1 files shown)


ebuild-maintenance/maintenance-tasks/text.xml
```diff
@@ -60,9 +60,8 @@ examples.
 </p>
 
 <p>
-When committing to git, all developers should use <c>repoman commit</c>
-instead of <c>git commit</c> to submit their ebuilds.  Before committing,
-please run <c>repoman full</c> to make sure you didn't forget something.
+Before committing, please run <c>repoman full</c> to make sure you didn't forget
+something.
```
ulm: I think we should keep `repoman commit` here, when `repoman full` is mentioned in the same context.

general-concepts/ebuild-revisions/text.xml
```diff
@@ -52,8 +52,8 @@ of thumb could be used as a guideline:
     etc.) and it would not be propagated using other means, then
     the ebuild should be renamed to a new revision. If the package has
     stable keywords, they should be moved to the new revision without
-    dropping. To commit the ebuild, <c>repoman commit
-    --straight-to-stable</c> option should be used.
+    dropping. When using <c>repoman commit</c>, the <c>--straight-to-stable</c>
+    option can be used to disable stable keyword warnings.
```
ulm: Sorry, but what was wrong with the old wording?
Why are we going to discourage repoman now?

general-concepts/manifest/text.xml
```diff
@@ -7,68 +7,16 @@
...
-</p>
-
-<ul>
-  <li>
-    Set <c>PORTAGE_GPG_DIR</c> to your <c>~/.gnupg/</c> directory
-    (or the directory where the keyring with your new key is).
-  </li>
-  <li>Set <c>PORTAGE_GPG_KEY</c> to the key id of your new key.</li>
-  <li>Set FEATURES="sign".</li>
-</ul>
```
ulm: AFAICS, `PORTAGE_GPG_KEY` is not mentioned anywhere else. This section shouldn't be removed, but be updated for git signing instead.

ebuild-maintenance/maintenance-tasks/text.xml
mgorny: `repoman full` has the added value of verifying ebuilds. `repoman commit` has no advantage over `git commit`, besides teaching developers not to learn git.

general-concepts/manifest/text.xml
mgorny: But what does git signing have to do with Manifests?

ebuild-maintenance/maintenance-tasks/text.xml
ulm: Huh, but `repoman commit` does a verification step first:
> **commit**
> Scan directory tree for QA issues; if OK, commit via VCS

mgorny: …and the original tip said to run both. In any case, this is not the place to discuss committing tools. If you prefer, I can make that into `repoman full` OR `repoman commit`.

ulm: `repoman commit` will add the tag line to the commit message, and I believe this is what we generally recommend?

So, I'd just leave the first sentence in place, and replace the second sentence by a short explanation that `repoman full` can be used to get a longer listing.

general-concepts/manifest/text.xml
ulm: Yeah, remove it for now. `ebuild-maintenance/git/` already says something about openpgp signing, it only doesn't mention `PORTAGE_GPG_KEY`.

general-concepts/manifest/text.xml
```diff
@@ -7,68 +7,16 @@
 <title>Generating the Manifest</title>
 <body>
 <p>
-In the tree, every package has a <c>Manifest</c> file. This file lives in the same
-directory as the ebuilds for the package. The <c>Manifest</c> file contains digests
-(currently RMD160, SHA1, SHA256, SHA512 and WHIRLPOOL) and file size data for every
+In the tree, every package has a <c>Manifest</c> file. This file lives
+in the same directory as the ebuilds for the package. The <c>Manifest</c> file
+contains digests (currently BLAKE2B and SHA512) and file size data for every
```
gktrk: If it took us this long to update `currently RMD160, SHA1, SHA256, SHA512 and WHIRLPOOL` to `currently BLAKE2B and SHA512`, I'd say remove mentions to specific hash functions.

general-concepts/ebuild-revisions/text.xml
gktrk: Yeah, I reject this hunk too.

ebuild-maintenance/maintenance-tasks/text.xml
gktrk: I agree with ulm on this one as well.

mgorny:
Ok, let's focus on Manifests. I've replaced the hash list with reference to `layout.conf`. I've also added a section on thick Manifests.

ulm:
Merged to master.

PR #137, opened 2020-01-07
ebuild-writing/functions/src_install: Don't mention einstall.

ulm:
I have this small update in a local branch since 2015, but for some reason never submitted it for review.

### DIFF
--- a/ebuild-writing/functions/src_install/text.xml
+++ b/ebuild-writing/functions/src_install/text.xml
@@ -112,16 +112,14 @@ if you hit an error.
 </note>
 
 <p>
-  Sometimes this will end up installing a few things into strange
-  places. If and only if this is the case, the <c>einstall</c>
-  function can be used. It is usually necessary to include additional
-  <c>dodoc</c> statements for the <c>README</c>, <c>ChangeLog</c>, etc
-  in these cases:
+Usually the package's build system will not install the <c>README</c>,
+<c>ChangeLog</c>, etc. files, so it is necessary to include additional
+<c>dodoc</c> statements for them:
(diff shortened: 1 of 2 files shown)


ebuild-writing/functions/src_install/text.xml
```diff
@@ -112,16 +112,14 @@ if you hit an error.
...
-  in these cases:
+Usually the package's build system will not install the <c>README</c>,
+<c>ChangeLog</c>, etc. files, so it is necessary to include additional
+<c>dodoc</c> statements for them:
 </p>
 
 <codesample lang="ebuild">
-	einstall
-	dodoc README CHANGES
+    emake DESTDIR="${D}" install
```
gktrk: Isn't `${ED}` more future-proof? If `EPREFIX` is empty, it'll fall back to `D` anyway.

ulm:
> There's a reference to `einstall` in https://devmanual.gentoo.org/function-reference/build-functions/index.html as well. Should re remove the mention of that too? https://devmanual.gentoo.org/ebuild-writing/eapi/index.html#eapi-6-helpers already mentions that `einstall` is banned.

I've removed `einstall` from build-functions as well.

ebuild-writing/functions/src_install/text.xml
ulm: It should definitely be `${D}` when used with `emake install`, otherwise `${EPREFIX}` would be prefixed twice.

It is documented here:
https://archives.gentoo.org/gentoo-dev/message/2a62689c71f95e4de5699a330b8b5524

PR #138, opened 2020-01-08
Head and tail

ulm:
Having recently seen some bad code convinces me that we shouldn't drop the "head and tail" section.

This adds another (quite horrible) example.

### DIFF
--- a/tools-reference/head-and-tail/text.xml
+++ b/tools-reference/head-and-tail/text.xml
@@ -32,8 +32,8 @@ file. The <c>-n</c> argument specifies how many lines to display.
 </p>
 
 <p>
-To specify "the last five lines", use <c>tail -n5</c>. To specify "all
-but the first five lines", use <c>tail -n+6</c>.
+To specify "the last five lines", use <c>tail -n 5</c>. To specify "all
+but the first five lines", use <c>tail -n +6</c>.
 </p>
 
 <codesample lang="ebuild">
--- a/tools-reference/head-and-tail/text.xml
+++ b/tools-reference/head-and-tail/text.xml
@@ -33,7 +33,7 @@ file. The <c>-n</c> argument specifies how many lines to display.
 
 <p>
 To specify "the last five lines", use <c>tail -n5</c>. To specify "all
-but the first five lines", use <c>tail -n+5</c>.
+but the first five lines", use <c>tail -n+6</c>.
 </p>
 
 <warning>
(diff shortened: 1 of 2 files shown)


tools-reference/head-and-tail/text.xml
```diff
@@ -69,6 +69,18 @@ sed -n -e '1,5{ s/foo/bar/g ; p }' input.txt &gt; output.txt
 # on line 5
 sed -n -e 's/foo/bar/gp ; 5q' input.txt &gt; output.txt
 
+# bad: replace 'foo' with 'bar', and drop first five lines
+# (this one is from a real example!)
```
gktrk: do we really need this comment here? it being real or artificial adds little value to the point it is trying to make.

tools-reference/head-and-tail/text.xml
```diff
@@ -69,6 +69,18 @@ sed -n -e '1,5{ s/foo/bar/g ; p }' input.txt &gt; output.txt
...
+sed -e 's/foo/bar/' input.txt &gt; tmp.txt
+tail -n $(($(wc -l tmp.txt | awk '{print $1}') - 5)) tmp.txt &gt; output.txt
+rm -f tmp.txt
+
+# better: let tail count lines from the beginning of the file
```
gktrk: there is an ambiguity in this ranking. I think `bad < good < better` makes more sense then `bad < better < good`. In other words, I think better is better than good. Or we can just get rid of better.

tools-reference/head-and-tail/text.xml
```diff
@@ -32,8 +32,8 @@ file. The <c>-n</c> argument specifies how many lines to display.
 </p>
 
 <p>
-To specify "the last five lines", use <c>tail -n5</c>. To specify "all
-but the first five lines", use <c>tail -n+5</c>.
+To specify "the last five lines", use <c>tail -n 5</c>. To specify "all
```
gktrk: I personally prefer `-n5` over `-n 5` but I don't care either way. However, we should maintain some consistency if possible. The `cut` page also omits the whitespace. If we're gonna go with `-n 5`, we should fix `cut` to insert a whitespace as well.

tools-reference/head-and-tail/text.xml
ulm: I've dropped the comment.

tools-reference/head-and-tail/text.xml
ulm: Example moved to the first section. The main point is to illustrate that tail can start counting from beginning of the file.

tools-reference/head-and-tail/text.xml
ulm: POSIX Utility Syntax Guidelines (Guideline 6) say that the whitespace should be there:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02

I have fixed `cut` as well.

PR #139, opened 2020-01-08
devmanual.css: Drop unused elements.

ulm:
Most elements of the stylesheet are either not used at all, or they are superseded by the Bootstrap or Tyrian stylesheets which are loaded later.

### DIFF
--- a/devmanual.css
+++ b/devmanual.css
@@ -1,138 +1,6 @@
-div.header {
-    border-bottom: 3px solid #333333;
-    background-color: #666666;
-    color: #eeeeee;
-    margin: 0em 0em 0em 0em;
-    padding: 0.2em;
-    font-size: 70%;
-}
-
-div.navtop {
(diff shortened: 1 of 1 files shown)


ulm:
> I don't know enough about css to double check these changes. I cloned https://gitweb.gentoo.org/sites/tyrian-theme.git/, built the assets and tried to cross examine against `tyrian.css` and `bootstrap.css`. Is that the right way to go about it?

I've done a first check using the crawler at https://www.jitbit.com/unusedcss/, double checked the results with `devbook.xsl` (e.g., there is no `tt`, so removing `tt` from the CSS should be safe), and finally I examined which of the classes are overridden by Bootstrap or Tyrian.

I also spot-checked the HTML with the old and new CSS, and I don't see any change in its visual appearance.

ulm:
I have merges this now. Let's see if we'll have any fallout.

PR #140, opened 2020-01-09
ebuild-writing/.../metadata: clarify that upstream maintainer takes no 'type'

gktrk:
Closes: https://bugs.gentoo.org/703626
Closes: https://github.com/gentoo/devmanual/pull/140

### DIFF
--- a/ebuild-writing/misc-files/metadata/text.xml
+++ b/ebuild-writing/misc-files/metadata/text.xml
@@ -174,6 +174,8 @@ metadata.xml:
     Provides information about the upstream maintainer. It requires a
     &lt;name&gt; subtag to be specified, supports an optional
     &lt;email&gt; subtag and an optional <c>status</c> attribute.
+    Note that the <c>type</c> attribute <e>must not</e> be specified
+    for upstream maintainers.
   </ti>
 </tr>
 <tr>


ebuild-writing/misc-files/metadata/text.xml
```diff
@@ -174,6 +174,8 @@ metadata.xml:
     Provides information about the upstream maintainer. It requires a
     &lt;name&gt; subtag to be specified, supports an optional
     &lt;email&gt; subtag and an optional <c>status</c> attribute.
```
ulm: According to metadata.dtd, the `email` subtag is mandatory and must be listed first. It can be followed by one or more `description` or `name` subtags.

mgorny: GLEP 68 doesn't require any particular order. It's a bug in DTD.

ulm: https://www.gentoo.org/glep/glep-0068.html#maintainer-descriptions

The `<maintainer/>` element contains the following elements, in any order:
- exactly one `<email/>` element that contains the maintainer's e-mail address (used as unique identifier),
- at most one `<name/>` element that contains the maintainer's human-readable name (real name or nickname),
- zero or more `<description/>` elements that explain the role of the maintainer in different languages (at most one `<description/>` for each language).

So indeed, order is arbitrary (which I guess would qualify as a mistake in the DTD). Still, `email` is mandatory and `name` is optional, while the devmanual currently has it the other way around.

ulm: Oops, wrong section. I've missed https://www.gentoo.org/glep/glep-0068.html#upstream-maintainer-descriptions.

Why again are we using the same name for different things? To confuse people?
Also, in GLEP 46 the elements could be the same:
> The maintainer element can be the same as the top-level maintainer element in cases where a developer decides to maintain the package in addition to/instead of the original upstream.

mgorny: Hysterical raisins? If I were to guess, someone thought it was smart to reuse the same element structure.

PR #143, opened 2020-01-12
Make the devbook-guide describe the XML variant used in the devmanual

ulm:


appendices/devbook-guide/text.xml
```diff
@@ -40,137 +39,111 @@ tags used in a GuideXML document:
...
-about the various authors of the document.  Each <c>&lt;author&gt;</c> tag
-allows for an optional <c>title</c> element, used to specify the author's
-relationship to the document (author, co-author, editor, etc.).  In this
-particular example, the authors' names are enclosed in another tag <d/> a
-<c>&lt;mail&gt;</c> tag, used to specify an email address for this particular
-person. The <c>&lt;mail&gt;</c> tag is optional and can be omitted, and at
-least one <c>&lt;author&gt;</c> element is required per guide document.
+document. Next, there's a <c>&lt;guide&gt;</c> tag <d/> the entire document is
+enclosed within a <c>&lt;guide&gt; &lt;/guide&gt;</c> pair. Its <c>self</c>
+attribute must point to the relative path of the document from the root node.
```
gktrk: At this point, there should be some explanation as to why the example has `self="quickstart"`. I think even using `self="appendices/devbook-guide"` is better.

appendices/devbook-guide/text.xml
```diff
@@ -40,137 +39,111 @@ tags used in a GuideXML document:
...
 
 <p>
-Next, we come to the <c>&lt;abstract&gt;</c>, <c>&lt;version&gt;</c> and
-<c>&lt;date&gt;</c> tags, used to specify a summary of the document, the
-current version number, and the current version date (in YYYY-MM-DD format)
-respectively. Dates that are invalid or not in the YYYY-MM-DD format will
-appear verbatim in the rendered document.
+Next, there is a <c>&lt;chapter&gt;</c> tag. Every document must have exactly
+one chapter. Its <c>&lt;title&gt;</c> is used to set the title for the entire
+document.
```
gktrk: Should we mention that the body of the title shall not contain any newlines? e.g. don't do:
```
<title>
Foo
</title>
```

appendices/devbook-guide/text.xml
```diff
@@ -40,137 +39,111 @@ tags used in a GuideXML document:
...
-<p>
-Finally we have the <c>&lt;license version="3.0"/&gt;</c> tag, used to publish
-the document under the <uri link="https://creativecommons.org/licenses/by-sa/3.0/">Creative
-Commons - Attribution / Share Alike</uri> license as required by the <uri
-link="/proj/en/gdp/doc/doc-policy.xml">Documentation Policy</uri>. Historically,
-the tag <c>&lt;license /&gt;</c> was used, which denoted the 2.5 version of the
-license. This is still accepted/allowed.
+This sums up the tags that should appear at the beginning of a document.
+These tags must not appear anywhere else except immediately inside the
+<c>&lt;guide&gt;</c> tag.
```
gktrk: `<title/>` appears under `<section/>` too. if you meant for it to apply recursively, this sentence becomes very redundant as everything appears under the `<guide/>` tag anyway.

appendices/devbook-guide/text.xml
```diff
@@ -40,137 +39,111 @@ tags used in a GuideXML document:
...
 Once the initial tags have been specified, you're ready to start adding the
-structural elements of the document.  Guide documents are divided into
-chapters, and each chapter can hold one or more sections.  Every chapter and
-section has a title.  Here's an example chapter with a single section,
-consisting of a paragraph.  If you append this XML to the XML in the <uri
-link="#doc_chap2_pre1">previous excerpt</uri> and append a
-<c>&lt;/guide&gt;</c> to the end of the file, you'll have a valid (if minimal)
-guide document:
+structural elements of the document. Chapters are divided into sections, each
+section can hold one or more subsections, and so on, down to the subsubsection
```
gktrk: The expression of `and so on` is weak here since `subsubsection` is literally the next level from `subsection`, there are no intermediate levels that would have been captured by `and so on`. Also, a section can hold **zero** or more subsections, not **one**. How about:

```suggestion
section can hold zero or more subsections, which are comprised of zero or more subsubsections,
```

appendices/devbook-guide/text.xml
```diff
@@ -40,137 +39,111 @@ tags used in a GuideXML document:
...
+level. Section elements at any level have a title. Here's an example section
```
gktrk: It doesn't hurt to be explicit when you only need to list 3 items, I think.

```suggestion
Section, subsection, and subsubsection elements must have a title. Here's an example section
```

appendices/devbook-guide/text.xml
```diff
@@ -40,137 +39,111 @@ tags used in a GuideXML document:
...
+with a single subsection, consisting of a paragraph. If you append this XML to
+the XML in the <uri link="::appendices/devbook-guide/#Basic structure">previous
+excerpt</uri> and append <c>&lt;/chapter&gt;</c> and <c>&lt;/guide&gt;</c>
```
gktrk: Instead of saying "append this to the end", should we point out the closing tags in the previous section and not mention it here at all. The following example does not show it anyway.

appendices/devbook-guide/text.xml
```diff
@@ -40,137 +39,111 @@ tags used in a GuideXML document:
...
+it contains the actual text content of this particular subsection. We'll look
+at the tags that are allowed inside a <c>&lt;body&gt;</c> element in a bit.
 </p>
 
 <note>
-A <c>&lt;guide&gt;</c> element must contain at least one <c>&lt;chapter&gt;</c>
-elements, a <c>&lt;chapter&gt;</c> must contain at least one
-<c>&lt;section&gt;</c> elements and a <c>&lt;section&gt;</c> element must
-contain at least one <c>&lt;body&gt;</c> element.  
+Section elements at any level (i.e., <c>&lt;chapter&gt;</c>,
```
gktrk: Arguably, `<chapter/>` is not a section element. We can start the sentence with `The elements <c>&lt;chapter&gt;<c/>, <c>&lt;section&gt;<c/>, <c>&lt;subsection&gt;<c/>, and <c>&lt;subsubsection&gt;<c/> may contain a <c>&lt;body&gt;<c/>.` and so on.

appendices/devbook-guide/text.xml
```diff
@@ -452,14 +491,13 @@ admonitions. A definition list comprises:
...
-hyperlinks.  You can create a link pointing to <uri link="#doc_chap1">Chapter
-One</uri> by typing <c>&lt;uri link="#doc_chap1"&gt;Chapter
-One&lt;/uri&gt;</c>.  To point to <uri link="#doc_chap1_sect2">section two of
-Chapter One</uri>, type <c>&lt;uri link="#doc_chap1_sect2"&gt;section two of
-Chapter One&lt;/uri&gt;</c>.  To refer to figure 3 in chapter 1, type
-<c>&lt;uri link="#doc_chap1_fig3"&gt;figure 1.3&lt;/uri&gt;</c>.  Or, to refer
-to <uri link="#doc_chap2_pre2">code listing 2 in chapter 2</uri>, type
-<c>&lt;uri link="#doc_chap2_pre2"&gt;code listing 2.2&lt;/uri&gt;</c>.
+hyperlinks. You can create a link pointing to another chapter, like
+<uri link="::ebuild-writing/file-format/">Ebuild File Format</uri>, by typing
```
gktrk: Shouldn't we also demonstrate the more compact `<uri link="::ebuild-writing/file-format/"/>` format?

appendices/devbook-guide/text.xml
ulm: I've changed the example to use `appendices/devbook-guide/`, and added a short explanation.

appendices/devbook-guide/text.xml
ulm: Done.

appendices/devbook-guide/text.xml
ulm: I have deleted the sentence.

appendices/devbook-guide/text.xml
ulm: Done.

ulm: "Comprised of" isn't quite accurate, because a subsection can consist of the body alone.

I have dropped "and so on" and changed the wording to say "zero or more".

appendices/devbook-guide/text.xml
ulm: Done.

appendices/devbook-guide/text.xml
ulm: Done.

appendices/devbook-guide/text.xml
ulm: Done.

appendices/devbook-guide/text.xml
ulm: Good idea. Done.

ulm:
Should we rename the whole thing from "Gentoo GuideXML Guide" to "Gentoo DevBook XML Guide"?

gktrk:
> Should we rename the whole thing from "Gentoo GuideXML Guide" to "Gentoo DevBook XML Guide"?

:+1: I think we have sufficiently diverted from GuideXML to stop referring to it. Even if we didn't, GuideXML is hardly relevant at this point.

PR #145, opened 2020-01-20
Drop sections on obsolete EAPIs.

ulm:
This series of commits drops the sections on EAPIs 0 to 3 from `ebuild-writing/eapi`, and incorporates their info into other sections (unless this would be redundant).

The EAPI 4 section could either be dropped immediately, or as soon as the last EAPI 4 has left the tree.

Any thoughts? Is this approach feasible?


general-concepts/use-flags/text.xml
```diff
@@ -194,6 +194,38 @@ To get a flag added to <c>use.mask</c>, ask the relevant arch team.
...
+</p>
+
+<important>
+The default USE-ordering is
+<c>USE_ORDER="env:pkg:conf:defaults:pkginternal:features:repo:env.d"</c>
+(see man make.conf).
+</important>
+
+<important>
+Disabling default <c>IUSE</c> is pretty much useless as it does not override
```
mgorny: I know you just moved it but I really have no clue what this means.

ebuild-writing/functions/src_prepare/text.xml
```diff
@@ -31,13 +31,10 @@
...
 <body>
-<p>
-Starting from EAPI=2, the src_prepare function is the appropriate area to perform
-any kind of patching and source code manipulation, instead of src_unpack.
-</p>
+
 <codesample lang="ebuild">
 src_prepare() {
-	true;
+	true
```
mgorny: Maybe use `:` if you're already changing it.

ebuild-writing/misc-files/metadata/text.xml
```diff
@@ -280,9 +280,10 @@ There are also some attributes that can be used with these tags:
     is used, other tags with or without the restrict attribute must be
     present to cover all the versions of the package. A tag without
     the restrict attribute serves as the default. The format of the
-    restrict attribute is that of a EAPI=0 package dependency
-    specification. Due to the limitations of XML, the "&lt;" and
-    "&gt;" need to be specified using "&amp;lt;" and "&amp;gt;".
+    restrict attribute is that of a EAPI=0 package dependency specification
+    (i.e., USE-conditional or slot dependendencies are not allowed).
+    Due to the limitations of XML, the <c>&lt;</c> and <c>&gt;</c> need to
```
mgorny: Since you are already changing it, maybe replace 'limitations' with 'syntax' or sth like that. It's weird to call it a limitation if that's intended syntax.

function-reference/build-functions/text.xml
```diff
@@ -24,7 +24,27 @@ during the unpack and compile stages.
       <c>unpack archives</c>
     </ti>
     <ti>
-      Unpack the specified archives.
+      <p>
+        Unpack the specified archives. Recognised formats are:
```
mgorny: Aren't we using AmE spelling in devmanual, i.e. `recognized`?

function-reference/build-functions/text.xml
```diff
@@ -24,7 +24,27 @@ during the unpack and compile stages.
...
+          <c>*.tar.gz</c>, <c>*.tgz</c>, <c>*.tar.Z</c>
+        </li>
+        <li>
+          <c>*.bz2</c>, <c>*.bz</c>,
+          <c>*.tar.bz2</c>, <c>*.tbz2</c>, <c>*.tar.bz</c>, <c>*.tbz</c>
+        </li>
+        <li><c>*.lzma</c>, <c>*.tar.lzma</c></li>
+        <li><c>*.xz</c>, <c>*.tar.xz</c>, <c>*.txz</c></li>
+        <li><c>*.zip</c>, <c>*.jar</c></li>
+        <li><c>*.7z</c></li>
```
mgorny: Maybe it'd be useful to mention for which formats you need extra deps.

general-concepts/use-flags/text.xml
ulm: It means that `pkginternal` has lower precedence than `conf` and `defaults`.
I'll try to explain it in an understandable way.

ebuild-writing/functions/src_prepare/text.xml
ulm: I had considered it, but then decided against. `true` is easier to understand for a beginner than a lone `:`.

ebuild-writing/misc-files/metadata/text.xml
ulm: Right. Also, a sentence should never begin with "due to".
Fixed.

function-reference/build-functions/text.xml
ulm: https://devmanual.gentoo.org/appendices/contributing/index.html#style-guidelines
> This document is in British English. Submissions in other kinds of English are welcome, but they may have their spelling corrected. 

Anyway, Macmillan lists both spellings, so I've changed it to the "ize" variant.

function-reference/build-functions/text.xml
ulm: Not sure. It is the "function reference" section, so I suppose that explanations should be brief. Maybe move the whole list to `ebuild-writing/functions/src_unpack`, with a longer explanation?

mgorny: Makes sense.

ulm: I've moved it to the other section, but have only added an `<important>` box. We can add more information later, but this is not the goal of this PR and shouldn't block it.

ebuild-writing/functions/src_compile/text.xml
```diff
@@ -33,67 +33,40 @@
...
 <codesample lang="ebuild">
 src_compile() {
 	use sparc &amp;&amp; filter-flags -fomit-frame-pointer
 	append-ldflags -Wl,-z,now
-
-	econf \
-		$(use_enable ssl ) \
-		$(use_enable perl perlinterp )
-
 	emake || die "Make failed!"
```
gktrk: There should be no `|| die` here

ebuild-writing/functions/src_prepare/text.xml
```diff
@@ -31,13 +31,10 @@
 <section>
 <title>Default <c>src_prepare</c></title>
 <body>
-<p>
-Starting from EAPI=2, the src_prepare function is the appropriate area to perform
-any kind of patching and source code manipulation, instead of src_unpack.
-</p>
+
```
gktrk: We should state here that the following applies to EAPI < 6

ebuild-writing/functions/src_unpack/text.xml
```diff
@@ -79,6 +72,45 @@ usually simpler to avoid working with <c>${A}</c>.
...
+  <li><c>*.LHA</c>, <c>*.LHa</c>, <c>*.lha</c>, <c>*.lzh</c></li>
+</ul>
+
+<p>
+In EAPI 6 and later, filename extensions are matched case-insensitively.
+</p>
+
+<important>
+Unless the utility needed for unpacking is in the system set, the ebuild must
+have a dependency on it.
```
gktrk:
```suggestion
specify the necessary build time dependency for it.
```

general-concepts/dependencies/text.xml
```diff
@@ -218,54 +215,68 @@ to the package manager.
...
 installing the package blocking it. However, it exempts the common
-files from file collision checks. Soft blockers are usually used
+files from file collision checks. Weak blockers are usually used
 to solve file collisions between packages and are meaningful only
 in <c>RDEPEND</c>.
 </p>
 
+<p>
+More specifically, installation of the newer package may overwrite any
+colliding files that belong to the older package which is explicitly blocked.
```
gktrk:
```suggestion
colliding files that belong to the older package that is explicitly blocked.
```

general-concepts/dependencies/text.xml
```diff
@@ -218,54 +215,68 @@ to the package manager.
...
-soft blockers in <c>RDEPEND</c>!
+weak blockers in <c>RDEPEND</c>!
 </warning>
 
 <p>
 If it is strictly necessary to resolve the blocker before the package
-is built (installed), a hard blocker must be used instead. Hard blockers
-are expressed using the following syntax:
+is built (installed), a strong blocker must be used instead. In this case,
+temporary simultaneous installation of the conflicting packages not allowed.
```
gktrk:
```suggestion
temporary simultaneous installation of the conflicting packages is not allowed.
```

general-concepts/use-flags/text.xml
```diff
@@ -194,6 +194,39 @@ To get a flag added to <c>use.mask</c>, ask the relevant arch team.
...
+
+<p>
+Add <c>+</c> or <c>-</c> before the name of the use flag in <c>IUSE</c> to turn
+it on or off by default.
+</p>
+
+<important>
+The default USE-ordering is
+<c>USE_ORDER="env:pkg:conf:defaults:pkginternal:features:repo:env.d"</c>
+with the highest precedence listed first (see man make.conf).
```
gktrk: I'm not sure if this is explanatory enough. We should either import the whole text from make.conf and not mention it IMO.

ebuild-writing/functions/src_compile/text.xml
ulm: Done.

ebuild-writing/functions/src_prepare/text.xml
ulm: Done.

ebuild-writing/functions/src_unpack/text.xml
ulm: Done.

general-concepts/dependencies/text.xml
ulm: Updated. (The "which" was in the text that I've copied, though.)

general-concepts/dependencies/text.xml
ulm: Done.

general-concepts/use-flags/text.xml
ulm: I have deleted the first `important` box, and moved the reference to make.conf(5) to the second one.

PR #146, opened 2020-01-21
Split "Maintenance Tasks" into separate chapters

ulm:
This follows up to the incomplete split of #132. It also includes the first commit of that PR, in order to avoid later merge conflicts.


ulm:
> Thanks for doing this! I presume you're asking for review of the moves, and not detailed review of old text ;-).

In fact, no changes in the text for the new chapters, except:
- Promote section to chapter etc., if the old section is the only section in the new chapter
- Title Case for chapter titles

ulm:
> How about we go alphabetical?

> Also, instead of saying "Conflicting Files", maybe we should say "Package Collisions" ?

Both done.

PR #147, opened 2020-01-23
keywording: Include sections about stabilization

ulm:
This tries to address the duplication between these two sections:
https://devmanual.gentoo.org/keywording/index.html
https://devmanual.gentoo.org/ebuild-maintenance/stabilization/index.html

The stabilization section originates from the devrel handbook (https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo/xml/htdocs/proj/en/devrel/handbook/hb-guide-ebuild-maintaining.xml?hideattic=0&revision=1.23&view=markup) being incorporated in the devmanual. Apparently not all duplications were cleaned up at that time.


keywording/text.xml
```diff
@@ -267,6 +291,38 @@ Vulnerability Treatment Policy</uri>
...
+
+<p>
+ALPHA: Maintainers may keyword their own packages but are reminded to inform
+the Alpha team if they can help out with testing and keywording packages so
+the team can keep an eye out for possible keywording mistakes.
+</p>
+
+<p>
+Exotic architectures (like alpha, ia64, sparc, hppa, ppc*) are short
+on manpower, so it's best if you avoid opening stabilization bugs for
```
SoapGentoo: maybe remove alpha from this section, as I will stop having stable keywords on the 25th.

keywording/text.xml
```diff
@@ -267,6 +291,38 @@ Vulnerability Treatment Policy</uri>
...
+them unless it is absolutely necessary (eg, a reverse dependency for
```
SoapGentoo: isn't it `e.g.` usually?

keywording/text.xml
```diff
@@ -267,6 +291,38 @@ Vulnerability Treatment Policy</uri>
...
+your package).
+</p>
+
+<p>
+Some architectures (like m68k, mips, s390, sh) do not maintain a
+stable keyword. So there is no use in marking a package stable for one
```
SoapGentoo: "do not maintain stable profiles" maybe?

keywording/text.xml
```diff
@@ -267,6 +291,38 @@ Vulnerability Treatment Policy</uri>
 
 </body>
 
+<subsection>
+<title>Stabilization rules</title>
+<body>
+
+<p>
+SPARC: You must have prior permission from the arch lead. Usually we expect
```
mgorny: Maybe we should verify whether this is up-to-date. With sparc going exp, and dakon being the only person to stabilize...

keywording/text.xml
mgorny: That's going to be confusing. It should talk of adding stable keywords to new packages; we still want people to stabilize new versions of existing packages.

keywording/text.xml
mgorny: Make it stronger. People really must not add stable keywords there.

ulm: I think it's keyword here, because stability of the profile doesn't necessarily coincide with it.

SoapGentoo: ...but `m68k` has stable keywords (:wave: vapier)?

keywording/text.xml
SoapGentoo: same goes for `hppa` in that case

keywording/text.xml
ulm: Done.

keywording/text.xml
ulm: Done.

keywording/text.xml
ulm: Can we postpone this to a separate commit, and not conflate it with merging of sections?

keywording/text.xml
ulm: I've changed it to say "avoid opening bugs for stabilization of new packages".

keywording/text.xml
ulm: Changed to: "So packages are not to be marked stable for one of these architectures."

ulm: Changed to say only mips, riscv.

PR #151, opened 2020-02-06
Makefile: New target "delete-old"

ulm:
After the "maintenance-tasks" subsection was split up some time ago, the generated HTML file for it still exists: https://devmanual.gentoo.org/ebuild-maintenance/maintenance-tasks/index.html

So, delete any orphaned html and image files that could be left over after the corresponding source file was moved or removed. Such files are also not covered by `clean`, so add the new target as prerequisite.

Run a single `find` pass to get a list of all files in `ALL_FILES` and subsequently extract the necessary subsets from it. The list of all files is needed anyway for deletion of orphans, which are not included in `HTMLS` or `IMAGES`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,16 @@
-# These "find" commands match text.xml and *.svg files, respectively,
-# but only after excluding the .git directory from the search for
-# performance and overall sanity reasons.
-XMLS := $(shell find . -name .git -prune -o -type f -name 'text.xml' -print)
-SVGS := $(shell find . -name .git -prune -o -type f -name '*.svg' -print)
+# Run a single "find" pass to get a list of all files (with the .git
+# directory excluded), then filter out what we need.
+ALL_FILES := $(shell find . -name .git -prune -o -type f -print)
+XMLS := $(filter %/text.xml,$(ALL_FILES))
+SVGS := $(filter %.svg,$(ALL_FILES))
 HTMLS := $(subst text.xml,index.html,$(XMLS))
-ECLASS_HTMLS := $(wildcard eclass-reference/*/index.html)
+ECLASS_HTMLS := $(filter ./eclass-reference/%/index.html,$(ALL_FILES))
 IMAGES := $(patsubst %.svg,%.png,$(SVGS))
 
 # Nonzero value disables external assets for offline browsing.
 OFFLINE = 0
 
-all: prereq validate build documents.js
+all: prereq validate delete-old build documents.js
 
 prereq:
 	@type rsvg-convert >/dev/null 2>&1 || \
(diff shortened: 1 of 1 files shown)


orlitzky:
Since this should only happen in a git repo (where else are you removing source files?), isn't this covered by `git clean -x -f -d`?

ulm:
> Since this should only happen in a git repo (where else are you removing source files?), isn't this covered by `git clean -x -f -d`?

That doesn't do what we want. Without the `-x` option, it won't remove any `*.html` and `*.png` because they're listed in `.gitignore`. And with `-x` it will remove all such files, even if there is a corresponding source file.

orlitzky:
> And with `-x` it will remove all such files, even if there is a corresponding source file.

But that's what `make clean` does now, and unless you're really on top of things, you're typically going to be using `make clean`. A possibly better option is to fix the gitignore file so that it doesn't ignore html/png files that shouldn't be there.

In any case, I don't object, I'm just wondering if there's an easier way to accomplish it.

ulm:
That `make clean` will work properly (again) is only a side effect.

The primary goal of this PR is to call the `delete-old` target as part of `make all` on the infra host. And for this, we don't want to remove any existing index.html files unless their corresponding source file is gone.

orlitzky:
Better, but it still deletes all of my png files with the default `make`.

orlitzky:
With the current image names you can work around that by changing that one filter to e.g. `$(filter %/index.html %/diagram.png,$(ALL_FILES)))`. Not great but I'd rather be over-conservative when deleting things than the other way around.

mgorny:
I agree that this looks weird, to say lightly. It's not like devmanual takes hours to build, and we can't live without incremental builds. In this case, it's entirely fine for infra to run `git clean -dfx` or whatever clean-all target we expect.

A better long-term solution would involve stopping to intersperse input and output files.

ulm:
> I agree that this looks weird, to say lightly. It's not like devmanual takes hours to build, and we can't live without incremental builds. In this case, it's entirely fine for infra to run `git clean -dfx` or whatever clean-all target we expect.

Really, the last thing that we need is additional scripting on the infra side. Eclass reference pages were broken for years, until we moved the generating script into the devmanual repo and fixed it ourselves.

ulm:
> With the current image names you can work around that by changing that one filter to e.g. `$(filter %/index.html %/diagram.png,$(ALL_FILES)))`. Not great but I'd rather be over-conservative when deleting things than the other way around.

There's no such rule that would require images to be named `diagram.svg`.

orlitzky:
> There's no such rule that would require images to be named `diagram.svg`.

Like I said, not great, but it works for now.

Instead we are back in a situation where a standard GNU target `make all` will delete random PNG files on your system, contrary to its purpose and user expectations.

ulm:
> > There's no such rule that would require images to be named `diagram.svg`.
> 
> Like I said, not great, but it works for now.
> 
> Instead we are back in a situation where a standard GNU target `make all` will delete random PNG files on your system, contrary to its purpose and user expectations.

Yeah, let's wait for a day so it can run once. :grin: Then I'll remove it from `all` again. (Maybe infra could automatically run `make delete-old` once in a while?)

On the long term, a cleaner solution would be to have a separate install directory. Probably one would create an empty dir, install into it, and move it quasi-atomically to the real location.

It's also a bit strange to have the whole `.git` directory at https://devmanual.gentoo.org/ (although it's practical that one can find the last commit id at https://devmanrual.gentoo.org/.git/refs/heads/master).

orlitzky:
> On the long term, a cleaner solution would be to have a separate install directory. Probably one would create an empty dir, install into it, and move it quasi-atomically to the real location.

I've hacked `PREFIX` and `DESTDIR` into a bare Makefile before to get this working.

In a perfect world, infra would be emerging app-doc/devmanual-9999 rather than pulling from the repo. That ensures a clean build and something like an atomic update.

mgorny:
I could make it do that but calling the (slow, error-prone) package manager from update script seems wrong.

mgorny:
That said, ebuild looks like a giant hack anyway. Building into separate builddir/outdir is the solution.

PR #152, opened 2020-02-07
Makefile: Add install target.

ulm:
As discussed in #gentoo-qa. This will allow to much simplify the app-doc/devmanual ebuild.

On the long term, it will enable infra to install into a separate target directory.


ulm:
Note to self: Once we're using separate build and install dirs, there will no longer be any need for the intermediate `_documents.js` file.

PR #153, opened 2020-02-09
ebuild-writing/eapi: Move EAPI=4 information to other sections.

ulm:
- utilities die, nonfatal: new section in ebuild-writing/error-handling, new chapter function-reference/error-functions
- dodoc -r: already in function-reference/install-functions, add additional example to ebuild-writing/functions/src_install
- doins symlink support: incorporated into function-reference/install-functions
- dosed, dohard banned: not mentioned anywhere else
- econf --disable-dependency-tracking: new section about econf options in ebuild-writing/functions/src_configure/configuring
- controllable compression and docompress: new chapter ebuild-writing/functions/src_install/docompress

- USE dependency defaults: incorporated into general-concepts/dependencies

- pkg_pretend: already in ebuild-writing/functions/pkg_pretend
- default src_install: already in ebuild-writing/functions/src_install
- pkg_info: incorporated into ebuild-writing/functions/pkg_info

- REQUIRED_USE: already in ebuild-writing/variables
- MERGE_TYPE, REPLACING_VERSIONS, REPLACED_BY_VERSION: moved to ebuild-writing/variables
- DOCS: incorporated into ebuild-writing/variables
- AA, KV removed: not mentioned anywhere else
- RDEPEND=DEPEND: already handled with EAPI=3


ebuild-writing/functions/src_configure/configuring/text.xml
```diff
@@ -74,5 +74,56 @@ form can be used.
...
+  <li>
+    <c>--libdir</c> is set from the value of the <c>LIBDIR_${ABI}</c> variable
+    in profiles.
+  </li>
+  <li><c>--disable-dependency-tracking</c></li>
+  <li><c>--disable-silent-rules</c></li>
+</ul>
+
+<p>
+In EAPI 6 and later, the following options are passed in addition:
```
mgorny: I'm not sure if 'in addition' doesn't require the subject.

ebuild-writing/functions/src_install/docompress/text.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0"?>
+<guide self="ebuild-writing/functions/src_install/docompress/">
+<chapter>
+<title>Controllable Compression</title>
+<body>
+
+<p>
+To compress files in the destination folder <c>${D}</c>, the <c>docompress</c>
```
mgorny: I think this wrongly suggests that docompress does the actual compression and/or is required for the compression to happen. Maybe 'The docompress function can be used to control ....'

ebuild-writing/functions/src_configure/configuring/text.xml
ulm: > I'm not sure if 'in addition' doesn't require the subject.

I don't think so: https://www.macmillandictionary.com/dictionary/british/in-addition

ebuild-writing/functions/src_install/docompress/text.xml
ulm: Good point, I'll fix (although I've just move that text).

ebuild-writing/variables/text.xml
```diff
@@ -153,6 +153,33 @@ for these variables when writing ebuilds.
 		satisfied by <c>BDEPEND</c>, typically executable build tools.
     </ti>
   </tr>
+  <tr>
+    <ti><c>MERGE_TYPE</c></ti>
+    <ti>
+      The type of package that is being merged. Possible values are:
+      <c>source</c> if building and installing a package from source,
+      <c>binary</c> if installing a binary package,
```
gktrk: Should we further clarify what we mean by a binary package? They apply to Gentoo-specific binary packages that are built using a PMS-compliant package manager. If you are merging `firefox-bin`, the merge type would still be `source` unless there is a binary package for `firefox-bin` is available somewhere. Thoughts?

ebuild-writing/variables/text.xml
```diff
@@ -153,6 +153,33 @@ for these variables when writing ebuilds.
...
+      install. It is a list, not a single optional value, to handle pathological
+      cases such as installing <c>foo-2:2</c> to replace <c>foo-2:1</c> and
+      <c>foo-3:2</c>. Available in <c>pkg_preinst</c> and <c>pkg_postinst</c>.
+    </ti>
+  </tr>
+  <tr>
+    <ti><c>REPLACED_BY_VERSION</c></ti>
+    <ti>
+      The single version (<c>PVR</c>) of this package that is replacing us,
+      if we are being uninstalled as part of an install, or an empty string
```
gktrk:
```suggestion
      from the perspective of the version being uninstalled as part of an install, or an empty string
```

ebuild-writing/variables/text.xml
ulm: We should, but this table isn't the right place for it. Presumably, we should have a dedicated section for binary packages (and there could be a link to it in the table).

Also, this text was there before, and I've only moved it to a new place.

ebuild-writing/variables/text.xml
ulm: Hm, but then the "us" in the line before should also be replaced. How about:
"The single version (<c>PVR</c>) of this package that is replacing this ebuild's version as the result of an update, or an empty string otherwise."

gktrk: instead of saying `ebuild's version`, would it be better to say `the version provided by this ebuild` or `the version associated with this ebuild` ? Also, is there a particular motivation behind choosing `update` in the earlier part of the sentence and `uninstall` at the end.

ebuild-writing/variables/text.xml
gktrk: I propose as a quick fix: s/a binary package/a Gentoo binary package/. What do you think?

ulm: "a previously built package"?

mgorny: How about:

```suggestion
      <c>binary</c> if installing a binary package built from the ebuild,
```

that should make it clear what kind of binary package we're talking about. I don't think it's a good idea to try to find a roundabout name for it, as that will only confuse people.

ebuild-writing/variables/text.xml
ulm: Right, "update" doesn't catch it, because it could also being a replacement by the same version.

"The single version (`PVR`) of this package that is replacing the version provided by this ebuild, if it is being uninstalled as part of an install. An empty string otherwise, i.e., if it is being uninstalled without replacement."

ebuild-writing/variables/text.xml
ulm: Even better, but I'd still say "**previously** built from the ebuild" to make it clear that we're not building it as part of this cycle.

mgorny:
> Even better, but I'd still say "**previously** built from the ebuild"
to make it clear that we're not building it as part of this cycle.

Sure, that LGTM.

ebuild-writing/variables/text.xml
ulm: It is a bit like in this classic from Morris Kline, "Mathematics in Western Culture": :smile:
> When a twelfth-century youth fell in love he did not take three paces backward, gaze into her eyes, and tell her she was too beautiful to live. He said he would step outside and see about it. And if, when he got out, he met a man and  broke his head—the other man's head, I mean—then that proved that his—the  first fellow's—girl was a pretty girl. But if the other fellow broke *his* head—not his own, you know, but the other fellow's—the other fellow to the second fellow, that is, because of course the other fellow would only be the other fellow to him, not the first fellow who—well, if he broke his head, then *his* girl—not the other fellow's, but the fellow who *was* the—
Look here, if A broke B's head, then A's girl was a pretty girl; but if B broke A's head, then A's girl wasn't a pretty girl, but B's girl was.

gktrk:
LGTM

PR #154, opened 2020-02-16
general-concepts/slotting: Clarify that SLOT must not be empty.

ulm:
Not sure from where the concept "an empty slot means disable slotting" originated. In any case, it is wrong, and it was not legal in any EAPI.


mgorny:
I think it originated in pre-EAPI times. I recall learning about it at some point (recruiting?). Not sure if it's still implemented in Portage.

PR #156, opened 2020-02-22
ebuild-writing/functions/src_unpack/vcs-sources: New file

ulm:
`cvs-sources` and `svn-sources` contain duplicate information, and it also applies to other VCSs. Combine them in a single chapter.


ulm:
Or should we just drop the specific "Using CVS Sources" and "Using Subversion Sources" sections? We refer to the eclass documentation anyway, and could leave it at that.

gktrk:
We made the same comments within 1 second of each other :D. Let's kill it with fire!

ulm:
> I'm tempted to move VCS-specific (CVS and SVN) documentation to their respective eclasses and not keep it in devmanual.

Heh, comments just crossed. :smile: But let's see if there are other opinions.

appendices/contributors/text.xml
```diff
@@ -23,7 +23,7 @@ This page lists the contributions to the Gentoo Development Guide:
 </author>
 <author name="Tom Martin" email="slarti@gentoo.org">
   <uri link="::appendices/editor-configuration/emacs"/>,
-  <uri link="::ebuild-writing/functions/src_unpack/svn-sources"/>,
+  Subversion Sources (removed),
```
gktrk: Any reason why we wouldn't just remove the reference to the specific contribution that doesn't exist anymore? If we do it this way, the list will forever grow and become a mess, no?

SoapGentoo:
sorry, wanted to suggest an edit, not actually commit it, I failed at GH :confused:

gktrk:
> sorry, wanted to suggest an edit, not actually commit it, I failed at GH

what a shameful display ...

appendices/contributors/text.xml
ulm: No strong opinion here. Let's remove it.

ulm:
> sorry, wanted to suggest an edit, not actually commit it, I failed at GH confused

Heh, I changed this from "can" to "should", then back to "can".

PR #159, opened 2020-02-25
Revise the navbar items

mmagorsc:
The navbar links have been shortened using css resulting in a
fixed navbar size. In addition the searchbar in the lower navbar
has been centered vertically.

Signed-off-by: Max Magorsch <arzano@gentoo.org>

### DIFF
--- a/devbook.xsl
+++ b/devbook.xsl
@@ -677,7 +677,7 @@
       -->
       <xsl:when test="count(/guide/include) &gt; 0">
         <xsl:variable name="doc" select="/guide/include[1]/@href"/>
-        <a href="{concat($doc, 'index.html')}"><xsl:value-of select="document(concat(/guide/@self, $doc, 'text.xml'))/guide/chapter[1]/title"/> &#160;<span class="fa fa-arrow-right"/></a>
+        <a class="w-200 text-center" href="{concat($doc, 'index.html')}"><span class="truncated-text d-inline-block max-w-150 mr-1"><xsl:value-of select="document(concat(/guide/@self, $doc, 'text.xml'))/guide/chapter[1]/title"/></span><span class="fa fa-arrow-right"/></a>
       </xsl:when>
       <xsl:otherwise>
         <!-- This document's path -->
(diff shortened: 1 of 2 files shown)


devbook.xsl
```diff
@@ -710,7 +700,7 @@
                 <xsl:with-param name="append">../</xsl:with-param>
               </xsl:call-template>
             </xsl:variable>
-            <a href="{concat($relative_path_depth_recursion, $relative_path, 'index.html')}"><xsl:value-of select="document(concat($parentItem_actual, 'text.xml'))/guide/chapter[1]/title"/> &#160;<span class="fa fa-arrow-right"/></a>
+            <a class="w-200 text-center" href="{concat($relative_path_depth_recursion, $relative_path, 'index.html')}"> <span class="truncated-text d-inline-block max-w-150 mr-1"><xsl:value-of select="document(concat($parentItem_actual, 'text.xml'))/guide/chapter[1]/title"/></span><span class="fa fa-arrow-right"/></a>
```
ulm:
```suggestion
            <a class="w-200 text-center" href="{concat($relative_path_depth_recursion, $relative_path, 'index.html')}"> <span class="truncated-text d-inline-block max-w-150 mr-1"><xsl:value-of select="document(concat($parentItem_actual, 'text.xml'))/guide/chapter[1]/title"/></span>&#160;<span class="fa fa-arrow-right"/></a>
```
This looks somewhat cramped, so I'd suggest to leave the space in place.
Same for `fa-arrow-left` below (three times).

devmanual.css
```diff
@@ -33,4 +33,44 @@ pre span.Constant { color: black; }
...
+    vertical-align: top;
+}
+
+.w-200 {
+    width: 200px;
+}
+
+.max-w-150 {
+    max-width: 150px;
+}
```
ulm: As discussed on IRC, 250px and 200px should still work.

devbook.xsl
mmagorsc: I have replaced ml-1 and mr-1 by ml-2 and mr-2. This way it should look pretty much as before now.

ulm:
Merged.
I have squashed the third commit with the first one.

PR #160, opened 2020-02-26
Makefile, depend.xsl: Use XSLT to generate the list of dependencies.

ulm:
Each HTML file must depend on its XML file with all its descendants (for the contents tree), all its ancestors (for breadcrumbs), and the previous and next documents (for backward and forward links).

Trying to generate the list of dependencies with make seems hopeless (especially, how would one determine the previous and next documents?). OTOH, the necessary templates already exist in devbook.xsl. Therefore, add a simple depend.xsl on top of it, which outputs a dependency list as plain text.

Closes: https://bugs.gentoo.org/710918

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.html
 *.png
+.depend
 documents.js
 eclass-reference/
--- a/Makefile
+++ b/Makefile
@@ -47,19 +47,22 @@ documents.js: bin/build_search_documents.py $(XMLS)
 	rsvg-convert --output=$@ $<
 
 # Secondary expansion allows us to use the automatic variable $@ in
-# the prerequisites. When it is used (and we have no idea when that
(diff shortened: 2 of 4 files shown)


orlitzky:
Another thing to keep in mind is that, since `.depend` depends on basically everything, you wind up regenerating it whenever any file changes. That's relatively quick (about 0.2s here), but a full build of every HTML file takes only about three seconds. The more complicated the dependencies get, the more attractive it is to just rebuild all of the HTML files whenever any text.xml changes.

ulm:
> The `.depend` target should depend on `depend.xsl`,`devbook.xsl`, and `xsl/*.xsl`.

I have added `depend.xsl` and `devbook.xsl`, but I've left out `xsl/*.xsl`. We need not depend on language highlighting.

PR #161, opened 2020-02-27
bin/gen-eclass-html.sh: New option -n to create a placeholder file.

ulm:
The new strict dependencies will cause the build to fail if eclass-reference/text.xml is missing, because it is needed as a dependency for index.html of the previous and next documents. Add an option that makes it possible to create the file, even if the eclass documentation is missing.

### DIFF
--- a/bin/gen-eclass-html.sh
+++ b/bin/gen-eclass-html.sh
@@ -100,8 +100,35 @@ guesscompress() {
 	esac
 }
 
-# We also need the ebuild man page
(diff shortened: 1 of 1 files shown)


mgorny:
Could you document the option somewhere? A comment on top of the file would suffice, or a `-h`(elp) option.

ulm:
> Could you document the option somewhere? A comment on top of the file would suffice, or a `-h`(elp) option.

Done.

PR #166, opened 2020-04-07
general-concepts/licenses: Mention the SPDX list and its caveats.

ulm:


hannob:
My personal opinion would be that I would rather have something that is stupidly named or excessively long and standardized than not standardized. Though I would soften the caveat.

ulm:
Partly, it is different goals. Their goal is to stick one identifier into a source file, so they have no problem with "short" IDs like `GPL-2.0-or-later` or `BSD-3-Clause-No-Nuclear-License-2014`. Whereas in our case we list several identifiers in `LICENSE` (or in users' `ACCEPT_LICENSE` or `package.license`). So we prefer the ID to be short, like `GPL-2+`.

Also (as I have said previously) we will always have identifiers that are not in the SPDX's list, identifiers that we had earlier, and identifiers where we agreed with the SPDX but where they have renamed. Therefore we will never have a 100% matching. One way out of it would be a mapping table between our and their names (as I had also suggested previously), but nobody has followed up to this.

general-concepts/licenses/text.xml
```diff
@@ -239,6 +239,15 @@ only do so if the license could be considered 'questionable' or if
 you are unsure as to the meaning of any part of it.
 </p>
 
+<p>
+When adding a new license, it should be checked whether there is an established
+name in the <uri link="https://spdx.org/licenses/">SPDX license list</uri>.
+However, Gentoo does not consider the Software Package Data Exchange to be an
+authoritative standard, so we may deviate from it if their "short identifier"
+is excessively long or stupid. Also, we generally don't rename our existing
+identifiers.
```
ulm, line 248: How about this:

```suggestion
authoritative standard, so we may sometimes deviate from it, e.g., if their
"short identifier" is excessively long. Also, we generally don't rename our
existing identifiers.
```

ulm:
> Tangentially related:
> 
> * I'd like to see a paragraph as well that mentions the `license` API of GitHub is limited to returning a single license (it does return a SPDX id for it) for a repo, whereas actual packages might use different licenses in different files.

We already have the following above, in the _"Determining the correct license"_ section:
> _**Warning:** Do not trust the license indicated by GitHub or other sites that use automation to guess the license! The information presented may be incomplete and/or incorrect._

Do you think this should be extended?

robbat2:
> > Tangentially related:
> > 
> > * I'd like to see a paragraph as well that mentions the `license` API of GitHub is limited to returning a single license (it does return a SPDX id for it) for a repo, whereas actual packages might use different licenses in different files.
> 
> We already have the following above, in the ["Determining the correct license"](https://devmanual.gentoo.org/general-concepts/licenses/index.html#determining-the-correct-license) section:
> 
> > _**Warning:** Do not trust the license indicated by GitHub or other sites that use automation to guess the license! The information presented may be incomplete and/or incorrect._
> 
> Do you think this should be extended?
Yes: I'll open a separate PR for it.

PR #168, opened 2020-04-15
Add tips for checking dependencies

ulm:


ulm:
@zlogene: This originates from [bug 465764](https://bugs.gentoo.org/465764) which relates to the ebuild quiz as well, therefore CCing you as reviewer.

ulm:
Ping. I am going to merge this by the end of this week if there are no comments by then.

general-concepts/dependencies/text.xml
```diff
@@ -592,9 +569,87 @@ DEPEND="
...
+
+<p>
+It is important to ensure that all the dependencies are complete for your
+package:
+</p>
+
+<dl>
+  <dt>Look at installed binaries/libraries</dt>
+  <dd>
+    Use a tool like <c>scanelf -n</c> to list <c>DT_NEEDED</c> entries
```
orlitzky, line 588: `objdump -p <file> | grep NEEDED` does the same thing and comes with binutils (scanelf needs app-misc/pax-utils).

aballier: pax-utils is a dep of portage so no real problem here

maybe this could be more explicit, I got this oneliner from Diego back in the days:
```
qlist $1 | scanelf -L -n -q -F '%n #F' | tr , ' ' | xargs qfile -C | sort -u
```

I called it `rundep` on my system and this gives me:
```
$ rundep util-linux
sys-apps/systemd: /lib64/libsystemd.so.0
sys-apps/systemd: /lib64/libudev.so.1
sys-apps/util-linux: /lib64/libblkid.so.1
sys-apps/util-linux: /lib64/libfdisk.so.1
sys-apps/util-linux: /lib64/libmount.so.1
sys-apps/util-linux: /lib64/libsmartcols.so.1
sys-apps/util-linux: /lib64/libuuid.so.1
sys-apps/util-linux: /usr/lib32/libblkid.so.1
sys-apps/util-linux: /usr/lib32/libuuid.so.1
sys-libs/glibc: /lib32/libc.so.6
sys-libs/glibc: /lib64/ld-linux.so.2
sys-libs/glibc: /lib64/ld-linux-x86-64.so.2
sys-libs/glibc: /lib64/libcrypt.so.1
sys-libs/glibc: /lib64/libc.so.6
sys-libs/glibc: /lib64/librt.so.1
sys-libs/glibc: /lib64/libutil.so.1
sys-libs/ncurses: /lib64/libncursesw.so.6
sys-libs/ncurses: /lib64/libtinfow.so.6
sys-libs/pam: /lib64/libpam_misc.so.0
sys-libs/pam: /lib64/libpam.so.0
sys-libs/readline: /lib64/libreadline.so.8
sys-libs/zlib: /lib64/libz.so.1
```

ulm: There are several scripts around, some of the more advanced may be [checkdeps](https://github.com/betelgeuse/scripts/blob/master/checkdeps) and [checkdeps.rb](https://github.com/betelgeuse/scripts/blob/master/checkdeps.rb), which uses `scanelf -B -F '%n#F'` internally. Not sure if the devmanual is the right place for such elaborate examples. That one-liner above might cause more questions than it answers.

Maybe we could say "a tool like `scanelf -n` (from app-misc/pax-utils) or `objdump -p` (from sys-devel/binutils)"?

aballier: your call in the end -- I'm not sure either -- but to me the change is suggesting to do this manually file per file, which I believe will also raise questions and pain.

I wouldn't mention the first checkdeps though as it uses ldd

orlitzky: Doesn't matter to me either, I only pointed out `objdump` because pax-utils is pretty Gentoo-specific.

I did notice we're missing the best answer of all, though, for how to determine dependencies:

1. Download an openstack image from http://distfiles.gentoo.org/experimental/amd64/openstack/
2. Turn off as many pointless profile- and USE-default flags as possible in make.conf and package.use
3. `emerge -uDN @world`
4. emerge --depclean
5. try to install your package and see if it breaks

mgorny: Why openstack and not plain stage3?

There's also https://github.com/mgorny/puremerge (which still needs improvements) that tries to unmerge everything besides `@system` and direct build-time dependencies.

There's the option of using `-B` and `-k` to build without runtime dependencies and then try to install. It has the advantage of verifying that binary packages work correctly.

orlitzky: The openstack images are kvm/qemu virtual machines that you can download to get a "bare" system in a few minutes.

The best way to do VM testing is to keep a truly-bare base image, and to do all of your testing in a delta VM that can be deleted afterwards (leaving you back at the bare base image), but the instructions for that are a lot more complicated than the one-sentence "download this bare(ish) VM and try it" that works with the openstack images.

ulm: > Maybe we could say "a tool like `scanelf -n` (from app-misc/pax-utils) or `objdump -p` (from sys-devel/binutils)"?

For now, I have amended the commit with this.

As always, perfect is the enemy of good, but I don't want to delay this any further (unless something in the commit would be factually incorrect). Additional approaches to dependency checking can be added later.

PR #169, opened 2020-04-16
Two small updates for metadata.xml description

ulm:


ebuild-writing/misc-files/metadata/text.xml
```diff
@@ -68,6 +68,7 @@ metadata.xml:
     <c>&lt;longdescription&gt;</c>,
     <c>&lt;maintainer&gt;</c>,
     <c>&lt;slots&gt;</c>,
+    <c>&lt;stabilize-allarches&gt;</c>,
```
mgorny, line 71: I wonder if we should make it `<.../>` since it's always empty.

ebuild-writing/misc-files/metadata/text.xml
```diff
@@ -140,6 +141,14 @@ metadata.xml:
     tag may appear at most once.
   </ti>
 </tr>
+<tr>
+  <ti><c>&lt;stabilize-allarches&gt;</c></ti>
+  <ti>
+    Indicates an architecture-independent package. An ebuild can be marked
+    stable for all relevant architectures, when it has been tested on only one
+    architecture. This tag must have empty content.
```
mgorny, line 149: Maybe it's worth adding that this applies only if the same version has `~arch` keywords already. I think a lot of devs are missing that.

ebuild-writing/misc-files/metadata/text.xml
```diff
@@ -271,8 +280,8 @@ There are also some attributes that can be used with these tags:
 <tr>
   <ti>restrict</ti>
   <ti>
-    <c>&lt;maintainer&gt;</c>,
-    <c>&lt;longdescription&gt;</c>, <c>&lt;flag&gt;</c>
+    <c>&lt;longdescription&gt;</c>, <c>&lt;maintainer&gt;</c>,
```
mgorny, line 283: I don't get the order here ;-).

ebuild-writing/misc-files/metadata/text.xml
ulm: I'd rather not, since we're also omitting the closing tag for all other elements. (I guess it would be more systematic to list the element names without the brackets, but that doesn't make it any more readable either.)

mgorny: You could also use `<.../>` for everything.

ebuild-writing/misc-files/metadata/text.xml
ulm: The word "relevant" was supposed to convey that. :) Any suggestion for a more explicit wording? Keep in mind that the goal here is to describe the metadata, but not necessarily the exact workflow of arch testers.

mgorny: Nah, it's good enough as is.

ebuild-writing/misc-files/metadata/text.xml
ulm: It's supposed to be the same ordering as in https://www.gentoo.org/glep/glep-0068.html#top-level-structure.

ulm:
> Nitpick: you can test the return value of `xmllint --xpath` to avoid a string comparison (an error is returned if nothing is selected),

I know, but in order to distinguish it from other errors, one would either have to check stderr for the message "XPath set is empty", or check the exact exit code (which seems to be 10, but this isn't even documented). Neither of them seems to make things easier.

> and `/pkgmetadata/maintainer` should be a tiny bit more robust than `/*/maintainer` if the DTD ever changes.

Yeah, the `*` is mainly there because the line would exceed 80 chars otherwise. :smile: How about this instead?
```
[[ $(xmllint --xpath "count(/pkgmetadata/maintainer)" metadata.xml) -eq 0 ]]
```

orlitzky:
> I know, but in order to distinguish it from other errors, one would either have to check stderr for the message "XPath set is empty", or check the exact exit code (which seems to be 10, but this isn't even documented). Neither of them seems to make things easier.

Ok, makes sense.


> > and `/pkgmetadata/maintainer` should be a tiny bit more robust than `/*/maintainer` if the DTD ever changes.
> 
> Yeah, the `*` is mainly there because the line would exceed 80 chars otherwise. smile How about this instead?

Ha, well.. the original can be made compatible with POSIX sh by dropping the double brackets and the `== true` part (which is what a test does anyway), so they both have merit. It's up to you.

ebuild-writing/misc-files/metadata/text.xml
ulm: I never understood why one would write `<.../>`, which is a shortcut for `<...></...>` and denotes an empty element, even if that element (e.g., `maintainer`) cannot be empty. Then again, I'm not an XML person.

If we go for strictness then we should say "element" in the table header and omit the brackets. But I believe readability is better if we leave them in place, even if that doesn't strictly distinguish the element from its opening tag.

orlitzky:
LGTM; for Christmas maybe QA can turn off the CI check and mass sed these comments out so that no one ever tries to grep XML again.

ulm:
> LGTM; for Christmas maybe QA can turn off the CI check and mass sed these comments out so that no one ever tries to grep XML again.

I guess the main problem is that xmllint is awkward to use. With grep, one can do quick things like `grep -r -l maintainer-needed`, while xmllint requires a for loop on `find` output for the same operation. Also XPath doesn't have a function that would return the current filename, so one xmllint process is needed for every file. (Even if it had such a function, there would be no obvious way to call it because XPath designers didn't care to include an `if` condition in the language!)

orlitzky:
I use `xmllint --xpath` to find all packages where I'm listed as an explicit maintainer (i.e. not as a member of a project), and the `find` command turns out to be a lot slower than running the `xmllint` processes once you've found them. Nevertheless, we've all got python installed, and this will do it in one process if you `find ... | xargs` to it:

```python
#!/usr/bin/python3

import sys
import xml.etree.ElementTree as ET

for f in sys.argv[1:]:
    if ET.parse(f).getroot().find("maintainer") is None: print(f)
```

You can even parse the package name out without having to fire off dirname/basename or `cut` processes.

PR #173, opened 2020-06-15
Mention other DEPEND variables

DarthGandalf:
To avoid situation like https://github.com/gentoo/gentoo/pull/14804

@juippis @Alessandro-Barbieri

### DIFF
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -318,6 +318,14 @@ The following variables may or must be defined by every ebuild.
     </ti>
   </tr>
   <tr>
+    <ti><c>xDEPEND</c></ti>
+    <ti>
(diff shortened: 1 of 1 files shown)


ulm:
There is no such concept of "variables reserved for future" in PMS. Also we don't know if a future EAPI will have a variable named `CDEPEND` or one named `COMMON_DEPEND`.

So, ebuilds are free to use any variable name that doesn't have a special meaning in their EAPI (and that isn't listed [section 12.3.17](https://projects.gentoo.org/pms/7/pms.html#x1-13700012.3.17)). That's exactly why the concept of EAPI exists.

juippis:
gentoo/gentoo#14804 has nothing to do with PMS or devmanual. For what I care, this issue can be closed. 

@DarthGandalf other requests were ignored too in that PR.

PR #174, opened 2020-06-21
general-concepts/filesystem: Correct description for /opt

mgorny:

### DIFF
--- a/general-concepts/filesystem/text.xml
+++ b/general-concepts/filesystem/text.xml
@@ -12,7 +12,7 @@ The basic filesystem layout and purpose is as follows:
   <li><c>/bin</c>: Boot-critical applications</li>
   <li><c>/etc</c>: System administrator controlled configuration files</li>
   <li><c>/lib</c>: Boot-critical libraries</li>
-  <li><c>/opt</c>: Binary-only applications.</li>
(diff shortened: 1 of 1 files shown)


ulm:
Are there actual examples of open-source software that installs into /opt? Won't we consider that a QA violation?

mgorny:
What about open-source versions of software that is otherwise proprietary? Looking at my system, OpenJDK fits that definition.

general-concepts/filesystem/text.xml
```diff
@@ -45,8 +45,9 @@ Any binary which links against a library under <c>/usr</c> must itself go into
 </p>
 
 <p>
-The <c>/opt</c> top-level should only be used for binary-only applications.
-Binary-only applications must not be installed outside of <c>/opt</c>.
+The <c>/opt</c> top-level should only be used for applications that
+do not conform to the standard filesystem layout. This paricularly includes
```
ulm, line 49:
```suggestion
do not conform to the standard filesystem layout. This particularly includes
```

mgorny: Fixed, thanks.

ulm:
Merged, thanks.

PR #177, opened 2020-08-12
ebuild-maintenance/removal: Process for virtual removal

mgorny:

### DIFF
--- a/ebuild-maintenance/removal/text.xml
+++ b/ebuild-maintenance/removal/text.xml
@@ -85,5 +85,48 @@ Date:   Tue Oct 3 21:43:03 2017 +1100
 
 </body>
 </section>
+
+<section>
+<title>Removing a virtual package</title>
+<body>
+
+<p>
+Virtual packages are generally removed when they have no more than one
+provider left. The removal is preceded by updating the remaining ebuilds
+not to use the virtual. Since virtuals do not install any files, there
+is little value in proactively forcing them to be uninstalled from user
+systems or unnecessarily informing the user about the fact. Therefore,
+an alternative removal process is recommended.
+</p>
+
+<p>
+In order to remove a virtual package, follow the following procedure:
+</p>
+
+<ol>
+  <li>
+    If the virtual is being removed along with any of its providers,
+    include the virtual in the last-rites mail. However, please
+    do not include it in the <c>package.mask</c> entry as users do not need
+    to be forced to proactively unmerge it. Instead, add it
+    to <c>package.deprecated</c> to warn developers not to depend on it.
+    Wait the time appropriate for the last rites.
+  </li>
+  <li>
+    Update all ebuilds not to reference the virtual. Since there is
+    no urgent need to remove the virtual from user systems
+    and the resulting rebuilds would be unnecessary, do not bump ebuilds
+    when replacing the dependency.
+  </li>
+  <li>
+    Remove the package directly
+  </li>
+  <li>
+    Perform the post-removal cleanup, as with regular packages
+  </li>
+</ol>
+
+</body>
+</section>
 </chapter>
 </guide>


ebuild-maintenance/removal/text.xml
```diff
@@ -83,6 +83,48 @@ Date:   Tue Oct 3 21:43:03 2017 +1100
...
+<section>
+<title>Removing a virtual package</title>
+<body>
+
+<p>
+Virtual packages are generally removed when they have no more than one
+provider left. The removal is preceded by updating the remaining ebuilds
+not to use the virtual. Since virtuals do not install any files, there
+is little value in proactively forcing them to be uninstalled from user
+systems and therefore an alternative removal process is recommended.
```
ulm, line 98: I wonder if this is applicable to all removals of virtuals? I.e. should the policy say "recommended", or merely point out that the alternative procedure exists?

ebuild-maintenance/removal/text.xml
```diff
@@ -83,6 +83,48 @@ Date:   Tue Oct 3 21:43:03 2017 +1100
...
+  </li>
+
+  <li>
+    Remove the package directly.
+  </li>
+
+  <li>
+    Perform the post-removal cleanup, the same as with regular packages.
+  </li>
+</ol>
```
ulm, line 120: By the style guide:
- No full stops if the item consists only of a single sentence
- No blank lines between items

mgorny:
> Shouldn't last rites be sent still, plus an appropriate (possibly shortened) waiting time?

I don't think that's really desirable. After all, last rites are mostly user-oriented, so I don't think we should be bothering users with 'hey, we are removing this empty package, nothing changes for you, it will be cleaned up eventually, so just ignore this message'.

> Maybe mention `package.deprecated`?

I was wondering about that. However, I suppose p.d makes sense if we're not removing the package immediately. However, is there a reason not to remove it immediately?

ulm:
> > Shouldn't last rites be sent still, plus an appropriate (possibly shortened) waiting time?
> 
> I don't think that's really desirable. After all, last rites are mostly user-oriented, so I don't think we should be bothering users with 'hey, we are removing this empty package, nothing changes for you, it will be cleaned up eventually, so just ignore this message'.

Right, but most of the time removal of a virtual will not be an isolated event, but go together with removal of one of its providers. I don't have a strong opinion about this, but maybe it is more natural to list the virtual along with the other package in the last-rites?

mgorny:
Actually, I thought about virtuals being removed much later, when someone discovers that all other providers are gone ;-). But sure, if it's a combined event, I suppose it's good to list it in last rites message (but not p.mask).

I suppose the simple procedure needs to be replaced with a more detailed explanation of different cases.

mgorny:
v2 pushed. Added an explicit paragraph about last rites case, and reworded the rest a little.

ebuild-maintenance/removal/text.xml
```diff
@@ -83,6 +83,49 @@ Date:   Tue Oct 3 21:43:03 2017 +1100
...
+</p>
+
+<p>
+In order to remove a virtual package, follow the following procedure:
+</p>
+
+<ol>
+  <li>
+    If the virtual is being removed along with its second to last
+    provider, include the virtual in last rites mail. However, please
```
ulm, line 109:
```suggestion
    provider, include the virtual in the last-rites mail. However, please
```

ebuild-maintenance/removal/text.xml
```diff
@@ -83,6 +83,49 @@ Date:   Tue Oct 3 21:43:03 2017 +1100
...
+    do not include it in the package.mask entry as users do not need
```
ulm, line 110:
```suggestion
    do not include it in the <c>package.mask</c> entry as users do not need
```

ebuild-maintenance/removal/text.xml
```diff
@@ -83,6 +83,49 @@ Date:   Tue Oct 3 21:43:03 2017 +1100
...
+    to be forced to proactively unmerge it. Instead, add it
+    to package.deprecated to warn developers not to depend on it.
```
ulm, line 112:
```suggestion
    to <c>package.deprecated</c> to warn developers not to depend on it.
```

orlitzky:
Retroactively blessing a mistake by the QA team because you're butthurt someone mentioned it on the mailing list?

ulm:
> Retroactively blessing a mistake by the QA team because you're butthurt someone mentioned it on the mailing list?

Huh, what are you referring to? Please provide a pointer.

orlitzky:
https://archives.gentoo.org/gentoo-dev/message/dcebabbd6f13aed6622424d439f7becc
https://archives.gentoo.org/gentoo-dev/message/cc9c1668f478c2e0b10a552b6e555442
https://archives.gentoo.org/gentoo-dev/message/09a1ec118e8e1dcc3d8cc8cf4267e3e7

mgorny:
> Retroactively blessing a mistake by the QA team because you're butthurt someone mentioned it on the mailing list?

No. Formally wording a good idea that was established after the high-profile virtual removal mess (was it virtual/shadow)? I'd really appreciate if developers were able to be professional and move on with their life instead of repeatedly trolling that they can't live with their beloved `--dynamic-deps`.

orlitzky:
The mess was due to the virtual being removed without new revisions for the consumers. I and several other users complained about it. The PMS doesn't support your theory of how the zero-byte package will be harmlessly left behind, and the experience of myself and several users (even with portage) contradicts it as well. Deleting packages that are still in use causes dependency resolution failures, and that's what you're trying to encourage here. I agree it would be nice if we could get away with it, but ignoring the problem doesn't seem to work.

ulm:
> [...] After all, last rites are mostly user-oriented, so I don't think we should be bothering users with 'hey, we are removing this empty package, nothing changes for you, it will be cleaned up eventually, so just ignore this message'.

Thinking about it, I don't believe this is true. Last rites are sent to -dev and -dev-announce, so they're intended mainly for developers. What users see is the `package.mask` comment (and in fact, the actual patch gets this right).

ulm:
> https://archives.gentoo.org/gentoo-dev/message/dcebabbd6f13aed6622424d439f7becc

OK, but that one wasn't specifically about removal of a virtual.

I still don't understand why you would see that behaviour. Had `x11-apps/mkfontdir` been installed on your system at that point?

orlitzky:
> 
> OK, but that one wasn't specifically about removal of a virtual.
> 
> I still don't understand why you would see that behaviour. Had `x11-apps/mkfontdir` been installed on your system at that point?

Probably at some point, but not when I reported the problem? Or maybe it was trying to rebuild but couldn't because the ebuild was gone? Once I figured out what the problem was, I manually re-emerged all consumers (on all of my workstations!) to fix the problem.

And while that thread wasn't about a virtual, but I posted almost the same thing about the virtual in question:

https://archives.gentoo.org/gentoo-dev/message/64c42804eb4cf4bc7d1161a2e9222c6a

Whether or not it's a virtual is immaterial: having these things masked/removed without revisioning the consumers can break dependency resolution somehow. Even if you manage to fix that in some version of portage, the PMS certainly doesn't say that everything should be OK, so you can't rely on everything just working when you delete a package that might be pulled into the dependency tree.

PR #178, opened 2020-09-12
eclass-writing: Update list of eclass examples.

ulm:


eclass-writing/text.xml
```diff
@@ -27,15 +27,16 @@ Roughly speaking, there are three kinds of eclass:
 <ul>
   <li>
     Those which provide common functions which are used by many ebuilds (for
```
mgorny, line 29: Shouldn't that be 'these that'? Maybe sort them too, the order looks arbitrary.

ulm: I haven't changed the wording. It is the text from 2005, which :smile: was written by a native speaker of English.

They are currently sorted by popularity. I can make it alphabetical if you prefer.

SoapGentoo: Nope, this sounds correct

SoapGentoo (approved):

mgorny (approved):
That said, I think I had a different (IMO better) classification somewhere, one based on fitting exported phase functions.

ulm:
> That said, I think I had a different (IMO better) classification somewhere, one based on fitting exported phase functions.

:shrug: The perfect is the enemy of the good.

This came about when I grepped the devmanual for `eutils`, and while at it, I've replaced the other examples as well. Rewriting the whole section isn't on my to-do list for now.

PR #180, opened 2020-10-26
function-reference/query-functions: Correct description of "has".

ulm:


mgorny:
Maybe needle and haystack?

ulm:
> Maybe needle and haystack?

I considered that, but it's multiple separate arguments for the items to be searched.

Also, "haystack" may be misleading when it contains only a handful of grassblades in the typical case. :smile:

PR #181, opened 2020-11-26
Change example with https://bugs.gentoo.org/685160#c9 in mind

DarthGandalf:
Signed-off-by: Alexey Sokolov <sokolov@google.com>

### DIFF
--- a/ebuild-writing/functions/src_compile/building/text.xml
+++ b/ebuild-writing/functions/src_compile/building/text.xml
@@ -47,28 +47,39 @@ It is <e>not</e> correct to use the <c>${CC}</c> variable for this purpose.
 Sometimes a package will not use the user's <c>${CFLAGS}</c> or <c>${LDFLAGS}</c>.
 This must be worked around, as sometimes these variables are used for specifying
 critical ABI options. In these cases, the build scripts should be modified (for
-example, with <c>sed</c>) to use <c>${CFLAGS}</c> or <c>${LDFLAGS}</c> correctly.
+example, with <c>sed</c> or via a patch) to use <c>${CFLAGS}</c> or
+<c>${LDFLAGS}</c> correctly.
 </p>
+
+<note>
+Try to not substitute the value with <c>sed</c> directly, but instead make the
+build script use the variables. The variables may contain characters such as a
+slash, a comma, or a colon, which are often used with <c>sed</c>, and this would
+break its syntax.
+</note>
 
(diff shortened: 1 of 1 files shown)


ulm:
This would make the example less general, because `$(foo)` variable references work only with make but not with shell scripts. So I think we should leave `build.sh` in place there and recommend another character as sed separator, like `%` or `|` (or both).

When going for the solution of https://bugs.gentoo.org/685160#c9, one would typically *not* use a sed tweak, but apply a proper patch to the Makefile (with the additional advantage that it can be submitted upstream).

orlitzky:
If you prefer `build.sh` to `Makefile` then all you have to do is change the patch to use curly braces. The fix should be submitted upstream in either case.

orlitzky:
And while we're here... `$(tc-getCC)` should be inserted as `${CC}` or `$(CC)`.

ulm:
> If you prefer `build.sh` to `Makefile` then all you have to do is change the patch to use curly braces. The fix should be submitted upstream in either case.

There are two possibilities:

1. sed's replacement is a constant string. Then you know what characters it contains and most likely the colon (or the slash) will be fine as separator. However, I think that best practice in this case is *not* to use sed, but patch instead. Patches will explicitly fail when they no longer apply (e.g., after a version bump) while sed will just do nothing and return with a good status nevertheless.
2. It is not a constant string, which is what the example is about. In this case, the colon may fail (due to gcc's idiocy in their option syntax).

So, unless you want to forbid using sed altogether, we should simply update the example and recommend another character.

orlitzky:
In that case I would support deleting the example and amending the previous paragraph to say "In these cases, the build scripts should be patched to use `${CFLAGS}` or `${LDFLAGS}` correctly."

As it stands, the previous paragraph suggests modifying the builds scripts to use `${CFLAGS}` or `${LDFLAGS}`, and then the sed example demonstrates something subtly different that is causing all these problems.

DarthGandalf:
PTAL

ulm:
I am still not happy. Now the change is down to modifying Makefile/build.sh with a constant replacement, which arguably should be done in src_prepare, not src_compile. Also, the resulting Makefile isn't anything that could be submitted upstream (e.g., it looks for MYCFLAGS instead of CFLAGS).

DarthGandalf:
Something like that?

ulm:
Merged with some minor changes.
See commit message of 9612233 for details.

PR #182, opened 2020-12-21
ebuild-writing/misc-files/metadata: Update description tag

jonasstein:
Description tag should be used to describe important information about
this maintainer. Not to explain that this maintainer does not want to
maintain the package, or allows others to help.

### DIFF
--- a/ebuild-writing/misc-files/metadata/text.xml
+++ b/ebuild-writing/misc-files/metadata/text.xml
@@ -223,7 +223,7 @@ metadata.xml:
 <tr>
   <ti><c>&lt;bugs-to&gt;</c></ti>
   <ti>
-    Should contain a place where bugs can be filed, a URL or an e-mail address prefixed
+    Should contain a place where bugs can be filed, an URL or an e-mail address prefixed
     with <c>mailto:</c>. May appear at most once.
   </ti>
 </tr>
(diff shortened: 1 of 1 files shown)


ebuild-writing/misc-files/metadata/text.xml
```diff
@@ -223,7 +223,7 @@ metadata.xml:
 <tr>
   <ti><c>&lt;bugs-to&gt;</c></ti>
   <ti>
-    Should contain a place where bugs can be filed, a URL or an e-mail address prefixed
+    Should contain a place where bugs can be filed, an URL or an e-mail address prefixed
```
ulm, line 226: Nope, this was correct. "A" vs "an" follows pronunciation, not spelling. For example, "a unicorn" but "an x-ray".

ulm:
Thinking about it some more, I am not happy with this change. What is allowed in the maintainer's description is specified in [GLEP 67](https://www.gentoo.org/glep/glep-0067.html#package-maintenance), and the devmanual must not restrict it beyond that.

For example, there are some packages that install an Emacs mode, and the GNU Emacs team can be listed as a secondary maintainer specifically (and only) for these parts of the package. The second sentence of your change seems to discourage this kind of maintenance description, which would effectively be a change of the (GLEP 67) policy.

ulm:
Closing for now, feel free to submit an updated patch.

PR #183, opened 2020-12-29
Improve the usage of the term "ebuild phase functions"

gktrk:
When the text specifically refers to phase functions, be more verbose about it. Ebuilds can define other functions that are internal. Eclasses also define functions. Using the term "phase functions" should help clarify the ambiguity.


ulm (approved):
LGTM

PR #185, opened 2021-02-01
ebuild-writing/common-mistakes: Drop note on Header/Id line

ulm:
This is no longer a common mistake. Exactly the same note is still present in ebuild-writing/file-format.


gktrk (approved):
I think 4 years is enough time. Let's kill it.

PR #92, opened 2018-11-28
Licenses major revamp

mgorny:
Given that a lot of developers fail hard at licenses, and that I spend significant effort teaching proxied maintainers the basics, I think it wouldn't be unreasonable to make the devmanual more complete on the topic.

CC @ulm for early comments since I don't want to have to reword first commits late on ;-).

PR #186, opened 2021-02-06
general-concepts/licenses: LICENSE should include build scripts

ulm:
By the FSF's definition (see for example GPL-3):
"The 'Corresponding Source' for a work in object code form means all the source code needed to generate, install, and (for an executable work) run the object code and to modify the work, including scripts to control those activities."

It makes much sense to apply this as a general definition. Applying it only to GPL licensed packages would mean:
- We would need another case distinction, making the rules for LICENSE even more complicated.
- As an example, consider a MIT licensed package with a CDDL licensed (i.e., GPL incompatible) build script: If that package was a library, then it couldn't be linked against a GPL licensed package, whereas a LICENSE variable listing only MIT would falsely indicate that it could.

@gentoo/licenses

### DIFF
--- a/general-concepts/licenses/text.xml
+++ b/general-concepts/licenses/text.xml
@@ -11,10 +11,16 @@ match files existing in the repository's <c>licenses/</c> directory.
 </p>
 
 <p>
-The value of this variable should include all licenses pertaining
-to the files installed by the package. If some parts of the package
-are installed only conditionally, or their license depends on the USE
-flag combination, you can use USE conditionals in <c>LICENSE</c>:
+The value of this variable should include all licenses pertaining to the
+"corresponding source" of the files installed by the package. This includes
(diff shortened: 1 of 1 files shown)


ulm:
So yeah, the present text originated from PR #92, and it went through several rounds of review. Not sure why I had missed the issue back then; presumably I didn't think enough about build scripts and what "complete source" means.

thesamesam:
@ajakk well done for spotting it!

mgorny:
Does that mean that any autoconf-based package must use a GPL-compatible license?

ulm:
> Does that mean that any autoconf-based package must use a GPL-compatible license?

No. Autoconf itself is part of the system, and m4 scripts and the generated configure are shipped under an all-permissive license (which most of the time we don't list, because its terms are a subset of any other free software license).

PR #187, opened 2021-02-14
keywording: Clarify self-testing on amd64/x86

thesamesam:
Some confusion occurred about whether this
was just for keywording or not, so I've
changed the phrasing.

I've also removed the requirement for a bug;
this isn't required in practice and adds
unnecessary overhead.

### DIFF
--- a/keywording/text.xml
+++ b/keywording/text.xml
@@ -299,10 +299,9 @@ Vulnerability Treatment Policy</uri>
 <body>
 
 <p>
-AMD64, X86: If you are the maintainer of a package that currently has open bugs
-for amd64 or x86 stabilization and own the respectively amd64 or x86 hardware,
-you can do your own testing and keywording of your packages; as long as it is
-not a core system set dependency.
+AMD64, X86: If you are the maintainer of a package and own the respective amd64
+or x86 hardware, you can do your own testing (stabilization and keywording) of
+your packages; as long as it is not a core system set dependency.
 </p>
 
 <p>
--- a/keywording/text.xml
+++ b/keywording/text.xml
@@ -301,7 +301,10 @@ Vulnerability Treatment Policy</uri>
 <p>
 AMD64, X86: If you are the maintainer of a package and own the respective amd64
 or x86 hardware, you can do your own testing (stabilization and keywording) of
-your packages; as long as it is not a core system set dependency.
+your packages; as long as it is not a core system set dependency. Note that
+it is acceptable to test x86 using a
+<uri link="https://wiki.gentoo.org/wiki/Project:AMD64/32-bit_Chroot_Guide">
+specialized environment on amd64</uri>.
 </p>
 
 <p>


keywording/text.xml
```diff
@@ -299,10 +299,9 @@ Vulnerability Treatment Policy</uri>
 <body>
 
 <p>
-AMD64, X86: If you are the maintainer of a package that currently has open bugs
-for amd64 or x86 stabilization and own the respectively amd64 or x86 hardware,
-you can do your own testing and keywording of your packages; as long as it is
-not a core system set dependency.
+AMD64, X86: If you are the maintainer of a package and own the (respective)
+amd64 or x86 hardware, you can do your own testing (stabilization and keywording)
+of your packages; as long as it is not a core system set dependency.
```
ulm, line 304: Two pairs of parentheses in one sentence isn't the best writing style.

Otherwise, looks good to me.

thesamesam: Thanks, updated!

keywording/text.xml
```diff
@@ -299,10 +299,9 @@ Vulnerability Treatment Policy</uri>
 <body>
 
 <p>
-AMD64, X86: If you are the maintainer of a package that currently has open bugs
-for amd64 or x86 stabilization and own the respectively amd64 or x86 hardware,
-you can do your own testing and keywording of your packages; as long as it is
-not a core system set dependency.
+AMD64, X86: If you are the maintainer of a package and own the respective amd64
+or x86 hardware, you can do your own testing (stabilization and keywording) of
```
juippis, line 303: The impression I get from x86 project page is that hardware is not a strict requirement, as long as you have a "pure" x86 "environment" (be it chroot, container, virtual machine...)
https://wiki.gentoo.org/wiki/Project:X86/Arch_Testers_FAQ#What_are_the_basic_system_requirements_if_any.3F
https://wiki.gentoo.org/wiki/Project:X86/Arch_Testers_FAQ#I_don.27t_run_stable_x86.2C_my_box_is_.7Ex86._How_can_I_setup_an_x86_chroot.3F
https://wiki.gentoo.org/wiki/Project:X86/Chroot_Guide

thesamesam: Good point! What do you think of bfb4b63bbd19f0e5e71b3b11e89064ab08fd6298?

juippis: Well as I tried to hint earlier, I'd call it "environment" because virtual machines etc are allowed... :)
(I've no clue about correct syntax for devmanual, but noticed that long urls get wrapped between
```
<uri
link="..."
```
which allows easier wrapping at 80 characters)

keywording/text.xml
```diff
@@ -299,10 +299,11 @@ Vulnerability Treatment Policy</uri>
...
+your packages; as long as it is not a core system set dependency. Note that
+it is acceptable to test x86 using a <uri link="https://wiki.gentoo.org/wiki/Project:AMD64/32-bit_Chroot_Guide">
+specialized chroot on amd64.</uri>
```
ulm, line 306:
```suggestion
it is acceptable to test x86 using a
<uri link="https://wiki.gentoo.org/wiki/Project:AMD64/32-bit_Chroot_Guide">
specialized chroot on amd64</uri>.
```

ulm (approved):

gktrk (approved):

PR #189, opened 2021-02-15
ebuild-writing/users-and-groups: add note about used_free_uidgids.sh

juippis:


ebuild-writing/users-and-groups/text.xml
```diff
@@ -17,8 +17,12 @@ First, check the <uri link="https://api.gentoo.org/uid-gid.txt">UID/GID
...
 101 and 499. If you are adding a user and a group with the same name,
 use the same number for their UID and GID, respectively. When in
-doubt,
-take the next free number from 499 downwards.
+doubt, take the next free number from 499 downwards. There is a <uri
+link="https://gitweb.gentoo.org/data/api.git/tree/bin/used_free_uidgids.sh">
+helper script</uri> available in the <uri
+link="https://gitweb.gentoo.org/data/api.git">data/api.git</uri> to aid finding
+the next free UID or GID. Run it by issuing <c>./bin/used_free_uidgids.sh</c>
+inside api.git repo.
```
ulm, line 25: Given that there's a link to data/api.git immediately above, I think the first `<uri>` tag isn't strictly necessary. The second one isn't very useful either, because (IIUC) the script won't work outside the repo and downloading it from here doesn't make much sense.

Also, I think that the wording should be either "in the data/api.git repository" or "in data/api.git".

juippis: Done, originally I thought about **bolding** or **otherwise styling** the data/api.git, but plaintext probably works fine too. As you said it's linked above already.

ebuild-writing/users-and-groups/text.xml
```diff
@@ -16,9 +16,10 @@ categories.
 First, check the <uri link="https://api.gentoo.org/uid-gid.txt">UID/GID
 assignment list</uri> and choose a free UID/GID in the range between
 101 and 499. If you are adding a user and a group with the same name,
-use the same number for their UID and GID, respectively. When in
-doubt,
-take the next free number from 499 downwards.
+use the same number for their UID and GID, respectively. When in doubt, take
+the next free number from 499 downwards. There is a helper script available in
+the data/api.git repository to aid finding the next free UID or GID. Run it by
+issuing <c>./bin/used_free_uidgids.sh</c> inside api.git repo.
```
ulm, line 22: Strictly speaking, you cannot run a command in a repository. :smirk:
```suggestion
issuing <c>./bin/used_free_uidgids.sh</c> in the api.git checkout.
```

juippis: Done! This wording is a bit harder for me to understand at least, but I guess it's technically more correct.

ulm (approved):

ebuild-writing/users-and-groups/text.xml
```diff
@@ -16,9 +16,10 @@ categories.
 First, check the <uri link="https://api.gentoo.org/uid-gid.txt">UID/GID
 assignment list</uri> and choose a free UID/GID in the range between
 101 and 499. If you are adding a user and a group with the same name,
-use the same number for their UID and GID, respectively. When in
-doubt,
-take the next free number from 499 downwards.
+use the same number for their UID and GID, respectively. When in doubt, take
+the next free number from 499 downwards. There is a helper script available in
+the data/api.git repository to aid finding the next free UID or GID. Run it by
+issuing <c>./bin/used_free_uidgids.sh</c> in the api.git checkout.
```
gktrk, line 22: What about this simplification:

```suggestion
the next free number from 499 downwards. The helper script
<c>used_free_uidgids.sh</c> available in the bin/ directory of
the data/api.git repository can be used to find the next available
UID or GID.
```

We shouldn't need to tell **developers** how to run a script.

juippis: In **my** eyes your suggestion doesn't make it simpler, but more complicated to understand :P 

The "users-and-group" web page already has enough text for people to process, so to me it's easier with this tiny little concrete example. With your words, I have to pick up the pieces and put them in correct place. 

I may be thinking this too much from the POV of someone who contributes sporadically to Gentoo and has been tasked with GLEP-81 update. The devmanual is a source for people who are learning, more than for people "already there". 

But this is all semantics as the information is there regardless, and I might not be the best one to answer :P so anyone else?

ulm: How about making it a real example, e.g. like this:

```
<p>
... A helper script in the data/api.git
repository can be used to find the next available UID or GID:
</p>

<pre>
$ bin/used_free_uidgids.sh
#ID               UID       GID
...
458              FREE      USED
459              USED      USED
460              USED      FREE
461..500         USED      USED
Recommended GID only: 460
Recommended UID only: 458
Recommended UID+GID pair: 325
Free UIDs: 197
Free GIDs: 174
Free UID+GID pairs: 157
</pre>

juippis: At first thought I was against it too, but the output example could be used throughout the page. In other words, update the ebuild code examples below to match the suggested IDs. I don't view this as important as just acknowledging the script exists, but am willing to update it if the devmanual people agree?

ulm: Not sure if I understand. Do you want to replace "suricata" and "477" by something else?

juippis: Yes, with `325` or manipulate the output to something similar :P

ulm: I don't think that would be a good idea. Whatever the output of the script is, the next available ID will likely be taken in the future, and then we'll have an example where user/group name and ID don't match. We currently have a real ebuild as example, which is perfect.

gktrk: > **Edit:** Thinking about it, this is a lot of output and maybe not very useful as an example.

I agree with that. It also becomes yet another example output for us to make sure that it's synced with the output of the script when the script is changed.

Based on the feedback from @juippis, how about: `The helper script <c>./bin/used_free_uidgids.sh</c> available in the data/api.git repository can be used to find the next available UID or GID.` Simple and to the point. Finding available numbers shouldn't be that complicated for people if they are capable of understanding that their package needs to introduce user/group packages as deps.

juippis: I like it. Updated.

@ulm I meant by "manipulating the output" that for the sake of this example it could print `477` as the "next free one" matching **suricata**, but I agree with you both it doesn't add too much value.

PR #190, opened 2021-02-15
*/*: Use modern example of a prefix arch/keyword

thesamesam:
*/*: Use modern example of a prefix arch/keyword

    Gentoo/OpenBSD and Gentoo/FreeBSD are gone, but
    we still have Prefix which has some arches
    like ~x64-macos, so let's use that to keep
    the same level of variety.

### DIFF
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -503,7 +503,7 @@ You need not assign the IUSE variable in your ebuild if it is empty.
 </note>
 
 <p>
-Arch USE flags (<c>sparc</c>, <c>mips</c>, <c>x86-fbsd</c> and so on) should not be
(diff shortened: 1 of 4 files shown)


ebuild-writing/variables/text.xml
```diff
@@ -503,7 +503,7 @@ You need not assign the IUSE variable in your ebuild if it is empty.
 </note>
 
 <p>
-Arch USE flags (<c>sparc</c>, <c>mips</c>, <c>x86-fbsd</c> and so on) should not be
+Arch USE flags (<c>sparc</c>, <c>mips</c>, <c>x64-macos</c> and so on) should not be
```
ulm, line 506: Long line.
Otherwise, LGTM.

gktrk (approved):

PR #192, opened 2021-02-16
ebuild-writing/misc-files/patches: Add section about clean patches

ulm:
Taken from https://dev.gentoo.org/~vapier/clean-patches, with some rewording.
(Work in progress.)

Bug: https://bugs.gentoo.org/247266

### DIFF
--- a/ebuild-writing/misc-files/patches/text.xml
+++ b/ebuild-writing/misc-files/patches/text.xml
@@ -114,123 +114,189 @@ from the <c>vim</c> patch tarball:
 <title>Clean Patch Howto</title>
 <body>
 
+<p>
+"Clean patch" does not refer to the patch itself (as in the changes it makes to
+the source code). It refers to all the metadata that exists in the patch to
+make it "maintainable".
+</p>
+
+</body>
+
+<subsection>
+<title>Why</title>
+<body>
+
+<p>
+This may take more effort "up front", but the amount of effort that it saves
+for everyone else in the future more than makes up for it. This refers to other
+distributions or upstream maintainers who read the patch, or future Gentoo
+maintainers. By keeping all patches "clean", people can quickly and easily
+assess a patch without searching through many other files.
+</p>
+
+</body>
+</subsection>
+
+<subsection>
+<title>File Naming</title>
+<body>
+
+<p>
+Your patch name should be short and to the point. When doing a file listing
+(e.g., <c>ls files/</c>), it's a lot easier to be able to scan for relevant
+patches when they have good keywords in their file names.
+</p>
+
+<p>
+It should also include the package name and the version it was written against.
+This way, people searching for patches or who happen to just stumble across the
+file itself have a clue as to what it's for. Stripping out the <c>${PN}</c>
+(and to a lesser extent, the <c>${PV}</c>) makes the filename significantly
+less useful. The fact the files are typically stored in
+<c>${CATEGORY}/${PN}/files/</c> is irrelevant, because the patch may be used
+outside Gentoo.
+</p>
+
+</body>
+</subsection>
+
+<subsection>
+<title>How</title>
+<body>
+
+<p>
+Here's a check list of things to keep in the patch header:
+</p>
+
+<ul>
+  <li>
+    External references
+    <ul>
+      <li>Upstream mailing archives</li>
+      <li>Upstream bug reports</li>
+      <li>Upstream commit links</li>
+      <li>Upstream ChangeLog entries</li>
+      <li>Gentoo bug reports</li>
+    </ul>
+  </li>
+  <li>
+    Short/medium explanation
+    <ul>
+      <li>Why is the patch needed?</li>
+      <li>What is it fixing?</li>
+      <li>Why is it fixing it the way it is?</li>
+      <li>Proposal for better fixes in the future?</li>
+      <li>Is it a stop gap measure (workaround)?</li>
+      <li>How was it regression tested?</li>
+      <li>
+        Examples of before/after behaviour
+        <ul>
+          <li>How to reproduce bug without patch</li>
+          <li>How to show bug is fixed after patch</li>
+          <li>
+            Maybe upstream fixed it in a different way, so this test can be
+            used to show that the patch is no longer needed with newer versions
+          </li>
+        </ul>
+      </li>
+    </ul>
+  </li>
+  <li>
+    Status
+    <ul>
+      <li>Was it merged/rejected/postponed/etc. upstream?</li>
+      <li>Is it distribution-specific?</li>
+    </ul>
+  </li>
+  <li>
+    Attribution
+    <ul>
+      <li>Who found the bug?</li>
+      <li>Who fixed the bug?</li>
+      <li>Who wrote the patch?</li>
+      <li>Who tested the patch?</li>
+      <li>Who gave advice on the patch?</li>
+    </ul>
+  </li>
+</ul>
+
+<p>
+All this information should be <e>in the patch itself</e>. It should never be
+found in something like the ebuild. If you really want to put a comment next
+to a patch in an ebuild, then this is about the only thing that is OK
+(where 93671 is the Gentoo bug number):
+</p>
+
+<codesample lang="ebuild">
+PATCHES=(
+	"${FILESDIR}"/${P}-dont-umask.patch #93671
+)
+</codesample>
+
+<p>
+When documenting these things, it might be useful to use RFC822/Git-style tags
+to format the metadata. So when documenting the author, use:
+</p>
+
 <pre>
----------------------
-  CLEAN PATCH HOWTO
----------------------
+From: Nice Person &lt;foo@cow.example.com&gt;
+</pre>
 
-when i say "clean patch", i am not referring to the patch itself (as in the
-changes it makes to the source code).  i am referring to all the metadata that
-exists in the patch to make it "maintainable".
+<p>
+Or when documenting relevant URLs, use something like:
+</p>
 
--------
-  WHY
--------
+<pre>
+Bug: https://upstream.example.com/12345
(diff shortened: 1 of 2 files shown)


ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,277 @@ from the <c>vim</c> patch tarball:
...
+      <li>Who fixed the bug?</li>
+      <li>Who wrote the patch?</li>
+      <li>Who tested the patch?</li>
+      <li>Who gave advice on the patch?</li>
+    </ul>
+  </li>
+</ul>
+
+<p>
+All this information should be <e>in the patch itself</e>. It should never ever
```
thesamesam, line 224:
```suggestion
All this information should be <e>in the patch itself</e>. It should never
```

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,277 @@ from the <c>vim</c> patch tarball:
...
+be found in something like the ebuild. If you really really want to put a
```
thesamesam, line 225:
```suggestion
be found in something like the ebuild. If you really want to put a
```

or

```suggestion
be found in something like the ebuild. If you must put a
```

ebuild-writing/misc-files/patches/text.xml
thesamesam, line 224:
```suggestion
All this information should be (at least) <e>in the patch itself</e>. It should never ever
```

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,277 @@ from the <c>vim</c> patch tarball:
...
+<pre>
+From: Nice Person &lt;foo@cow.example.com&gt;
+</pre>
+
+<p>
+Or when documenting relevant URLs, use something like:
+</p>
+
+<pre>
+Bug: https://upstream.tracker.com/12345
```
thesamesam, line 248:
```suggestion
Bug: https://upstream.example.com/12345
```

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,277 @@ from the <c>vim</c> patch tarball:
...
+</p>
+
+<p>
+Also note that <c>patch</c> uses the timestamp info in order to remove empty
+files automatically. Alternatively, you can specify the <c>-E</c> option with
+<c>eapply</c> if you want to remove an empty file.
+</p>
+
+<p>
+The following script will help deleting these things:
```
thesamesam, line 292: Maybe note that people shouldn't do this in the ebuild, although it's obvious.

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,277 @@ from the <c>vim</c> patch tarball:
...
+
+<p>
+All this information should be <e>in the patch itself</e>. It should never ever
+be found in something like the ebuild. If you really really want to put a
+comment next to a patch in an ebuild, then this is about the only thing that is
+OK (where 12345 is the Gentoo bug number):
+</p>
+
+<codesample lang="ebuild">
+eapply "${FILESDIR}"/${P}-fatty-cow.patch #12345
```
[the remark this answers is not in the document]
SoapGentoo: agreed

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,277 @@ from the <c>vim</c> patch tarball:
...
+                  ^^^^^^^^^^^^ keep this part
+</pre>
+
+<p>
+Extra points if you make the filename in the <c>---</c>/<c>+++</c> section
+consistent and sane. That is, remove different leading <c>backup/paths/</c>
+and <c>.orig</c>/<c>.new</c> suffixes. Your patch should be in the <c>-p1</c>
+format because this tends to be much more standard than any other <c>-p#</c>.
+It is also what <c>eapply</c> understands by default. A good suggestion is to
+use the package name/version as the leading portion that gets stripped.
```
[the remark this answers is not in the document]
SoapGentoo: `a/` and `b/` are more forward-agnostic, so I would go with this too.

SoapGentoo:
I agree with all of @thesamesam's suggestions, let's get those in.

ebuild-writing/misc-files/patches/text.xml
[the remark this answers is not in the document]
ulm: I've mentioned both alternatives.

ebuild-writing/misc-files/patches/text.xml
ulm: I haven't included this one.

ebuild-writing/misc-files/patches/text.xml
ulm: Added a `#!/bin/bash` shebang which should make it clear.

ebuild-writing/misc-files/patches/text.xml
thesamesam: No problem.

ebuild-writing/misc-files/patches/text.xml
thesamesam: Sure.

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,281 @@ from the <c>vim</c> patch tarball:
...
+<p>
+All this information should be <e>in the patch itself</e>. It should never be
+found in something like the ebuild. If you really want to put a comment next
+to a patch in an ebuild, then this is about the only thing that is OK
+(where 12345 is the Gentoo bug number):
+</p>
+
+<codesample lang="ebuild">
+PATCHES=(
+	"${FILESDIR}"/${P}-fatty-cow.patch #12345
```
thesamesam, line 232: Let's do one more: different patch name, given we changed the others?

thesamesam (approved):

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,282 @@ from the <c>vim</c> patch tarball:
...
+<p>
+And if you want to note your copyright signoff, slap on a Signed-off-by tag:
+</p>
+
+<pre>
+Signed-off-by: Diligent Developer &lt;larry@gentoo.org&gt;
+</pre>
+
+<p>
+Finally, your patch should be clear of useless cruft, including dates. If it
```
gktrk, line 263: Is the `dates` different from `the timestamps on the files` 4 lines below?

gktrk (commented):
Other than the one comment I have, I'm good with all the changes.

ebuild-writing/misc-files/patches/text.xml
ulm: The "dates" were added in a previous round of review. Changed to "timestamps" now.

gktrk: My point rather was that if we mention it 4 lines below in a more meaningful context, is there a value to mentioning it here? `useless cruft` is ambiguous anyway, which is why it needs the sentences that follow it for clarification. Simply adding `, including timestamps`, only to mention it again in more detail later, doesn't do a good job helping clarify the ambiguity IMO.

ulm: Right, I misunderstood that. Removed now.

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,282 @@ from the <c>vim</c> patch tarball:
...
+
+<ul>
+  <li>
+    External references
+    <ul>
+      <li>Upstream mailing archives</li>
+      <li>Upstream bug reports</li>
+      <li>Upstream commit links</li>
+      <li>upstream changelog entries</li>
+      <li>Gentoo bug reports</li>
```
juippis, line 179: One lowercase `u` there.

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,282 @@ from the <c>vim</c> patch tarball:
...
+
+<subsection>
+<title>Why</title>
+<body>
+
+<p>
+This may take more effort "up front", but the amount of effort that it saves
+for everyone else in the future more than makes up for it. This refers to other
+distributions or upstream maintainers who read the patch, or future Gentoo
+maintainers/developers. By keeping all patches "clean", people can quickly and
```
juippis, line 133: `Gentoo maintainers/developers` does this need to be separated? Aren't package maintainers called maintainers regardless of their status?

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -109,5 +109,282 @@ from the <c>vim</c> patch tarball:
...
+<pre>
+@@ -80,6 +82,7 @@ case $sys in
+                  ^^^^^^^^^^^^ keep this part
+</pre>
+
+<p>
+Extra points if you make the filename in the <c>---</c>/<c>+++</c> section
+consistent and sane. That is, remove different leading <c>backup/paths/</c>
+and <c>.orig</c>/<c>.new</c> suffixes. Your patch should be in the <c>-p1</c>
+format because this tends to be much more standard than any other <c>-p#</c>.
```
juippis, line 282: I'd say to be more confident in here and say it is more standard than the others, especially to unify major Linux distributions in patch sharing.

ebuild-writing/misc-files/patches/text.xml
ulm: Yep, "maintainers" is enough here.

ebuild-writing/misc-files/patches/text.xml
ulm: I think the wording is fine as-is. We mention the eapply default in the next sentence, so there should be enough rationale.

PR #193, opened 2021-02-16
ebuild-maintenance/general: New section on maintaining ebuilds

thesamesam:
This is a draft new section on "Maintaining your package".

It intends to state the obvious (but possibly unwritten) rules we have around maintaining packages in Gentoo, such as checking for open bugs regularly, and keeping up with stabilisation once your package has received stable keywords.


ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,89 @@
...
+You can run <c>pkgcheck scan</c> on your packages to detect most of these,
+but some only become visible when running the scans on the whole tree
+(e.g. broken dependency graph).
+
+Please note that repoman, while advisable to use, at present detects only
+a subset of the issues that pkgcheck does. Given pkgcheck powers Gentoo's CI,
+it is wise to make use of pkgcheck or both tools.
+
+You can view the current full CI report
+<uri link="http://qa-reports.gentoo.org/output/gentoo-ci/output.html">
```
ajakk, line 43: Maybe say something about the `;maintainer=` parameter?

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,95 @@
...
+here</uri>.
+
+The CI bot should automatically email you if you are responsible for new
+warnings, but you should still check regularly for existing warnings.
+The 'croaker' IRC bot will also mention critical warnings in the #gentoo-dev IRC
+channel to allow tree consistency to be repaired ASAP.
+
+Ask a fellow developer if you're unsure of the meaning or how to fix a warning.
+If in doubt about who to ask, head to the #gentoo-dev or #gentoo-qa IRC channels.
```
ajakk, line 58: This is the devmanual, but nondevelopers often use it too, so it might be worth mentioning channels that don't have +m?

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,95 @@
...
+
+<e>Rationale: This prevents stable users from being stuck on old versions of the
+package with no justification.</e>
+</p>
+
+<p>
+Keep up with changes regarding 'declaration of support variables',
+e.g.<c>PYTHON_COMPAT</c> and <c>USE_RUBY</c>. Do <b>not</b> wait until a bug is
+filed for your package, be proactive and look out for prompts on the gentoo-dev
+mailing list.
```
ajakk, line 80: the mailing list and CI messages

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,95 @@
...
+channel to allow tree consistency to be repaired ASAP.
+
+Ask a fellow developer if you're unsure of the meaning or how to fix a warning.
+If in doubt about who to ask, head to the #gentoo-dev or #gentoo-qa IRC channels.
+</p>
+
+<p>
+If your package has stable keywords, you <b>must</b> request stabilisation of a
+newer version within a reasonable timeframe (usually at least 30 days, but
+see the Stabilization <uri link="::keywording/">section</uri>), provided there
```
ulm, line 64:
```suggestion
see the <uri link="::keywording/#Moving from ~arch to arch">stabilization section</uri>),
provided there
```

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,95 @@
...
+<p>
+Check regularly for new bugs on
+<uri link="https://bugs.gentoo.org/">bugs.gentoo.org</uri>.
+
+This includes CI/tinderbox style bugs, which are found by testing the latest
+ebuilds (by commit) and random packages, respectively.
+
+Note that <uri link="https://packages.gentoo.org/">packages.gentoo.org</uri>
+has an experimental Bugs listing per maintainer (see below) but it is
+not yet complete.
```
ulm, line 35: I think this sentence is more distracting than it is helpful. Bugzilla has a powerful search function with the possibility to save searches, so why would anyone prefer p.g.o (which IIUC is aimed primarily at users) to Bugzilla itself?

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,95 @@
...
+it is wise to make use of pkgcheck or both tools.
+
+You can view the current full CI report
+<uri link="http://qa-reports.gentoo.org/output/gentoo-ci/output.html">
+here</uri>.
+
+The CI bot should automatically email you if you are responsible for new
+warnings, but you should still check regularly for existing warnings.
+The 'croaker' IRC bot will also mention critical warnings in the #gentoo-dev IRC
+channel to allow tree consistency to be repaired ASAP.
```
ulm, line 55: Does mentioning of croaker help developers with ebuild maintenance? Using IRC is optional, and anyone present in #-dev will notice these messages.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,95 @@
+<?xml version="1.0"?>
+<guide self="ebuild-maintenance/general/">
+<chapter>
+<title>Maintaining your package</title>
```
ulm, line 4: Use _Title Case_ in headings.
```suggestion
<title>Maintaining Your Package</title>
```

ebuild-maintenance/general/text.xml
thesamesam: Agreed, dropping.

ebuild-maintenance/general/text.xml
thesamesam: Agreed, let's restore it if the functionality is complete in b.g.o. Note that the maintainer pages on p.g.o are designed for developers and feedback is welcome on how to make them better for that.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,95 @@
...
+Gentoo's Quality Assurance (QA) team also maintains a
+<uri link="https://projects.gentoo.org/qa/policy-guide/">policy guide</uri>,
+which contains useful advice on ebuild structure and behaviour.
+</note>
+  
+<p>
+You should always use the latest <uri link="::ebuild-writing/eapi/">EAPI</uri>.
+The main exception is when an eclass you consume does not yet support it.
+You should file a bug or otherwise contact the eclass maintainer to enquire
+about its status with respect to the new EAPI.
```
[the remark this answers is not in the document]
thesamesam: Very good point, thank you!

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+e.g.<c>PYTHON_COMPAT</c> and <c>USE_RUBY</c>. Do <b>not</b> wait until a bug is
+filed for your package, be proactive and look out for prompts on the gentoo-dev
+mailing list.
+
+You may need to wait until your package's dependencies support newer
+versions of e.g. Python first, so file a bug or speak to the package
+maintainer in such a case.
+
+<e>Rationale: It's important to add the latest
+e.g. Python versions which your package is compatible with as soon as possible
```
sbraz, line 103: I would move the "with":
>  e.g. Python versions with which your package is compatible as soon as possible

or drop it and write something like:
> e.g. Python versions which your package supports as soon as possible

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+to avoid conflicts for users when the default Python target in the Gentoo
+profiles change.</e>
```
sbraz, line 105: change**s**

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+You can run <c>pkgcheck scan</c> on your packages to detect most of these,
+but some only become visible when running the scans on the whole tree
+(e.g. broken dependency graph).
+
+Please note that repoman, while advisable to use, detects different
+(mostly a subset, with one or two exceptions) issues from what pkgcheck can.
+Given pkgcheck powers Gentoo's CI, it is wise to make use of pkgcheck.
+
+You can view the current full CI report
+<uri link="https://qa-reports.gentoo.org/output/gentoo-ci/output.html">
```
sbraz, line 55: Could we also mention the per-maintainer URLs (`;maintainer=<name>`)?

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+The CI bot should automatically email you if you are responsible for new
+warnings.
+
+Ask a fellow developer if you're unsure of the meaning or how to fix a warning.
+If in doubt about who to ask, head to the #gentoo-dev, #gentoo-qa, or
+#gentoo-dev-help (for non-developers too) IRC channels.
+</p>
+
+<p>
+Most Gentoo developers tend to agree that pkgcheck in its own serves the QA needs
```
gktrk, line 67: Citation needed :P I doubt anyway would refute the technical superiority of pkgcheck. Perhaps we should work around this controversy by simply stating that fact and telling that irrespective of the tool being used, it's the developer's responsibility to ensure quality control. This way we can keep this paragraph short and to the point.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+see the
+<uri link="::keywording/#Moving from ~arch to arch">stabilization section</uri>),
+provided there are no regressions compared to the last stable version.
+
+Conveniently, <uri link="https://packages.gentoo.org/">packages.gentoo.org</uri>
+has a per-maintainer report which lists packages which are now
+eligible for stabilisation based on time on the repository. Find this by looking
+at the "Maintainers" tab and clicking your name, then "Stabilization".
+
+<e>Rationale: This prevents stable users from being stuck on old versions of the
```
gktrk, line 88: I think this is redundant. No developer with common sense should be arguing that not stabilizing their packages is better.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+package with no justification.</e>
+</p>
+
+<p>
+Keep up with changes regarding 'declaration of support variables',
+e.g.<c>PYTHON_COMPAT</c> and <c>USE_RUBY</c>. Do <b>not</b> wait until a bug is
+filed for your package, be proactive and look out for prompts on the gentoo-dev
+mailing list.
+
+You may need to wait until your package's dependencies support newer
```
gktrk, line 98: If this is meant to be a separate paragraph, it should have its own `<p/>`. But I think we can meld this to the previous paragraph by simply saying `If you are blocked by a dependency not supporting compatible support variables, file a bug or contact the maintainer of the dependency to promptly resolve it.`

gktrk (changes requested):
Without blocking the inclusion of this PR, we should start thinking about how this fits into ebuild-maintenance. Before, we had a single page that's similar to this in concept and we ended up creating more subsections eventually. The title doesn't quite fit IMO but like I said, it shouldn't block this PR.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+
+<note>
+Gentoo's Quality Assurance (QA) team also maintains a
+<uri link="https://projects.gentoo.org/qa/policy-guide/">policy guide</uri>,
+which contains useful advice on ebuild structure and behaviour.
+</note>
+  
+<p>
+You should always use the latest <uri link="::ebuild-writing/eapi/">EAPI</uri>.
+
```
juippis, line 18: If we are going to make this canon, I'd like this to be more precise:
```suggestion
You should always use the latest <uri link="::ebuild-writing/eapi/">EAPI</uri>
when adding new ebuilds or eclasses to ::gentoo repository.
```
Because it's fine to not bump EAPI when you add a patch, or eg a subslot binder to a dep with new subslot. Of course this is also common sense.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+The main exception is for critical packages like Portage or
+those typically maintained by base-system which normally wait
+~2-3 years.
+
+Note that for new EAPIs, we require 1 year before using them in
+stable ebuilds.
```
juippis, line 24: What does this mean? If EAPI-8 was pushed today, no EAPI-8 ebuild was going to be stabilized for a year? That doesn't seem right.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+about its status with respect to the new EAPI.
+
+<e>Rationale: Newer EAPIs add improvements which are often user-visible,
+e.g. EAPI 6's eapply_user and EAPI 7's improved cross-compile support.</e>
+</p>
+
+<p>
+Check regularly for new bugs on
+<uri link="https://bugs.gentoo.org/">bugs.gentoo.org</uri>.
+</p>
```
juippis, line 37: "Respond and resolve bugs assigned to you. If you do not use package assigned to you anymore or can't maintain it, pass it on to someone else." (with a link to https://devmanual.gentoo.org/general-concepts/package-maintainers/index.html#adding-and-removing-maintainers)

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+
+<p>
+Address CI warnings for your packages. These are generated by pkgcheck.
+You can run <c>pkgcheck scan</c> on your packages to detect most of these,
+but some only become visible when running the scans on the whole tree
+(e.g. broken dependency graph).
+
+Please note that repoman, while advisable to use, detects different
+(mostly a subset, with one or two exceptions) issues from what pkgcheck can.
+Given pkgcheck powers Gentoo's CI, it is wise to make use of pkgcheck.
```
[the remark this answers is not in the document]
juippis: I feel like devmanual should have a section dedicated to pkgcheck, as it's hard to find documentation on how to use it properly.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,111 @@
...
+
+<p>
+Most Gentoo developers tend to agree that pkgcheck in its own serves the QA needs
+of Gentoo. Indeed, the Gentoo repository itself uses only pkgcheck for its CI.
+However, it is important to note that the older, venerable repoman is still used
+and respected by many Gentoo developers. It is an area for future research to
+determine which checks are included/missing from one or the other tool <d/> in
+the meantime, it is safest to use both (and honestly, it doesn't really take
+that long).
+</p>
```
juippis, line 74: This whole paragraph seems hypotethical and not fitting for devmanual? There's a lot of assumption and "thoughtflowing" involved. I'd rather have it replaced with concrete example how to run `repoman` and `pkcheck` parallel in everyday ebuild maintenance.

ulm:
> Without blocking the inclusion of this PR, we should start thinking about how this fits into ebuild-maintenance. Before, we had a single page that's similar to this in concept and we ended up creating more subsections eventually. The title doesn't quite fit IMO but like I said, it shouldn't block this PR.

Pasting conversation from #gentoo-dev for reference:

&lt;@ulm&gt; sam_: gokturk has a point that a section with title "general" may indicate a structural problem. especially because ebuild-maintenance was sort of a catch-all section and it took some effort to clean it up
&lt;@ulm&gt; in fact, it was added to the devmanual very late, when we went through the old devrel handbook and collected all the forgotten things
&lt;@ulm&gt; and now you've filed a PR that "intends to state the obvious (but possibly unwritten) rules we have around" which is similar to what we had done before :)
&lt;@ulm&gt; sam_: so, maybe it would be better to split it up and add it to more specific chapters
&lt;@ulm&gt; for example, EAPI stuff could go to ebuild-writing/eapi
&lt;@sam_&gt; I mostly agree but I also think it's useful to have a single list of responsibilities
&lt;@sam_&gt; we have a problem where people say "I didn't realise I needed to do X"
&lt;@sam_&gt; perhaps we can find homes for some of the entries but maybe not all
&lt;@sam_&gt; I'll need to find the old section to compare with the content then
&lt;@ulm&gt; the stabilisation things could obviously be added to "Keywording and Stabilization" which is a toplevel chapter
&lt;@ulm&gt; old section?
&lt;@sam_&gt; you mentioned there was a previous section of this type
&lt;@sam_&gt; i'd like to compare to see if the old content was more amenable to being split (as it evidently was) and compare
&lt;@ulm&gt; sam_: you'd have to go through commit history of ebuild-maintenance/text.xml
&lt;@ulm&gt; and ebuild-writing/ebuild-maintenance/text.xml before that
&lt;@ulm&gt; it's messy :)
&lt;@ulm&gt; so yeah, it's basically three things you're talking about, EAPI, stabilisation, and pkgcheck/ci
&lt;@ulm&gt; which aren't much related, except that they're all "dev responsibilities". but we could say that about the whole devmanual content :)

thesamesam:
Further context too for me to take into account on revisions:
```
[2021-02-17T15:42:45+0000] <@sam_> ulm: I'm not fully sure what to say
[2021-02-17T15:42:57+0000] <@sam_> ulm: We can add them to the individual pages, but I'm worried this will make it easier for them to get lost
[2021-02-17T15:43:10+0000] <@sam_> the idea was to list things which you need to actively do, rather than things to consider when e.g. writing a new ebuild
[2021-02-17T15:49:40+0000] <@antarus> I'd consider just making a section for responsibilities
[2021-02-17T15:49:50+0000] <@antarus> many of them are...principles? perhaps?
[2021-02-17T15:49:54+0000] <@antarus> more than specific guidelines
[2021-02-17T15:50:25+0000] <@antarus> I mean, maybe first, start by making a page with all of the responsibilities on it
[2021-02-17T15:50:32+0000] <@antarus> and then try to figure out where they all might go?
[2021-02-17T15:51:03+0000] <@antarus> e.g. is it a 'dev responsibility' not to cause excessive rebuilds?
[2021-02-17T15:51:14+0000] <@antarus> is it a 'dev responsibility' not to cause excessive metadata regen?
[2021-02-17T15:51:26+0000] <@antarus> is it a 'dev responsibility' not to cause nasty depgraphs for end users?
[2021-02-17T15:53:25+0000] <@sam_> i think that sounds like a good plan
[2021-02-17T15:53:30+0000] <@sam_> start with that, go from there
```

thesamesam:
Will return to this.

PR #195, opened 2021-02-23
Add permalinks to section headers

ulm:
See [bug 772005](https://bugs.gentoo.org/772005).


ulm:
@arzano: Could you take a look at the CSS part? Is this a reasonable way of doing things?

thesamesam (approved):
This is exactly what I had in mind & works well here. Thank you.

PR #197, opened 2021-03-05
general-concepts/dependencies: sort dependencies, add revision notice

thesamesam:
- Add note regarding sorting dependencies
- Sort dependencies in example
- Include a notice to consider the Ebuild Revisions section

### DIFF
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -431,7 +431,9 @@ This can be nested:
 
 <codesample lang="ebuild">
 DEPEND="!build? (
+	&gt;=sys-libs/ncurses-5.2-r2
 	gcj? (
+		&gt;=media-libs/libart_lgpl-2.1
 		gtk? (
 			x11-libs/libXt
 			x11-libs/libX11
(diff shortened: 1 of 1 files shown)


general-concepts/dependencies/text.xml
```diff
@@ -431,7 +431,9 @@ This can be nested:
 
 <codesample lang="ebuild">
 DEPEND="!build? (
+	&gt;=sys-libs/ncurses-5.2-r2
 	gcj? (
+		&gt;=media-libs/libart_lgpl-2.1
```
ulm, line 442: This example shows its age. It was added in 2008 via [bug 174555](https://bugs.gentoo.org/174555), and was taken straight from gcc-4.2.2.ebuild. Some of the packages listed (e.g. x11-proto/*) no longer exist.

Maybe replacing the example by something more modern would be better?

thesamesam: I'm in favour of modernising it, but let's do it as part of a separate PR. I've filed a bug so that I don't forget: https://bugs.gentoo.org/774744.

ulm: wfm

PR #199, opened 2021-03-11
*/*: make it clear that <stabilize-allarches> ends in slash, />

juippis:
Modified all appearances of <stabilize-allarches>:
 - keywording/text.xml
 - ebuild-writing/misc-files/metadata/text.xml

### DIFF
--- a/keywording/text.xml
+++ b/keywording/text.xml
@@ -348,7 +348,7 @@ arches that you would like to stabilize.
 
 <p>
 If your package is architecture-independent, you should add the
-<c>&lt;stabilize-allarches&gt;</c> tag to metadata.xml. This allows consistency
+<c>&lt;stabilize-allarches/&gt;</c> tag to metadata.xml. This allows consistency
 in future stabilizations and saves arch teams considerable work.
 </p>
 


sbraz:
Wouldn't `pkgcheck` and `repoman` complain about that?

juippis:
Yes, I thought it'd be nice to give a working example out-of-the-box for everyone. AFAIK this is the only tag that doesn't have content in metadata.xml.

keywording/text.xml
```diff
@@ -348,7 +348,7 @@ arches that you would like to stabilize.
 
 <p>
 If your package is architecture-independent, you should add the
-<c>&lt;stabilize-allarches&gt;</c> tag to metadata.xml. This allows consistency
+<c>&lt;stabilize-allarches&gt;/</c> tag to metadata.xml. This allows consistency
```
ulm, line 351: That's the only place where I'd add it.
However:
```suggestion
<c>&lt;stabilize-allarches/&gt;</c> tag to metadata.xml. This allows consistency
```

ebuild-writing/misc-files/metadata/text.xml
```diff
@@ -120,7 +120,7 @@ metadata.xml:
   </ti>
 </tr>
 <tr>
-  <ti><c>&lt;stabilize-allarches&gt;</c></ti>
+  <ti><c>&lt;stabilize-allarches&gt;/</c></ti>
```
ulm, line 123: We don't mention a closing tag for any of the other tags, so I'd rather not for reasons of consistency.

Besides, the description says "This tag must have empty content."

keywording/text.xml
juippis: Heh, I guess I was blinded by the `</c>`. Re-did this based on your suggestion.

ulm:
HTML afficionados would write `<stabilize-allarches />` (note the blank to appease the IE gods) or `<stabilize-allarches></stabilize-allarches>`.

Just kidding. :smile:

PR #200, opened 2021-03-12
devbook-guide: Clearer section structure

ulm:
Drop the "Devbook XML" section header which was not very useful.
Instead, promote "Basic structure" from subsection to section,
and introduce two new sections "Body elements" and "Inline elements".

So, the resulting section structure is:

- DevBook XML design goals
- Basic structure
  - Sections and subsections
  - Including sub-documents
  - An example &lt;body&gt;
- Body elements
  - Code samples and colour-coding
  - Figures
  - Tables
  - Lists
- Inline elements
  - &lt;c&gt;, &lt;b&gt;, and &lt;e&gt;
  - &lt;uri&gt;
  - Intra-document references
- Coding Style
  - Internal Coding Style
  - External Coding Style

### DIFF
--- a/appendices/devbook-guide/text.xml
+++ b/appendices/devbook-guide/text.xml
@@ -24,8 +24,6 @@ documents.
 </section>
 
 <section>
-<title>Devbook XML</title>
-<subsection>
 <title>Basic structure</title>
 <body>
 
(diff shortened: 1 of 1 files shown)


thesamesam (approved):
Helpful change, I think.

PR #201, opened 2021-03-12
*/*: drop eapi7-ver examples/references

thesamesam:


ulm:
OK, I've cherry-picked the two unrelated trivial fixes for now.
Can you please rebase onto master, so we can see what's going on?

thesamesam:
@ulm Done.

ebuild-writing/functions/src_unpack/rpm-sources/text.xml
```diff
@@ -55,9 +55,9 @@ patches. The filename should be <c>suse-fetchmail-6.2.5.54.1.ebuild</c>.
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI="6"
+EAPI=7
```
ulm, line 58: That whole example ebuild needs to be updated. It will no longer work since rpm.eclass was updated in 2009 (!).
So maybe we should just leave it for now?

In fact, the snippet above is just as old, e.g. calls epatch in src_unpack.

thesamesam: No idea how I missed that ;)

Let's take the first commit and forget about this for now and re-examine as part of our general examples audit.

PR #202, opened 2021-03-12
ebuild-writing/using-eclasses: mention sorting inherit arguments

thesamesam:

### DIFF
--- a/ebuild-writing/using-eclasses/text.xml
+++ b/ebuild-writing/using-eclasses/text.xml
@@ -26,6 +26,15 @@ link="::general-concepts/portage-cache"/>).
 </p>
 
 <p>
+When using <c>inherit</c>, it is best practice to sort the arguments (eclasses)
+alphabetically. An exception is where the phases exported by an eclass are
+affected by subsequent arguments. For example, <c>multilib-minimal.eclass</c>
+mentions in its
+<uri link="::eclass-reference/multilib-minimal.eclass/">documentation</uri>
+that it should be inherited last because it overrides most phases.
+</p>
+
+<p>
 After inheriting an eclass, its provided functions can be used as
 normal. Here's an example ebuild, <c>foomatic-0.1-r2.ebuild</c>, which
 uses three eclasses:


ebuild-writing/using-eclasses/text.xml
```diff
@@ -25,6 +25,14 @@ illegal (except where the inheritance criteria are cache-constant <d/> see <uri
 link="::general-concepts/portage-cache"/>).
 </p>
 
+<p>
+When using <c>inherit</c>, sort the arguments (eclasses) alphabetically, except
+for where the phases exported by an eclass are affected by subsequent arguments,
```
ulm, line 30: Could this be worded a little weaker, because I don't think is is imperative as a policy? For example "... it is good practice to sort the eclasses alphabetically ..."

ebuild-writing/using-eclasses/text.xml
```diff
@@ -25,6 +25,14 @@ illegal (except where the inheritance criteria are cache-constant <d/> see <uri
...
+e.g. <c>multilib-minimal.eclass</c> mentions in its
+<uri link="::eclass-reference/multilib-minimal.eclass/">documentation</uri>
+that it should be inherited last because it overrides most phases.
```
ulm, line 33: This sentence is too long. Full stop after "arguments", then start a new sentence "For example, ...".

thesamesam: Done.

ebuild-writing/using-eclasses/text.xml
thesamesam: I've modified it but requested we make it [policy](https://bugs.gentoo.org/775686).

ebuild-writing/using-eclasses/text.xml
```diff
@@ -25,6 +25,15 @@ illegal (except where the inheritance criteria are cache-constant <d/> see <uri
 link="::general-concepts/portage-cache"/>).
 </p>
 
+<p>
+When using <c>inherit</c>, it is best practice to sort the arguments (eclasses)
+alphabetically. An exception is where the phases exported by an eclass are
+affected by subsequent arguments. For example, <c>multilib-minimal.eclass</c>
+mentions in its
+<uri link="::eclass-reference/multilib-minimal.eclass/">documentation</uri>
+that it should be inherited last because it overrides most phases.
```
ulm, line 34: Thinking about it, an ebuild shouldn't rely on inherit order.

Instead, in the (relatively rare?) case that there's more than one eclass exporting a phase function, the ebuild should define that phase function and call the eclass functions in a well-defined order.

ebuild-writing/using-eclasses/text.xml
```diff
@@ -25,6 +25,15 @@ illegal (except where the inheritance criteria are cache-constant <d/> see <uri
...
+</p>
```
ulm, line 35: How about something similar to the following instead?

```suggestion
<p>
When using <c>inherit</c>, it is best practice to sort the arguments (eclasses)
alphabetically.
</p>

<important>
Don't rely on inherit order. It can be different from what it appears to be
because eclasses can indirectly inherit each other. If several eclasses export
the same phase functions, the ebuild should therefore define that function
and explicitly call the eclass functions in a well defined order.
</important>
```

ebuild-writing/using-eclasses/text.xml
SoapGentoo: are you going to rewrite all the `multilib-minimal` using ebuilds? I doubt it. Having the last eclass be the canonical one has been standard practice for a long time

ebuild-writing/using-eclasses/text.xml
SoapGentoo: Again, disagree, noone is going to clutter multilib ebuilds like this.

ulm: "This eclass must always go last" is strange advice. What if more than one eclass says that?

SoapGentoo: The eclass won't say it, it's the use that implies where it does (even though anything overriding `multilib-minimal` would seem strange)

ulm: But it does:
```
# multilib-minimal should _always_ go last in inherit order!
```

ulm:
For reference, the [old Devrel Handbook](https://github.com/gktrk/devrel-handbook/blob/master/hb-guide-eclass.xml) had some wisdom about this:

> Oh, and you can inherit several eclasses at the same time by saying:
> ```
> inherit eclass1 eclass2 [...]
> ```
> ...but watch their order! Remember, eclasses can inherit one another and override each other's settings, so you should be careful when dealing with multiple eclasses that might influence one another.
>
> [...]
>
> Let's look at another problem for a moment. Suppose that foo.eclass and bar.eclass both define src_compile(). If you inherit both foo and bar you'll get a different src_compile() depending on the order in which you inherit them. That's ok; you're supposed to keep track of your inheritance order. But you may want to call either of the two src_compile()s explicitly.

This is still up to date, so maybe some of it could be reused?

(I also think that the last sentence from above is what I remembered as a policy, but looks like it was never an absolute requirement.)

PR #203, opened 2021-03-12
*/*: miscellaneous fixes

thesamesam:
Please let me know if you'd prefer I split this up.

- bash: make clear that ```[[ ]]``` tests are preferred when we reference them earlier on (and why)
- diff-and-patch: fix man page notation
- error-handling: mention the need to || die with external commands
- tools-reference: refer to aforementioned die

### DIFF
--- a/ebuild-writing/error-handling/text.xml
+++ b/ebuild-writing/error-handling/text.xml
@@ -57,9 +57,10 @@ Sometimes displaying additional error information beforehand can be useful. Use
 <c>eerror</c> to do this. See <uri link="::ebuild-writing/messages"/>.
 </p>
 
-<p>
-It's best to use <c>|| die</c> too often than too little.
-</p>
+<note>
+You should use <c>die</c> on almost all external commands in ebuilds.
+</note>
+
 </body>
 </section>
 
--- a/tools-reference/bash/text.xml
+++ b/tools-reference/bash/text.xml
@@ -120,6 +120,14 @@ all new code.
 </important>
 
 <p>
+POSIX compliance is not a concern for ebuilds, as their interpreter is
+guaranteed to be GNU Bash. POSIX style tests have different semantics
+and using the common forms of tests adheres to the principle of least surprise.
+Most developers will be used to Bash test semantics and behaviour and
+deviating from this in ebuilds may be confusing.
+</p>
+
+<p>
 This is because <c>[[ ]]</c> is a bash syntax construct, whereas <c>[ ]</c> is a
 program which happens to be implemented as an internal <d/> as such, cleaner
 syntax is possible with the former. For a simple illustration, consider:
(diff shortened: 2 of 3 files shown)


ebuild-writing/error-handling/text.xml
```diff
@@ -35,6 +35,10 @@ Decent error handling is important because:
 <title>The <c>die</c> Function</title>
 <body>
 
+<note>
+You should use <c>die</c> on almost all external commands in ebuilds.
+</note>
+
```
ulm, line 41: Move the note to the end of the section.

Also, it says almost the same as the last paragraph of the section ("It's best to use || die too often than too little."), so maybe that paragraph could be dropped.

tools-reference/bash/text.xml
```diff
@@ -119,6 +119,14 @@ The <c>[[ ]]</c> form is generally safer than <c>[ ]</c> and should be used in
...
 </important>
 
+<p>
+Note that POSIX-compliance is not a concern for ebuilds themselves as bash
+is guaranteed to be present and ebuilds are written in bash. Therefore, authors
+should use the canonical standard tests to adhere to the principle of
+least surprise. This aids review and ensures that use is consistent throughout
+the repository.
+</p>
+
```
ulm, line 129: Sorry, but IMHO this paragraph has horrible style. Suggestion for first sentence: "POSIX compliance is not a concern for ebuilds, as their interpreter is guaranteed to be GNU Bash."

It's not entirely clear to me what you want to say with the second sentence?

tools-reference/text.xml
```diff
@@ -9,6 +9,12 @@ This section provides an overview of various useful standard Unix and Gentoo
 tools and utilities that may be used within ebuilds or when working with
 ebuilds.
 </p>
+
+<p>
+Please keep in mind the need for
+<uri link="::ebuild-writing/error-handling/#the-die-function">error handling</uri>
```
ulm, line 15: We use source references, not translated references in fragments (so if we change the translation in the XSL stylesheet, we don't have to update all links in the document). Therefore:
```suggestion
<uri link="::ebuild-writing/error-handling/#The die Function">error handling</uri>
```

ulm:
> Please let me know if you'd prefer I split this up.

Again, I've cherry-picked the trivial things (3 commits) so we get them out of the way

tools-reference/bash/text.xml
thesamesam: For reference, it sounds a bit better to say "How about ...?" or "This may be clearer: ..." rather than just "horrible", i.e. focus on the actual improvement. But I know it wasn't intended badly.

How about now?

ulm: Sorry if you perceived my wording as rude, that wasn't my intention. Cultural difference, I guess?

ulm (approved):

tools-reference/bash/text.xml
thesamesam: No worries at all, I know it wasn't! 🙂

thesamesam:
Let's get this one merged.

ulm:
> Let's get this one merged?

Yeah, it's been a week. I'm going to merge it later today.

PR #204, opened 2021-03-15
Unpack minor fixes

thesamesam:


ulm:
Should we mention that BDEPEND exists in EAPI 7 only?

thesamesam:
I thought about this and lent no because of the text it'd add (not worth it IMO).

Expanding:
Developers need to be aware of the different types of DEPENDency class anyway and by now, if you're not sure what unpacking is, hopefully you're on EAPI 7 and writing a new ebuild.

I take your point though -- maybe we should include this kind of information as a courtesy until things are in EAPI - 2. So, we wouldn't assume BDEPEND knowledge until EAPI 8 is in tree?

ulm:
No, I guess it's fine. Also, any remaining problem will go away with time.

PR #205, opened 2021-03-20
*/*: fix brace style, Bash tests, eapply

thesamesam:
Mostly brace style fixing but also one instance of POSIX -> Bash tests and epatch -> eapply.

### DIFF
--- a/ebuild-writing/common-mistakes/text.xml
+++ b/ebuild-writing/common-mistakes/text.xml
@@ -284,6 +284,11 @@ part.
 <title>DEPEND is incomplete</title>
 <body>
 
+<note>
+The tips in this section apply to all dependency classes, not just
+<c>DEPEND</c>.
+</note>
+
 <p>
 This is another very common error. The ebuild submitter submits an ebuild
 that "just works" without checking if the dependencies are correct. Here are
@@ -317,7 +322,7 @@ some tips on how to find the correct dependencies.
     programs, etc.</e>
     Usually the build process requires some dependencies such as intltool,
     libtool, pkg-config, doxygen, scrollkeeper, gtk-doc, etc. Make sure those
-    are clearly stated.
+    are clearly stated, usually in <c>BDEPEND</c>.
   </li>
 </ul>
 
--- a/ebuild-writing/functions/pkg_config/text.xml
+++ b/ebuild-writing/functions/pkg_config/text.xml
@@ -48,7 +48,7 @@ Taken from the <c>mysql</c> ebuilds. Note the use of <c>${ROOT}</c>.
 
 <codesample lang="ebuild">
 pkg_config() {
-	if [ ! -d "${ROOT}"/var/lib/mysql/mysql ] ; then
+	if [[ ! -d ${ROOT}/var/lib/mysql/mysql ]] ; then
 		einfo "Press ENTER to create the mysql database and set proper"
 		einfo "permissions on it, or Control-C to abort now..."
 		read
--- a/ebuild-writing/functions/src_compile/building/text.xml
+++ b/ebuild-writing/functions/src_compile/building/text.xml
@@ -59,7 +59,7 @@ src_prepare() {
 
 	# We have a weird Makefile to work with which ignores our
 	# compiler preferences. yay!
-	# Note the single quotes.
+	# Note the single quotes (hence the delimiter is not an issue)
 	sed -i -e 's:cc -O2:$(CC) $(CFLAGS) $(LDFLAGS):' Makefile \
 		|| die "sed fix failed. Uh-oh..."
 }
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -461,6 +461,12 @@ DEPEND="!build? (
 <title>Any of Many Dependencies</title>
 <body>
 
+<note>
+To ease dependency resolution for Portage, it is recommended that you sort
+the elements in <e>preferred</e> order first. See
+<uri link="https://bugs.gentoo.org/489458">this bug</uri> for more context.
+</note>
+
 <p>
 To depend on either <c>foo</c> or <c>bar</c>:
 </p>
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -626,12 +626,12 @@ package:
   </dd>
   <dt>
     Remember non-binary dependencies such as pkg-config, doc generation
-    programs, etc.
+    programs, etc. Such programs would usually belong in <c>BDEPEND</c>.
   </dt>
   <dd>
     Usually the build process requires some dependencies such as intltool,
     libtool, pkg-config, doxygen, scrollkeeper, gtk-doc, etc. Make sure those
-    are clearly stated.
+    are clearly stated. Again, such dependencies usually belong in <c>BDEPEND</c>.
   </dd>
 </dl>
 
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -347,7 +347,7 @@ the package name, where 'SLOT' is the <c>SLOT</c> of the package wanted:
 </p>
 
 <codesample lang="ebuild">
-DEPEND="qt3? ( x11-libs/qt:3 )
+DEPEND="qt5? ( dev-qt/qtcore:5 )
 	gtk? ( x11-libs/gtk+:2 )
 </codesample>
 
@@ -356,7 +356,7 @@ To depend on a specific version or version-range within a SLOT we use:
 </p>
 
 <codesample lang="ebuild">
-DEPEND="qt3? ( ~x11-libs/qt-3.3.8:3 )
+DEPEND="qt5? ( ~dev-qt/qtcore-5.15.2:5 )
 	gtk? ( &gt;=x11-libs/gtk+-2.24.9:2 )
 </codesample>
 </body>
--- a/tools-reference/sed/text.xml
+++ b/tools-reference/sed/text.xml
@@ -132,7 +132,12 @@ The above will replace <c>irksome texting</c> with
 <p>
 If the pattern or the replacement string contains the forward slash character,
 it is usually easiest to use a different delimiter. Most punctuation characters
-are allowed, although backslash and any form of brackets should be avoided.
+are allowed, although backslash and any form of brackets should be avoided. You
+should choose your delimiter <b>with care</b> to ensure it cannot appear in any
+strings involved in the subject/replacement. For example, using <c>sed</c> with
+CFLAGS is hazardous because it is user-supplied data (so may contain any
+character), but one should in particular avoid e.g.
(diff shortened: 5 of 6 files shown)


ebuild-writing/functions/pkg_config/text.xml
```diff
@@ -49,7 +48,7 @@ Taken from the <c>mysql</c> ebuilds. Note the use of <c>${ROOT}</c>.
 
 <codesample lang="ebuild">
 pkg_config() {
-	if [ ! -d "${ROOT}"/var/lib/mysql/mysql ] ; then
+	if [[ ! -d "${ROOT}"/var/lib/mysql/mysql ]] ; then
```
ulm, line 51: No quotation marks here.

ebuild-writing/functions/src_compile/building/text.xml
```diff
@@ -60,7 +60,7 @@ src_prepare() {
 	# We have a weird Makefile to work with which ignores our
 	# compiler preferences. yay!
 	# Note the single quotes.
-	sed -i -e 's:cc -O2:$(CC) $(CFLAGS) $(LDFLAGS):' Makefile \
+	sed -i -e 's;cc -O2;$(CC) $(CFLAGS) $(LDFLAGS);' Makefile \
```
ulm, line 63: Why this change? We should stick to the common separator characters (like `/` or `:`). Unusual ones like `;` only make the reader wonder if there's a reason for it.

ulm (changes requested):

ebuild-writing/functions/src_compile/building/text.xml
thesamesam: Did you read the commit message? :)

```
Author: Sam James <sam@gentoo.org>
Date:   Sat Mar 20 09:55:09 2021 +0000

    ebuild-writing/functions/src_compile/building: use ; in CFLAGS sed

    A colon is a valid character in CFLAGS, so we shouldn't use it
    as a sed delimiter.

    Bug: https://bugs.gentoo.org/685160
```

ebuild-writing/functions/pkg_config/text.xml
thesamesam: Really? This is common style.

ulm: Which doesn't make it right. `[[ ]]` is a syntax element, and no word splitting or globbing will take place. So the quotation marks are excessive (but they were necessary inside `[ ]`).

ebuild-writing/functions/src_compile/building/text.xml
ulm: I don't see any colon in `$(CC) $(CFLAGS) $(LDFLAGS)` (and neither in `cc -O2`).

Maybe I need new glasses? :smile:

ulm: Ah, I think I understand now.
Note the single quotes, `$(CFLAGS)` is a literal string here.

ulm: Also, if it wasn't a literal string, then `;` would only marginally be better than `:`. CFLAGS originates from user configuration, so one cannot assume anything about it, and it may contain _any_ character. Expanding the variable in a sed replacement is simply asking for trouble and should be avoided.

thesamesam: Ahhh, I missed it was literal! It's still extremely common to end up sedding CFLAGS but I take the point that it's a rabbit hole.

I'd still say that using something other than a comma or a colon is ideal for where people don't want to fix the build system though. The example in the devmanual should probably be 'less wrong', but maybe it'd be useful to put a note about this as well somewhere?

ebuild-writing/functions/pkg_config/text.xml
thesamesam: You're right, of course, but I've noticed a common theme between both of the changes I've suggested here.

Both are imperfect (and one is unnecessary) but both _suggest_ better habits, even if imperfect. I'm happy to drop it here though.

ebuild-writing/functions/src_compile/building/text.xml
thesamesam: Or maybe just a warning in the sed section about choosing your delimiter?

ebuild-writing/functions/pkg_config/text.xml
ulm: Yes, please drop the quotes. If you leave them in place, there's little point in using `[[ ]]` instead of `[ ]`.

Of course this is very minor, so I wouldn't file a bug for seeing quotes in `[[ ]]`. But I would comment about it in eclass review.

ebuild-writing/functions/src_compile/building/text.xml
ulm: We say something about it [here](https://devmanual.gentoo.org/tools-reference/sed/index.html#simple-text-substitution-using-sed).

ebuild-writing/functions/pkg_config/text.xml
thesamesam: Understood. Let's fix that.

ebuild-writing/functions/src_compile/building/text.xml
thesamesam: I've added some more information there.

ebuild-writing/common-mistakes/text.xml
```diff
@@ -259,7 +259,7 @@ part.
 </body>
 </subsection>
 <subsection>
-<title>DEPEND is incomplete</title>
+<title>(B/R)DEPEND is incomplete</title>
```
ulm, line 262: Is this heading change necessary? It will break external links to the section.

Maybe mention instead in the first section that this applies to all dependency variables?

ebuild-writing/functions/src_compile/building/text.xml
```diff
@@ -59,8 +59,8 @@ src_prepare() {
 
 	# We have a weird Makefile to work with which ignores our
 	# compiler preferences. yay!
-	# Note the single quotes.
-	sed -i -e 's:cc -O2:$(CC) $(CFLAGS) $(LDFLAGS):' Makefile \
+	# Note the single quotes (hence the delimiter is not an issue)
+	sed -i -e 's;cc -O2:$(CC) $(CFLAGS) $(LDFLAGS):' Makefile \
```
ulm, line 63: Missed one.

general-concepts/dependencies/text.xml
```diff
@@ -461,6 +461,12 @@ DEPEND="!build? (
 <title>Any of Many Dependencies</title>
 <body>
 
+<note>
+To ease dependency resolution for Portage, it is recommended that you sort
+the atoms in <e>preferred</e> order first. See
```
ulm, line 466: No "atoms" please. We're trying to abandon this term ever since [bug 174322](https://bugs.gentoo.org/174322) (with mixed success, I must admit).

general-concepts/dependencies/text.xml
```diff
@@ -626,12 +632,13 @@ package:
   </dd>
   <dt>
     Remember non-binary dependencies such as pkg-config, doc generation
-    programs, etc.
+    programs, etc. Such programs would usually belong in <c>BDEPEND</c>
```
ulm, line 635: Full stop.

general-concepts/dependencies/text.xml
```diff
@@ -626,12 +632,13 @@ package:
...
   </dt>
   <dd>
     Usually the build process requires some dependencies such as intltool,
     libtool, pkg-config, doxygen, scrollkeeper, gtk-doc, etc. Make sure those
-    are clearly stated.
+    are clearly stated. Again, such dependencies usually belong in
+    <c>BDEPEND</c>.
```
ulm, line 641: Line break is not necessary here.

tools-reference/sed/text.xml
```diff
@@ -65,7 +65,7 @@ constructs are specific to <c>GNU sed 4</c> <d/> on non-GNU userland archs, the
 guaranteed to be installed as part of <c>@system</c>. This was not
 always the case,
 which is why some packages, particularly those which use <c>sed -i</c>, have
-<c>DEPEND</c> s upon <c>&gt;=sys-apps/sed-4</c>.
+<c>BDEPEND</c> s upon <c>&gt;=sys-apps/sed-4</c>.
```
ulm, line 68: What's that lonely "s" doing here?

Also, this is about old packages, which most certainly have &gt;=sed-4 in DEPEND, not in BDEPEND.

ulm (commented):
While I previously said that combining several small issues in one PR is fine, *adding* unrelated issues to an existing PR definitely isn't, especially after a first round of review has taken place. It makes it a) hard to see what has changed in the existing commits, and b) makes review of the new stuff more difficult.

So please, file a separate PR next time. (Don't spilt this one, though.)

tools-reference/sed/text.xml
thesamesam: Oh, duh. Let's fix the s though.

general-concepts/dependencies/text.xml
ulm: Also, more specifically, [bug 603970](https://bugs.gentoo.org/603970) for the devmanual.

ebuild-writing/common-mistakes/text.xml
thesamesam: Done.

general-concepts/dependencies/text.xml
thesamesam: Ah, I should've remembered that. I'm not sure this is realistic until Portage itself doesn't use 'atom' in its output though. That's exactly where I picked it up.

ulm (approved):

ulm:
I've cherry-picked some trivial things to get them out of the way. Please rebase.

ulm:
Also, can you drop that change of `:` to `;` from commit 55acfbc8995253760d2cd33ca2c6e0311bc9c236, instead of reverting it in a later commit?

thesamesam:
Done.

general-concepts/dependencies/text.xml
```diff
@@ -461,6 +461,12 @@ DEPEND="!build? (
 <title>Any of Many Dependencies</title>
 <body>
 
+<note>
+To ease dependency resolution for Portage, it is recommended that you sort
+the elements in <e>preferred</e> order first. See
+<uri link="https://bugs.gentoo.org/489458">this bug</uri> for more context.
```
ulm, line 467: I'd prefer "[bug 489458](https://bugs.gentoo.org/489458)" instead of "[this bug](https://bugs.gentoo.org/489458)" here. I'll update this when merging, if you don't mind.

thesamesam: Completely fine with me. I'll keep this style preference in mind in future.

ulm:
Thanks, merged.

PR #206, opened 2021-03-21
*/*: further miscellaneous fixes

thesamesam:


ebuild-maintenance/new-ebuild/text.xml
```diff
@@ -24,9 +24,14 @@ The Gentoo repository should only be used for storing <c>.ebuild</c> files,
...
 in the <c>mycat/mypkg/files</c> directory, which can be referenced as
-<c>${FILESDIR}</c> from within ebuilds. Larger patch files should be placed
-in your developer space on <c>dev.gentoo.org</c> instead, in order to minimize
-the repository size.
+<c>${FILESDIR}</c> from within ebuilds.
+</p>
+
+<p>Larger patch files should be placed in your developer space on <c>dev.gentoo.org</c>
+instead of in the tree, in order to minimize the repository size. Non-developers
+such as proxied-maintainers can host patches on e.g. GitHub or some other stable
```
sbraz, line 32: I don't think we need a hyphen in "proxied maintainers".

ebuild-maintenance/new-ebuild/text.xml
```diff
@@ -24,9 +24,14 @@ The Gentoo repository should only be used for storing <c>.ebuild</c> files,
...
 files. These files should have a size of at most 20 KiB and must be placed
 in the <c>mycat/mypkg/files</c> directory, which can be referenced as
-<c>${FILESDIR}</c> from within ebuilds. Larger patch files should be placed
-in your developer space on <c>dev.gentoo.org</c> instead, in order to minimize
-the repository size.
+<c>${FILESDIR}</c> from within ebuilds.
+</p>
+
+<p>Larger patch files should be placed in your developer space on <c>dev.gentoo.org</c>
+instead of in the tree, in order to minimize the repository size. Non-developers
```
sbraz, line 31: Maybe just "instead of the tree"?

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -30,15 +30,17 @@ mirrored</uri>, preferably on the Gentoo Infrastructure. When
...
+for more information.
 </p>
 
 <note>
 Patches included in <c>${FILESDIR}</c> should never be compressed.
 </note>
 
 <warning>
-Starting from EAPI=6 a strip patch level was limited to the <c>-p1</c>.
+Starting from EAPI=6, the patch strip level was limited to the <c>-p1</c>.
```
sbraz, line 43: I'd say something like
> Starting from EAPI=6, the patch strip level defaults to `-p1`.

I feel like _limited_ sounds weird.

tasks-reference/pam/text.xml
```diff
@@ -7,7 +7,7 @@
 <p>
 PAM (Pluggable Authentication Modules) is a mechanism which allows different
 applications to authenticate using various specified parameters, using for
-example a passwd/shadow file, a Kerberos server, an LDAP server or an a NT
+example a passwd/shadow file, a Kerberos server, an LDAP server, or an a NT
```
sbraz, line 10: "an a NT" → "a NT"

ulm: "an NT"

ebuild-maintenance/new-ebuild/text.xml
ulm, line 32: I wouldn't recommend any specific hosting here, especially not a non-free one like github.

ebuild-writing/functions/pkg_nofetch/text.xml
```diff
@@ -33,13 +33,13 @@
 <body>
 <codesample lang="ebuild">
 pkg_nofetch() {
-	[ -z "${SRC_URI}" ] &amp;&amp; return
+	[[ -z "${A}" ] &amp;&amp; return
```
ulm, line 36: Again, omit the quotation marks. Don't mix different delimiters.

ebuild-writing/misc-files/patches/text.xml
ulm: +1
"to the -p1" sounds strange as well.

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -66,8 +68,8 @@ later. Good things to include in comments are:
     What the patch actually does. Bug numbers are good here.
   </li>
   <li>
-    Where the patch came from. Is it an upstream VCS pull,
-    something from Bugzilla, something you wrote?
+    Where the patch came from. e.g. Is it from an upstream VCS,
```
ulm, line 71: I'd write "For example, is it from ..." here.

ebuild-writing/user-submitted/text.xml
```diff
@@ -29,14 +29,15 @@ commit message format section</uri>, is highly encouraged.
 Note that ebuilds received in the form of git patches or pull requests
 will have the user as the commit author by default, in which case
 including the user information in the commit message explicitly
-may not be necessary.
+may not be necessary. If that is not the case, you may wish to use
+git commit's <c>--author</c> parameter to explicitly give them credit.
```
ulm, line 33: This is problematic. If the patch doesn't originate from `git format-patch` (i.e. doesn't have the correct author), then chances are that it will also miss a `Signed-off-by:` line and therfore violate our copyright policy.

So I'd rather discourage using the --author option. Any nontrivial patches should be submitted with format-patch, and trivial (= not copyrightable) suggestions can be tagged with `Suggested-by:` or `Reported-by:` as the section already says.

tools-reference/cat/text.xml
```diff
@@ -73,7 +73,7 @@ tells <c>cat</c> to strip leading tabs (please note that when you want
 to copy the example, you have to replace the leading spaces with
 tabs), so that, as Azarah puts it, "we won't have the ebuilds looking
 from space effect". For such small files as the example above, a set
-of here documents is more elegant, and easier to maintain, than the
+of here documents (heredocs) is more elegant, and easier to maintain, than the
```
ulm, line 76: I've never heard the term "heredocs" before. The bash manual doesn't mention it either.

ebuild-maintenance/new-ebuild/text.xml
```diff
@@ -24,9 +24,14 @@ The Gentoo repository should only be used for storing <c>.ebuild</c> files,
 as well as any small companion files, such as patches or sample configuration
 files. These files should have a size of at most 20 KiB and must be placed
 in the <c>mycat/mypkg/files</c> directory, which can be referenced as
-<c>${FILESDIR}</c> from within ebuilds. Larger patch files should be placed
-in your developer space on <c>dev.gentoo.org</c> instead, in order to minimize
-the repository size.
+<c>${FILESDIR}</c> from within ebuilds.
+</p>
+
+<p>Larger patch files should be placed in your developer space on <c>dev.gentoo.org</c>
```
ulm, line 30: `<p>` goes on a line of its own.
Please wrap lines at 80 chars

tools-reference/cat/text.xml
thesamesam: Wikipedia:
>In computing, a here document (here-document, here-text, heredoc, hereis, here-string or here-script) is a file literal or input stream literal: [...]

If you _really_ want, we can drop it, but I think adding jargon is useful so people learn.

ebuild-writing/user-submitted/text.xml
thesamesam: I don't think that's likely, actually.. you should obviously be checking for our ```Signed-off-by``` either way, and often, contributors on Bugzilla will attach an ebuild and post the sign-off in a comment when requested.

@gktrk was generally a fan of this suggestion in the past. Of course, ```format-patch``` is preferred, but not always doable by users.

ebuild-writing/misc-files/patches/text.xml
thesamesam: Thanks for the suggestions.

ebuild-maintenance/new-ebuild/text.xml
thesamesam: Sure.

tools-reference/cat/text.xml
ulm: Yes, please drop. "Here document" is clear enough, and is the term used in the Bash manual (most of the time, with an hyphen, so arguably we could do that as well).

ebuild-writing/user-submitted/text.xml
ulm: > [...] you should obviously be checking for our `Signed-off-by` either way, [...]

Mention that then?

thesamesam: Good plan.

tools-reference/cat/text.xml
thesamesam: Sure.

thesamesam:
OK?

ulm:
Looks good from my side, but let's give the team the usual week (counting from the last push) for review.

ulm:
Merged (with some whitespace changes).

PR #208, opened 2021-03-21
ebuild-writing/variables: mention style, RESTRICT mini-section, etc

thesamesam:


ebuild-writing/variables/text.xml
```diff
@@ -524,6 +524,23 @@ GLEP 23</uri> for details.
 </body>
 </section>
 
+<section>
+<title>RESTRICT</title>
+<body>
+
+<p>
+While Portage may recognise several different <c>RESTRICT</c> tokens, it is
```
sbraz, line 532: I think we use American English in the rest of the manual → recognize.

ebuild-writing/variables/text.xml
```diff
@@ -313,7 +313,7 @@ The following variables may or must be defined by every ebuild.
     <ti><c>PDEPEND</c></ti>
     <ti>
     A list of packages to be installed (if possible) after the package
-    is merged. Use this only when <c>RDEPEND</c> would cause cyclic
+    is merged. Use this <b>only</b> when <c>RDEPEND</c> would cause cyclic
```
ulm, line 316: `<e>` for normal emphasis here.

ebuild-writing/variables/text.xml
ulm: No, we definitely use British English: https://devmanual.gentoo.org/appendices/contributing/index.html#style-guidelines

ebuild-writing/variables/text.xml
```diff
@@ -524,6 +524,23 @@ GLEP 23</uri> for details.
...
 
+<section>
+<title>RESTRICT</title>
+<body>
+
+<p>
+While Portage may recognise several different <c>RESTRICT</c> tokens, it is
+important that you do not rely on them existing. That is, you should ensure
+your ebuild does not fail if those tokens are not exposed or given a different
+name by another package manager. You can make use of Portage-provided <c>RESTRICT</c>
```
ulm, line 535: Line length.

ebuild-writing/variables/text.xml
```diff
@@ -614,8 +632,8 @@ to read.
 <p>
 Some ebuilds use calls to <c>sed</c>, <c>awk</c> and / or <c>cut</c> to do this. This
 must <e>not</e> be done for any new code, and should be fixed to use either
-eapi7-ver or bash substitution where possible. Global scope non-bash code is
-highly discouraged.
+<c>eapi7-ver</c> (before EAPI 7) or Bash substitution where possible. Global scope
+non-Bash code is <b>strongly</b> discouraged.
```
ulm, line 636: I think this is misleading. The possibilities are a) Bash substitution for the cases where it is possible, b) version manipulation commands in EAPI 7, and c) eapi7-ver in older EAPIs.

ulm (changes requested):

ulm:
Merged.

PR #210, opened 2021-03-21
ebuild-writing/file-format: style notes (blank line after EAPI, inherit)

thesamesam:


ebuild-writing/file-format/text.xml
```diff
@@ -221,7 +232,7 @@ positions wide, and multibyte characters are just one position wide.
 <title>Character Set</title>
 <body>
 <p>
-All ebuilds (and eclasses, metadata files and ChangeLogs) must use the
+All ebuilds (and eclasses, metadata files, etc) must use the
```
sbraz, line 235: Isn't it usually `etc.`?

ebuild-writing/file-format/text.xml
```diff
@@ -177,9 +179,18 @@ exact copy of the contents of
 header.txt</uri></c> in the top directory of the Gentoo repository.
 </p>
 
+<p>
+It is good style to have one blank line after the EAPI line and another blank
+line after any inherits.
+</p>
```
ulm, line 185: This doesn't belong here, because it is not related to the ebuild header.

ebuild-writing/file-format/text.xml
```diff
@@ -177,9 +179,18 @@ exact copy of the contents of
...
+
 <codesample lang="ebuild">
 # Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit foo
```
ulm, line 193: Again, this doesn't belong here. The codesample is supposed to show `header.txt` which contains neither an EAPI nor an inherit line.

ulm (changes requested):

ebuild-writing/file-format/text.xml
thesamesam: Can you please suggest where this (and the other item) _should_ go?

ebuild-writing/file-format/text.xml
thesamesam: Yep!

ebuild-writing/file-format/text.xml
ulm: Not in this chapter. ebuild-writing/using-eclasses may be a good place.

ulm:
I've merged the second and third commit, as they're the ones applying to ebuild-writing/file-format.

ulm:
Not sure how to proceed here. All commits except the first have been merged.

Maybe the cleanest approach would be to close this PR and open a new one for `ebuild-writing/eapi` and/or `ebuild-writing/using-eclasses`? OTOH, I'm not sure if we should really give advice about blank lines, apart from what we show in examples.

ebuild-writing/file-format/text.xml
```diff
@@ -177,9 +177,18 @@ exact copy of the contents of
...
```
Zlogene, line 193: can't we use more natural inherit

Zlogene (changes requested):

ulm:
> Not sure how to proceed here. All commits except the first have been merged.
> 
> Maybe the cleanest approach would be to close this PR and open a new one for `ebuild-writing/eapi` and/or `ebuild-writing/using-eclasses`? OTOH, I'm not sure if we should really give advice about blank lines, apart from what we show in examples.

No progress, therefore closing per my previous comment.

PR #212, opened 2021-03-21
*/*: add links to relevant QA policy

thesamesam:


ulm:
Nope. We don't want sections to start with notes referring to other documents (or otherwise we would also have links to PMS, GLEPs, etc. all the time). The devmanual is supposed to be useful stand-alone, so if there's any important QA policy missing, then it should be copied to here.

Similarly, the devmanual never referred to the devrel handbook.

Idea: Maybe a section could be added to appendices/further-reading, mentioning such other documents like PMS or QA policy?

PR #214, opened 2021-03-22
eclass-writing: Add more eclass doc tags, information on etiquette

thesamesam:
- Add more eclass doc tags. Note that I wasn't sure if they were left out for a reason or just nobody got to it yet. Please consider this part more of a draft. 

Specifically:
```
* @PRE_INHERIT
* @USER_VARIABLE
* @OUTPUT_VARIABLE
* @INCLUDES_EPREFIX
* @DEPRECATED
```

- Mention etiquette when modifying eclasses

- Various extra eclass writing information/tips/terminology

### DIFF
--- a/eclass-writing/text.xml
+++ b/eclass-writing/text.xml
@@ -90,6 +90,17 @@ Committing a broken eclass can kill huge numbers of packages. Since
 A simple way to verify syntax is <c>bash -n foo.eclass</c>.
 </p>
 
+<note>
+Given that updating an eclass will invalidate the cache of all consumer ebuilds,
+it is good etiquette to ping other developers in e.g. <c>#gentoo-dev</c> on IRC
+or informally determine if there are other similar changes (mainly documentation)
+which should be pushed at the same time in order to avoid unnecessary cache
+regeneration within a few hours or days of each other.
+
+The same applies to your own work <d/> please prepare all of your commits and
+batch push them at once where possible.
+</note>
+
 </body>
 </section>
 
--- a/eclass-writing/text.xml
+++ b/eclass-writing/text.xml
@@ -568,12 +568,8 @@ using the following tags:
   <th>description</th>
 </tr>
 <tr>
-  <ti>
-    <c>@VARIABLE:</c>
-  </ti>
-  <ti>
-    NO
-  </ti>
+  <ti><c>@VARIABLE:</c></ti>
+  <ti>NO</ti>
   <ti>
     Name of the function-specific variable to which the documentation applies.
   </ti>
@@ -584,61 +580,53 @@ using the following tags:
   </ti>
 </tr>
 <tr>
-  <ti>
-    <c>@DEFAULT_UNSET</c>
-  </ti>
-  <ti>
-    YES
-  </ti>
-  <ti>
-    <d/>
-  </ti>
+  <ti><c>@USER_VARIABLE</c></ti>
+   <ti>YES</ti>
+   <ti><d/></ti>
+   <ti>
+     This tag describes whether the variable is unsuitable for use in ebuilds,
+     i.e. if it is solely for user consumption via e.g. make.conf or a similar
+     mechanism.
+   </ti>
+</tr>
+<tr>
+  <ti><c>@DEFAULT_UNSET</c></ti>
+  <ti>YES</ti>
+  <ti><d/></ti>
   <ti>
     Indicates that this variable is unset by default if not set by the
     developer.
   </ti>
 </tr>
 <tr>
-  <ti>
-    <c>@INTERNAL</c>
-  </ti>
-  <ti>
-    YES
-  </ti>
-  <ti>
-    <d/>
-  </ti>
-  <ti>
-    Indicates that the variable is internal to the eclass function.
-  </ti>
+  <ti><c>@INCLUDES_EPREFIX</c></ti>
+  <ti>YES</ti>
+  <ti><d/></ti>
+  <ti>Indicates that the variable is a path which has ${EPREFIX} prepended.</ti>
 </tr>
 <tr>
-  <ti>
-    <c>@REQUIRED</c>
-  </ti>
-  <ti>
-    YES
-  </ti>
-  <ti>
-    <d/>
-  </ti>
-  <ti>
-    Indicates that this variable must be set by the developer.
-  </ti>
+  <ti><c>@INTERNAL</c></ti>
+  <ti>YES</ti>
+  <ti><d/></ti>
+  <ti>Indicates that the variable is internal to the eclass function.</ti>
 </tr>
 <tr>
-  <ti>
-    <c>@DESCRIPTION:</c>
-  </ti>
-  <ti>
-    NO
-  </ti>
-  <ti>
-    Multiline freetext.
-  </ti>
-  <ti>
-    Long description for the function variable.
-  </ti>
+  <ti><c>@REQUIRED</c></ti>
+  <ti>YES</ti>
+  <ti><d/></ti>
+  <ti>Indicates that this variable must be set by the developer.</ti>
+</tr>
+<tr>
+  <ti><c>@DEPRECATED</c></ti>
+  <ti>YES</ti>
+  <ti>Optionally, the name of any replacement variable.</ti>
+  <ti>Declares that this variable should no longer be used in ebuilds.</ti>
+</tr>
+<tr>
+  <ti><c>@DESCRIPTION:</c></ti>
+  <ti>NO</ti>
+  <ti>Multiline freetext.</ti>
+  <ti>Long description for the function variable.</ti>
 </tr>
 </table>
 
--- a/eclass-writing/text.xml
+++ b/eclass-writing/text.xml
@@ -658,10 +658,10 @@ domacosapp() {
 
 <p>
 An eclass may provide default implementations for any of the ebuild phase
-functions (<c>src_unpack</c>, <c>pkg_postinst</c> etc). This can be done either as a
+functions (<c>src_unpack</c>, <c>pkg_postinst</c>, etc). This can be done either as a
 simple function definition (which is not multiple eclass friendly) or using
 <c>EXPORT_FUNCTIONS</c>. Functions given to <c>EXPORT_FUNCTIONS</c> are implemented
-as normal, but have their name prefixed with <c>${ECLASS}_</c>.
+as normal, but have their name prefixed ("namespaced") with <c>${ECLASS}_</c>.
 </p>
 
 <important>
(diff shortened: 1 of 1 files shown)


eclass-writing/text.xml
```diff
@@ -47,11 +47,23 @@ Roughly speaking, there are three kinds of eclass:
...
+within a few hours or days of each other.
+
+The same applies to your own work <d/> please prepare all of your commits and
+batch push them at once where possible.
+</note>
+
 <p>
 Before committing a new eclass to the tree, it must be emailed to the gentoo-dev
-mailing list with a justification and a proposed implementation. Do not skip
+mailing list with a justification, and a proposed implementation. Do not skip
```
sbraz, line 63: I don't think this comma is useful.

ulm, line 63: These are only two terms, so I don't think the comma is right.

eclass-writing/text.xml
```diff
@@ -47,11 +47,23 @@ Roughly speaking, there are three kinds of eclass:
...
 this step <d/> sometimes a better implementation or an alternative which
-does not require a new eclass will be suggested.
+does not require a new eclass will be suggested. It might even be that a
+similar eclass has existed in the past which you weren't aware of!
```
ulm, line 66: Does this case happen so often that it's worth mentioning? I think it can be subsumed under "better implementation" above.

eclass-writing/text.xml
```diff
@@ -47,11 +47,23 @@ Roughly speaking, there are three kinds of eclass:
 <title>Adding and Updating Eclasses</title>
 <body>
 
+<note>
+Given that updating eclasses will invalidate the cache of all consumer ebuilds,
+it is good etiquette to ping other developers in e.g. <c>#gentoo-dev</c> on IRC
+or informally determine if there are other similar changes (mainly documentation)
+which should be pushed at the same time in order to avoid unnecessary cache regeneration
```
ulm, line 54: Line length.

eclass-writing/text.xml
```diff
@@ -327,6 +350,53 @@ variables are as follows:
...
+    <c>@OUTPUT_VARIABLE</c>
+  </ti>
+  <ti>
+    YES
+  </ti>
+  <ti>
+    <d/>
+  </ti>
+  <ti>
+    This tag indicates the variable's value (which will be set by the eclass)
```
sbraz, line 396: I think adding "that" makes the sentence less ambiguous:
> This tag indicates _that_ the variable's value (which will be set by the eclass) should be read within an ebuild.

When I first read the existing sentence, I thought that it meant "This tag indicates the variable's value".

eclass-writing/text.xml
```diff
@@ -327,6 +350,53 @@ variables are as follows:
     this tag. Must appear as the first tag in the comment block.
   </ti>
 </tr>
+<tr>
+  <ti>
+    <c>@PRE_INHERIT</c>
+  </ti>
```
ulm, line 356: This (and all the following ones) will fit in one line:
```suggestion
  <ti><c>@PRE_INHERIT</c></ti>
```

See https://devmanual.gentoo.org/appendices/devbook-guide/index.html#internal-coding-style

eclass-writing/text.xml
thesamesam: I wanted to raise the idea that it's worth consulting the past. For example, a newer developer didn't know about PROVIDES and was casually chatting with me about its benefits. They were really surprised to learn it used to be a thing!

How about we compromise and say "better (or relevant historical) implementation", which is shorter?

eclass-writing/text.xml
thesamesam: Was too sleepy. Absolutely.

eclass-writing/text.xml
thesamesam: Agreed, thank you!

eclass-writing/text.xml
thesamesam: Yeah, this should go.

eclass-writing/text.xml
ulm: But this has parentheses which should be used sparingly.

I really think that the two examples "better implementation" and "no eclass needed" are enough justification here. We don't need a complete enumeration of all possible things that could happen. IMHO it would only distract from the main message of the paragraph.

eclass-writing/text.xml
```diff
@@ -342,6 +412,20 @@ variables are as follows:
...
+    <c>@INCLUDES_EPREFIX</c>
+  </ti>
+  <ti>
+    YES
+  </ti>
+  <ti>
+    <d/>
+  </ti>
+  <ti>
+    Indicates that the variable is a path which has ${EPREFIX} prepended.
```
sbraz, line 426: "prepended to it"

sbraz, line 426: "prepended to it"

eclass-writing/text.xml
```diff
@@ -472,6 +570,20 @@ documentation are:
...
+    <c>@INCLUDES_EPREFIX</c>
+  </ti>
+  <ti>
+    YES
+  </ti>
+  <ti>
+    <d/>
+  </ti>
+  <ti>
+    Indicates whether the function returns a path with ${EPREFIX} prepended.
```
sbraz, line 584: "prepended to it"

eclass-writing/text.xml
thesamesam: Okay, removed.

thesamesam:
All comments should've been addressed. I've added 80eedff2aba11e778d032f515ccbfd99d3e85dea (mentioning ```@DEPRECATED``` in the removal section) as the PR was still marked as a draft - I can remove it if needed though.

Let's get this in -- it's tremendously useful when trying to address eclassdoc issues in the tree.

eclass-writing/text.xml
```diff
@@ -47,6 +47,17 @@ Roughly speaking, there are three kinds of eclass:
...
+<note>
+Given that updating eclasses will invalidate the cache of all consumer ebuilds,
+it is good etiquette to ping other developers in e.g. <c>#gentoo-dev</c> on IRC
+or informally determine if there are other similar changes (mainly documentation)
+which should be pushed at the same time in order to avoid unnecessary cache
+regeneration within a few hours or days of each other.
+
+The same applies to your own work <d/> please prepare all of your commits and
+batch push them at once where possible.
+</note>
```
ulm, line 59: This applies to the major eclasses only. For example, it is not a concern if I update elisp.eclass because ebuilds inheriting it typically don't inherit anything else.

Also, this note shouldn't go to the beginning of the section.

eclass-writing/text.xml
```diff
@@ -94,6 +108,12 @@ A simple way to verify syntax is <c>bash -n foo.eclass</c>.
 <title>Removing Eclasses</title>
 <body>
 
+<note>
+Before considering removal, please mark the eclass as <c># @DEPRECATED</c>
+to ensure consumers are warned if they are still using the eclass. Do this
+for a period before removing the eclass.
+</note>
```
ulm, line 115: Again, this note shouldn't go to the beginning of the section.

eclass-writing/text.xml
```diff
@@ -644,10 +682,10 @@ domacosapp() {
 
 <p>
 An eclass may provide default implementations for any of the ebuild phase
-functions (<c>src_unpack</c>, <c>pkg_postinst</c> etc). This can be done either as a
+functions (<c>src_unpack</c>, <c>pkg_postinst</c>, etc). This can be done either as a
 simple function definition (which is not multiple eclass friendly) or using
 <c>EXPORT_FUNCTIONS</c>. Functions given to <c>EXPORT_FUNCTIONS</c> are implemented
-as normal, but have their name prefixed with <c>${ECLASS}_</c>.
+as normal, but have their name prefixed ("namespaced") with <c>${ECLASS}_</c>.
```
ulm, line 688: "Namespaced"? Even if this is a word (it's neither in the Macmillan nor in the Merriam-Webster dictionary), it looks strange. It also doesn't add much information, the description is quite clear without it. My suggestion would be to drop this change.

ulm (changes requested):
I've seen you adding notes to the very beginning of sections. IMHO that's less than optimal. At least the basic concepts of the section should be explained first (and often the note will only be understandable after reading that explanation first).

So, typically notes should go _after_ the paragraph/subsection/section they apply to.

eclass-writing/text.xml
SoapGentoo: do we have to come up with arbitrary thresholds now for deciding when something is "major" enough to cause needless cache churn? I fully agree with this and would like to have more devs abide by it.

eclass-writing/text.xml
SoapGentoo: where should it go instead then? Please provide actual suggestions instead of just pure criticism

eclass-writing/text.xml
SoapGentoo: https://en.wiktionary.org/wiki/namespace#Verb
Please, let's not bikeshed this to hell. This is a *very* common term, it's in my company's style guide.

eclass-writing/text.xml
ulm: As I said, it makes little sense to coordinate updates of eclasses like elisp.eclass (or kernel-2.eclass, toolchain.eclass, texlive-module.eclass) which are pretty much orthogonal to other eclasses.

eclass-writing/text.xml
ulm: See above:

> I've seen you adding notes to the very beginning of sections. IMHO that's less than optimal. At least the basic concepts of the section should be explained first (and often the note will only be understandable after reading that explanation first).
>
> So, typically notes should go after the paragraph/subsection/section they apply to.

eclass-writing/text.xml
ulm: Maybe I've misunderstood that note. Reading it again, it seems to apply to independent changes within _the same_ eclass, and then it makes sense of course.

Changing it to singular may help avoiding that ambiguity:
```suggestion
<note>
Given that updating an eclass will invalidate the cache of all consumer ebuilds,
it is good etiquette to ping other developers in e.g. <c>#gentoo-dev</c> on IRC
or informally determine if there are other similar changes (mainly documentation)
which should be pushed at the same time in order to avoid unnecessary cache
regeneration within a few hours or days of each other.

The same applies to your own work <d/> please prepare all of your commits and
batch push them at once where possible.
</note>
```

thesamesam: Sounds good.

eclass-writing/text.xml
thesamesam: It would've been helpful to directly specify where in this case, as I wasn't really sure where could be more clear than the beginning.

eclass-writing/text.xml
thesamesam: This is common jargon which developers are going to use, let's keep it.

ulm:
Merged.

PR #215, opened 2021-04-02
general-concepts/use-flags: Fix indentation in code example

Whissi:
Fix indentation [in code example](https://devmanual.gentoo.org/general-concepts/use-flags/#conflicting-use-flags).


ulm:
Another problem with that example is that it should use src_configure rather than src_compile. #216 now.

PR #216, opened 2021-04-02
general-concepts/use-flags: Change example to src_configure()

ulm:


ulm:
I wonder if we shouldn't rewrite the example to be more compact:
```
	if use ssl; then
		myconf+=" --enable-ssl --with-ssl=$(usex gnutls gnutls openssl)"
	else
		myconf+=" --disable-ssl"
	fi	
```

ulm:
Merged.

PR #217, opened 2021-04-10
ebuild-writing/misc-files/patches: discourage commenting out lines

thesamesam:
We shouldn't add comments (#) within patch changes to source code (etc) because
it's unnecessary noise. Comments *at the top* of the patch are, of course, welcoome.

### DIFF
--- a/ebuild-writing/misc-files/patches/text.xml
+++ b/ebuild-writing/misc-files/patches/text.xml
@@ -96,8 +96,8 @@ from the <c>vim</c> patch tarball:
 <pre>
 # Detect Gentoo apache files properly. Gentoo bug 83565.
 
---- runtime/filetype.vim.orig   2005-03-25 01:44:12.000000000 +0000
-+++ runtime/filetype.vim        2005-03-25 01:45:15.000000000 +0000
+--- a/runtime/filetype.vim
++++ b/runtime/filetype.vim
 @@ -93,6 +93,9 @@
  " Gentoo apache config file locations (Gentoo bug #76713)
  au BufNewFile,BufRead /etc/apache2/conf/*/* setf apache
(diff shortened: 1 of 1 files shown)


ebuild-writing/misc-files/patches/text.xml
```diff
@@ -293,6 +293,14 @@ files automatically. Alternatively, you can specify the <c>-E</c> option with
 <c>eapply</c> if you want to remove an empty file.
 </p>
 
+<p>
+You should also avoid adding commented lines in your patch to the actual
+changes to e.g. source code.  Patches show removed lines by prefixing them with
+a <c>-</c>, so no information is lost by simply deleting lines rather than
+commenting them out (which adds noise). This makes the patch shorter and
+more readable.
+</p>
```
ulm, line 302: I don't understand. If the patch is intended for upstream and comments help understanding the added lines, why would we need the extra step of removing them again? That unnecessary workload for devs, and the benefit is only that the patch files in the gentoo repo are slightly smaller.

thesamesam: Often, submitted patches with commented out lines add nothing and it would have been better to just remove said lines.

I think maybe my phrasing was unclear here? I am referring to patches which include commenting out lots of lines with no real use rather than simply removing them, not explanatory comments

ulm: Yes, "adding commented lines" was confusing. Just say that removed lines should be removed, not commented out.

ulm: And maybe don't refer to any comment char like `#` because it differs between languages.

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -293,6 +293,14 @@ files automatically. Alternatively, you can specify the <c>-E</c> option with
 <c>eapply</c> if you want to remove an empty file.
 </p>
 
+<p>
+Removed lines should not appear in the patch because they are commented <d/> just
```
ulm, line 297: Long line.
Otherwise patch LGTM.

PR #218, opened 2021-04-10
keywording: Mention un-CCing arches after substantial changes

thesamesam:


keywording/text.xml
```diff
@@ -280,6 +280,22 @@ blank or <c>-</c> sanity-check, so please fix this if it occurs <d/> or ask for
...
+the field if you make any changes to the package list to ensure
+they are read by NATTkA.
+</note>
+
+<note>
+If you are changing the package list of a bug after it has been processed,
+<d/> i.e. after NATTkA has CCed arches <d/> you should un-CC all arches so that
+NATTkA CCs the correct new set of arches, in addition to the advice above.
+</note>
+
```
ulm, line 298: That's now three `<note>`s in a row. I wonder if the first one could be changed to a regular paragraph? It's length suggests that it should.

keywording/text.xml
```diff
@@ -280,6 +280,22 @@ blank or <c>-</c> sanity-check, so please fix this if it occurs <d/> or ask for
 help to do so.
 </p>
 
+<note>
+If NATTkA gives a failed sanity check result in a comment or otherwise
+sets the <c>sanity-check</c> field to <c>-</c>, you must clear it by
+setting it to blank in order for NATTkA to consider e.g. changes
+in the package list made in response to any issues. Failure to do
```
mgorny, line 287: This is simply not true. Changes to package list invalidate cached result.

keywording/text.xml
thesamesam: I think this should be obsolete now.

keywording/text.xml
```diff
@@ -280,6 +280,12 @@ blank or <c>-</c> sanity-check, so please fix this if it occurs <d/> or ask for
 help to do so.
 </p>
 
+<note>
+If you are changing the package list of a bug after it has been processed,
+<d/> i.e. after NATTkA has CCed arches <d/> you should un-CC all arches so that
+NATTkA CCs the correct new set of arches, in addition to the advice above.
```
sbraz, line 286: I'd write "In addition" at the beginning of the paragraph and leave out  the "in addition to the advice above" part.

thesamesam: Thanks!

PR #219, opened 2021-04-10
tasks-reference/init-scripts: mention required POSIX compliance

thesamesam:
OpenRC init scripts need to be POSIX compliant and should not therefore
use Bash-specific tests or constructs, unlike ebuilds.

### DIFF
--- a/tasks-reference/init-scripts/text.xml
+++ b/tasks-reference/init-scripts/text.xml
@@ -14,6 +14,12 @@ should be handled via entries in <c>/etc/conf.d</c> with the same filename <d />
 </p>
 
 <p>
+Please note that unlike ebuilds, init scripts for OpenRC are expected to be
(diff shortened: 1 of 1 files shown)


tasks-reference/init-scripts/text.xml
```diff
@@ -13,6 +13,12 @@ should be handled via entries in <c>/etc/conf.d</c> with the same filename <d />
 <c>doconfd</c> or <c>newconfd</c> can be used to install these.
 </p>
 
+<p>
+Please note that unlike ebuilds, init scripts for OpenRC are expected to be
+POSIX-compliant and should therefore avoid e.g. Bash-specific tests and
```
ulm, line 18:
```suggestion
POSIX-compliant and must therefore avoid e.g. Bash-specific tests and
```

PR #220, opened 2021-04-10
ebuild-writing/common-mistakes: add 'Calling pkg-config directly'

thesamesam:
Quite a lot of ebuilds do this. We should instead use the tc-getPKG_CONFIG
helper from toolchain-funcs.eclass which respects ${PKG_CONFIG} - useful
for e.g. cross-compiling.


thesamesam:
Using #223 instead.

PR #222, opened 2021-04-10
ebuild-writing/functions/src_compile/building: add ${CPPFLAGS} to cc …

thesamesam:
…invocation

### DIFF
--- a/ebuild-writing/functions/src_compile/building/text.xml
+++ b/ebuild-writing/functions/src_compile/building/text.xml
@@ -60,7 +60,7 @@ src_prepare() {
 	# We have a weird Makefile to work with which ignores our
 	# compiler preferences. yay!
 	# Note the single quotes (hence the delimiter is not an issue)
-	sed -i -e 's:cc -O2:$(CC) $(CFLAGS) $(LDFLAGS):' Makefile \
+	sed -i -e 's:cc -O2:$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS):' Makefile \
 		|| die "sed fix failed. Uh-oh..."
 }
 


ebuild-writing/functions/src_compile/building/text.xml
```diff
@@ -60,7 +60,7 @@ src_prepare() {
 	# We have a weird Makefile to work with which ignores our
 	# compiler preferences. yay!
 	# Note the single quotes (hence the delimiter is not an issue)
-	sed -i -e 's:cc -O2:$(CC) $(CFLAGS) $(LDFLAGS):' Makefile \
+	sed -i -e 's:cc -O2:$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS):' Makefile \
```
ulm, line 63: I'm sceptical about this one. The example is perfectly clear, so what would we win by packing more variables into it, especially when CPPFLAGS isn't even used in the src_compile() function below.

thesamesam: My view is that we'd benefit from encouraging people to consider an often-forgotten variable, but we should add it to the below part too?

Is that okay with you?

ulm: It is an example for patching the `Makefile` with sed, and as such should be simple. IMHO it would be the wrong place to aim for a complete list of variables such as CPPFLAGS or ASFLAGS. We have a section in [ebuild-writing/variables](https://devmanual.gentoo.org/ebuild-writing/variables/index.html#user-environment) for this.

Also, do users really set CPPFLAGS in their `make.conf`? It's not in Portage's `make.conf.example` at least.

orlitzky: I think `CPPFLAGS` should be there. The compile/link command here is fairly standard, and the only other thing that is potentially missing is `LIBS`. However, the fact that there are no `LIBS` or `-l<foo>` in the original (pre-sed) command indicates that this particular file probably just doesn't link with anything. Thus it's fine to leave off `$LIBS` from the post-sed command.

With that in mind, the "correct" compilation command that e.g. Automake will generate for you is something like

```Make
$(CC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) $<
```

as the PR desires.

Both users and developers set `CPPFLAGS` occasionally, and we file bugs against packages that don't respect them. So ultimately, I don't think there's any slippery slope here (this is all of the variables), and we would be justly chastised in the event that someone had a bug filed against him for something he copy/pasted directly out of the devmanual.

ulm:
Thanks. Closing.

PR #223, opened 2021-04-10
ebuild-writing/common-mistakes: misc additions

thesamesam:
- Leftover flags in IUSE
- Use latest EAPI where possible
- Mention HTTPS for HOMEPAGE

### DIFF
--- a/ebuild-writing/common-mistakes/text.xml
+++ b/ebuild-writing/common-mistakes/text.xml
@@ -503,8 +503,29 @@ the ebuild's quality and ease review.
 </p>
 </body>
 </subsection>
-</section>
 
+<subsection>
+<title>Calling pkg-config directly</title>
+<body>
+
+<p>
+You should not call <c>pkg-config</c> directly in ebuilds because this is
+problematic for e.g. cross-compiling. The correct helper respects
+<c>${PKG_CONFIG}</c>. Instead, use <c>tc-getPKG_CONFIG</c> from
+<c>toolchain-funcs.eclass</c>, e.g.
+</p>
+
+<codesample lang="ebuild">
+sed -i -e "s:-lncurses:$($(tc-getPKG_CONFIG) --libs ncurses):" || die
+</codesample>
+
+<p>
+Don't then forget to add <c>virtual/pkgconfig</c> to BDEPEND!
+</p>
+
+</body>
+</subsection>
+</section>
 
 <section>
 <title>Common Ebuild Submission Mistakes</title>
--- a/ebuild-writing/common-mistakes/text.xml
+++ b/ebuild-writing/common-mistakes/text.xml
@@ -381,6 +381,17 @@ Make sure when you bump a version, the stable KEYWORDS are all marked as
 </body>
 </subsection>
 <subsection>
+<title>Unused flags in IUSE</title>
+<body>
+
+<p>
+Sometimes obsolete USE flags remain in IUSE despite having no function. Remember
+to prune these.
+</p>
+
+</body>
+</subsection>
+<subsection>
 <title>SLOT missing</title>
 <body>
 
--- a/ebuild-writing/common-mistakes/text.xml
+++ b/ebuild-writing/common-mistakes/text.xml
@@ -411,11 +411,12 @@ SLOT="0"
 <body>
 
 <p>
+Make sure the <c>DESCRIPTION</c> is not overly long. Good descriptions will
+describe the main function of the package in a sentence.
+
 Please check if the <c>HOMEPAGE</c> variable is right and leads users to the
-right page if they want to find out more about the package. And make sure
-the <c>DESCRIPTION</c> is not overly long. Good descriptions will describe
-the main function of the package in a sentence. If no homepage for the package
-is available, set the <c>HOMEPAGE</c> variable to
+right page if they want to find out more about the package. If no homepage for
+the package is available, set the <c>HOMEPAGE</c> variable to
 <c>https://wiki.gentoo.org/wiki/No_homepage</c>. Please also strive to test
 HTTPS support for the site used in <c>HOMEPAGE</c>.
 </p>
--- a/ebuild-writing/common-mistakes/text.xml
+++ b/ebuild-writing/common-mistakes/text.xml
@@ -405,7 +405,8 @@ right page if they want to find out more about the package. And make sure
 the <c>DESCRIPTION</c> is not overly long. Good descriptions will describe
 the main function of the package in a sentence. If no homepage for the package
 is available, set the <c>HOMEPAGE</c> variable to
-<c>https://wiki.gentoo.org/wiki/No_homepage</c>.
+<c>https://wiki.gentoo.org/wiki/No_homepage</c>. Please also strive to test
+HTTPS support for the site used in <c>HOMEPAGE</c>.
 </p>
 
 </body>
--- a/ebuild-writing/common-mistakes/text.xml
+++ b/ebuild-writing/common-mistakes/text.xml
@@ -477,6 +477,20 @@ at least when the ebuild hits the stable branch.
 </p>
 </body>
 </subsection>
+
+<subsection>
+<title>Not using latest EAPI</title>
+<body>
+
+<p>
+Often, user-submitted ebuilds do not use the latest EAPI and may even
+be several versions behind. EAPIs bring new helper functions and improved
+methods for completing common tasks. It's recommended that submitted ebuilds
+use the latest EAPI possible (subject to eclass constraints) to improve
+the ebuild's quality and ease review.
+</p>
+</body>
+</subsection>
 </section>
 
 


ebuild-writing/common-mistakes/text.xml
```diff
@@ -380,6 +380,19 @@ Make sure when you bump a version, the stable KEYWORDS are all marked as
 
 </body>
 </subsection>
+<subsection>
+<title>Unused flags in IUSE</title>
+<body>
+
+<p>
+Sometimes vestigal USE flags remain in IUSE despite having no function. Remember
+to prune these.
```
ajakk, line 389: Maybe mention CI tools that would enable something like this to be caught before committing?

ebuild-writing/common-mistakes/text.xml
```diff
@@ -477,8 +491,42 @@ at least when the ebuild hits the stable branch.
...
-</section>
 
+<subsection>
+<title>Not using latest EAPI</title>
+<body>
+<p>
+Often, user-submitted ebuilds do not use the latest EAPI and may even
+be several versions behind. EAPIs bring new helper functions and improved
+methods for completing common tasks. It's recommended that you port submitted
+ebuilds to the latest EAPI possible (subject to eclass constraints) to improve
```
ajakk, line 502: Subject to eclass and base-system (https://wiki.gentoo.org/wiki/Project:Base#Rules_and_limitations) constraints?

ebuild-writing/common-mistakes/text.xml
```diff
@@ -380,6 +380,19 @@ Make sure when you bump a version, the stable KEYWORDS are all marked as
 
 </body>
 </subsection>
+<subsection>
+<title>Unused flags in IUSE</title>
+<body>
+
+<p>
+Sometimes vestigal USE flags remain in IUSE despite having no function. Remember
```
ulm, line 388: "vestigal"? That's an uncommon word, so non-native speakers may have to look it up (I had). Also, looks like it's uncommon enough for you to misspell it :smile: (my dictionary says "vestigial").

ebuild-writing/common-mistakes/text.xml
```diff
@@ -380,6 +380,19 @@ Make sure when you bump a version, the stable KEYWORDS are all marked as
...
+to prune these.
+</p>
+
+</body>
+</subsection>
+
+
```
ulm, line 395: No double blank lines, please (the style guide even doesn't require a single one before `<subsection>`.

ebuild-writing/common-mistakes/text.xml
```diff
@@ -405,7 +418,8 @@ right page if they want to find out more about the package. And make sure
 the <c>DESCRIPTION</c> is not overly long. Good descriptions will describe
 the main function of the package in a sentence. If no homepage for the package
 is available, set the <c>HOMEPAGE</c> variable to
-<c>https://wiki.gentoo.org/wiki/No_homepage</c>.
+<c>https://wiki.gentoo.org/wiki/No_homepage</c>. Please also strive to test
+HTTPS support for the site used in <c>HOMEPAGE</c>.
```
ulm, line 422: Not your fault, but that paragraph talks about HOMEPAGE first, then about DESCRIPTION, then about HOMEPAGE again. Maybe it should be split in two separate paragraphs, first one for DESCRIPTION, second for HOMEPAGE?

ebuild-writing/common-mistakes/text.xml
```diff
@@ -477,8 +491,42 @@ at least when the ebuild hits the stable branch.
...
 
+<subsection>
+<title>Not using latest EAPI</title>
+<body>
+<p>
+Often, user-submitted ebuilds do not use the latest EAPI and may even
+be several versions behind. EAPIs bring new helper functions and improved
+methods for completing common tasks. It's recommended that you port submitted
+ebuilds to the latest EAPI possible (subject to eclass constraints) to improve
+the ebuild's quality and ease review.
```
ulm, line 503: Hm. Is "you port" the best approach? Maybe better to ask the user to update his submission, otherwise you'll have the same problem next time again.

ebuild-writing/common-mistakes/text.xml
```diff
@@ -477,8 +491,42 @@ at least when the ebuild hits the stable branch.
...
+
+<p>
+You should not call <c>pkg-config</c> directly in ebuilds because this is
+problematic for e.g. cross-compiling. The correct helper respects
+<c>${PKG_CONFIG}</c>. Instead, use <c>tc-getPKG_CONFIG</c> from
+<c>toolchain-funcs.eclass</c>, e.g.
+</p>
+
+<codesample lang="ebuild">
+sed -i -e 's/-lncurses/$($(tc-getPKG_CONFIG) --libs ncurses)/' || die
```
ulm, line 520: This happens to work for ncurses, but generally `pkg-config --libs` may contain slashes in its output:
```
$ pkg-config --libs libpq
-L/usr/lib64/postgresql-13/lib64 -lpq
```
Change the separator to `:` at least (it's still not 100% water-proof, but at least it's safer than `/`).

ulm (changes requested):

ebuild-writing/common-mistakes/text.xml
```diff
@@ -504,8 +504,29 @@ the ebuild's quality and ease review.
 </p>
 </body>
 </subsection>
-</section>
 
+<subsection>
```
Zlogene, line 508: How does it differ from any other prefixed tools (e.g. CC/NM/STRING/so on) and why should we use a separate section for this?

ebuild-writing/common-mistakes/text.xml
Zlogene: Please do not include it into "a common mistakes". It is nit a mistake. A person may use whatever EAPI they want as long as it is not deprecated. Maybe make a note but not in common mistakes.

ebuild-writing/common-mistakes/text.xml
thesamesam: No tools cover this as far as I know. They cover leftover flags in metadata.xml but not in the ebuild itself which are noops (it'd be tricky to analyse this right now given it'd involve checking eclasses too).

ebuild-writing/common-mistakes/text.xml
thesamesam: Heh. Okay, good point. Fixing.

ebuild-writing/common-mistakes/text.xml
thesamesam: I've changed it to "It's recommended that submitted ebuilds are..."

ebuild-writing/common-mistakes/text.xml
thesamesam: Well, we can change it to "calling prefixed tools directly". But this is a particularly annoying issue which has far less attention (I've had to fix a lot of examples of it).

ebuild-writing/common-mistakes/text.xml
thesamesam: Done!

ebuild-writing/common-mistakes/text.xml
thesamesam: In theory, sure, it's not a mistake. But every single time a user makes a PR and doesn't bump the EAPI, we will usually ask them to.

ebuild-writing/common-mistakes/text.xml
```diff
@@ -400,12 +411,14 @@ SLOT="0"
...
 Please check if the <c>HOMEPAGE</c> variable is right and leads users to the
-right page if they want to find out more about the package. And make sure
-the <c>DESCRIPTION</c> is not overly long. Good descriptions will describe
-the main function of the package in a sentence. If no homepage for the package
-is available, set the <c>HOMEPAGE</c> variable to
-<c>https://wiki.gentoo.org/wiki/No_homepage</c>.
+right page if they want to find out more about the package. If no homepage for
+the package is available, set the <c>HOMEPAGE</c> variable to
+<c>https://wiki.gentoo.org/wiki/No_homepage</c>. Please also strive to test
+HTTPS support for the site used in <c>HOMEPAGE</c>
```
ulm, line 421:
```suggestion
HTTPS support for the site used in <c>HOMEPAGE</c>.
```

ebuild-writing/common-mistakes/text.xml
```diff
@@ -477,8 +490,42 @@ at least when the ebuild hits the stable branch.
 </p>
 </body>
 </subsection>
-</section>
 
+<subsection>
+<title>Not using latest EAPI</title>
+<body>
+<p>
```
ulm, line 497:
```suggestion
<body>

<p>
```
Blank line after `<body>` and before every `<p>` (see style guide).

ebuild-writing/common-mistakes/text.xml
```diff
@@ -477,8 +490,42 @@ at least when the ebuild hits the stable branch.
...
+
+<p>
+You should not call <c>pkg-config</c> directly in ebuilds because this is
+problematic for e.g. cross-compiling. The correct helper respects
+<c>${PKG_CONFIG}</c>. Instead, use <c>tc-getPKG_CONFIG</c> from
+<c>toolchain-funcs.eclass</c>, e.g.
+</p>
+
+<codesample lang="ebuild">
+sed -i -e 's:-lncurses:$($(tc-getPKG_CONFIG) --libs ncurses):' || die
```
ulm, line 519: This won't work.
```suggestion
sed -i -e "s:-lncurses:$($(tc-getPKG_CONFIG) --libs ncurses):" || die
```

ulm (changes requested):

ulm (approved):

ulm:
OK from my side, but note that there are still unresolved comments from others.

ebuild-writing/common-mistakes/text.xml
```diff
@@ -378,6 +378,17 @@ Make sure when you bump a version, the stable KEYWORDS are all marked as
 <c>~</c>.
 </p>
 
+</body>
+</subsection>
+<subsection>
+<title>Unused flags in IUSE</title>
```
juippis, line 384: While at it, why not include eclasses ;) especially the infamous *eutils.eclass*!
```suggestion
<title>Unused USE flags and eclass inherits in the ebuild</title>
```

ebuild-writing/common-mistakes/text.xml
```diff
@@ -400,12 +411,14 @@ SLOT="0"
 <body>
 
 <p>
+Make sure the <c>DESCRIPTION</c> is not overly long. Good descriptions will
+describe the main function of the package in a sentence.
```
juippis, line 415: I personally wouldn't add this, since it's covered by repoman and pkgcheck already. Can't say this has been a **common** mistake we've seen.

thesamesam: This is just reformatting from [`7ef76ee` (#223)](https://github.com/gentoo/devmanual/pull/223/commits/7ef76ee3e006e2d0e92fe81cfa8f15616db59494)

ebuild-writing/common-mistakes/text.xml
thesamesam: Great idea!

juippis:
Few more ideas I'd include:
 - copy-pasting and old ebuild,
 - use latest EAPI (included), 
 - unguarded external calls (use `|| die`, link to error-handling),
 - GLEP-81.

Since these are pretty common mistakes we see all the time.

thesamesam:
Thanks, I'll work on those newer ideas next!

PR #224, opened 2021-04-10
ebuild-writing/functions: mention defining phases in the canonical order

thesamesam:
This aids readability.


ebuild-writing/functions/text.xml
```diff
@@ -16,6 +16,11 @@ As some phases haven't been introduced from the beginning, you can have a look a
 <uri link="::ebuild-writing/eapi"/> for an overview, what have been introduced in which EAPI.
 </p>
 
+<p>
+Note that ebuilds should usually define phases in the order they are called,
+as set out above, for readability.
```
ulm, line 21: "Note that" seems redundant here.

PR #226, opened 2021-04-13
function-reference/query-functions: document missing pms functions

ajakk:
Signed-off-by: John Helmert III <ajak@gentoo.org>

Documents `usex` and `in_iuse` functions which are in PMS but not documented in the Devmanual as far as I can tell.

### DIFF
--- a/function-reference/query-functions/text.xml
+++ b/function-reference/query-functions/text.xml
@@ -47,7 +47,17 @@ query variables and similar state.
   </tr>
   <tr>
     <ti>
-      <c>use_enable flag str val</c> 
+    <c>usex flag [true output] [false output] [true suffix] [false suffix]</c>
+    </ti>
+    <ti>
+      If <c>flag</c> is enabled, echo [true output][true suffix], otherwise echo [false output][false suffix].
+      If unspecified, true and false outputs are equal to "yes" and "no" respectively. The suffixes
+      default to empty strings.
+    </ti>
+  </tr>
+  <tr>
+    <ti>
+      <c>use_enable flag str val</c>
     </ti>
     <ti>
       Echoes either <c>--enable-str=val</c> or <c>--disable-str</c> depending upon <c>useq flag</c>.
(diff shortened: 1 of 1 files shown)


ulm:
`usex` is documented in https://devmanual.gentoo.org/ebuild-writing/eapi/index.html#eapi-5-helpers and `in_iuse` in https://devmanual.gentoo.org/ebuild-writing/eapi/index.html#eapi-6-helpers, but of course they should be explained in the function reference section, too.

function-reference/query-functions/text.xml
```diff
@@ -47,7 +47,17 @@ query variables and similar state.
   </tr>
   <tr>
     <ti>
-      <c>use_enable flag str val</c> 
+    <c>usex flag [true output] [false output] [true suffix] [false suffix]</c>
+    </ti>
+    <ti>
+      If flag is set, echo [true output][true suffix] (defaults to "yes"),
+      otherwise echo [false output][false suffix] (defaults to "no"). If unspecified,
+      the suffixes default to empty strings.
```
ulm, line 55: From this wording it is not entirely clear that the yes/no defaults are for "true output" and "false output".

function-reference/query-functions/text.xml
```diff
@@ -63,6 +73,14 @@ query variables and similar state.
       As <c>use_enable</c>, but <c>--with-</c> or <c>--without-</c>.
     </ti>
   </tr>
+  <tr>
+    <ti>
+    <c>in_iuse flag</c>
+    </ti>
+    <ti>
+      Returns true if <c>flag</c> is in IUSE_EFFECTIVE, false otherwise.
```
ulm, line 81: I'd rather avoid mentioning IUSE_EFFECTIVE because it is only a conceptual variable in PMS, i.e. it isn't exported to the ebuild environment (and therefore the devmanual doesn't explain it).

Maybe say something like this: "... if the ebuild can use &lt;c&gt;flag&lt;/c&gt; in &lt;c&gt;use&lt;/c&gt; queries, ..."

ulm (approved):

ulm:
Thanks, merged.

PR #227, opened 2021-05-05
src_test: Clarify how localhost can be used

mgorny:

### DIFF
--- a/ebuild-writing/functions/src_test/text.xml
+++ b/ebuild-writing/functions/src_test/text.xml
@@ -172,7 +172,7 @@ a more complete test suite.
 <p>
 It is generally considered acceptable to rely on IPv4 <c>localhost</c>
 being resolvable and available for binding to and connecting to
-the previously bound services.  However, it is not acceptable to connect
+services started within the phase. However, it is not acceptable to connect
 to the daemons run outside the test environment.
 </p>
 
--- a/ebuild-writing/functions/src_test/text.xml
+++ b/ebuild-writing/functions/src_test/text.xml
@@ -170,8 +170,15 @@ a more complete test suite.
 </p>
 
 <p>
-Local server access within the build procedure is additionally
-forbidden for the following reasons:
+It is generally considered acceptable to rely on IPv4 <c>localhost</c>
+being resolvable and available for binding to and connecting to
+the previously bound services.  However, it is not acceptable to connect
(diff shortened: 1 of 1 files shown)


ebuild-writing/functions/src_test/text.xml
```diff
@@ -170,8 +170,15 @@ a more complete test suite.
 </p>
 
 <p>
-Local server access within the build procedure is additionally
-forbidden for the following reasons:
+It is generally considered acceptable to rely on IPv4 <c>localhost</c>
+being resolvable and available for binding to and connecting to
+the previously bound services.  However, it is not acceptable to connect
+to the daemons run outside the test environment.
```
ulm, line 176: I don't understand what you mean by "previously bound" in the first sentence.

Also, not sure if the second sentence is correct grammatically (but I'm not a native speaker). I think it should either read "daemons running" or "daemons that run".

mgorny: I was trying to say that it can connect to whatever it started but not anything else.

ebuild-writing/functions/src_test/text.xml
```diff
@@ -170,8 +170,15 @@ a more complete test suite.
 </p>
 
 <p>
-Local server access within the build procedure is additionally
-forbidden for the following reasons:
+It is generally considered acceptable to rely on IPv4 <c>localhost</c>
+being resolvable and available for binding to and connecting to
+the previously bound services.  However, it is not acceptable to connect
```
thesamesam, line 175:
```suggestion
services started within the phase. However, it is not acceptable to connect
```

thesamesam (changes requested):

ebuild-writing/functions/src_test/text.xml
Flowdalic: FWIW I understood the "previously bound", but the first sentence is still very complex. How about splitting it up? For example:

It is generally considered acceptable to rely on IPv4 localhost being resolvable and available for binding. Tests should only connect to services that are started as part of the testsuite. It is not acceptable to connect to daemons run outside the text environment.

ebuild-writing/functions/src_test/text.xml
ulm: With this change, PR looks good to me.

PR #228, opened 2021-05-28
ebuild-writing/variables: 'live' is also a valid PROPERTIES' value

Flowdalic:
Signed-off-by: Florian Schmaus <flo@geekplace.eu>

### DIFF
--- a/ebuild-writing/variables/text.xml
+++ b/ebuild-writing/variables/text.xml
@@ -276,7 +276,7 @@ The following variables may or must be defined by every ebuild.
     <ti><c>PROPERTIES</c></ti>
     <ti>
     A space-delimited list of properties, with conditional syntax support.
-    <c>interactive</c> is the only valid value for now.
+    <c>interactive</c>, <c>live</c>, and <c>test_network</c> are valid values.
     </ti>
   </tr>
   <tr>


ebuild-writing/variables/text.xml
```diff
@@ -276,7 +276,7 @@ The following variables may or must be defined by every ebuild.
     <ti><c>PROPERTIES</c></ti>
     <ti>
     A space-delimited list of properties, with conditional syntax support.
-    <c>interactive</c> is the only valid value for now.
+    <c>interactive</c> and <c>live</c> are valid values.
```
ulm, line 279: Maybe  add also `test_network` while at it?
https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-670007.3.5

Flowdalic: Sure, but I can't find `test_network` anywhere in the provided link?

ulm: https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-670007.3.5

#### 7.3.5 Properties

The following tokens are permitted inside `PROPERTIES`:
<dl>
<dt>interactive</dt><dd>The package may require interaction with the user via the tty.</dd> 
<dt>live</dt><dd>The package uses “live” source code that may vary each time that the package is installed.</dd>
<dt>test_network</dt><dd>The package manager may run tests that require an internet connection, even if the ebuild has <code>RESTRICT=test</code>.</dd>
</dl>
Package managers may recognise other tokens. Ebuilds may not rely upon any token being supported.

ulm (approved):

mgorny (approved):
Would be nice to have a more detailed description somewhere but that's irrelated to this change.

ulm:
Thank you.

PR #229, opened 2021-05-30
tools-reference/diff-and-patch/: add a link to "clean patches" page

juippis:
- add a text link from tools-reference/diff-and-patch/index.html to
   ebuild-writing/misc-files/patches


thesamesam (approved):

ulm:
Thanks.

PR #231, opened 2021-06-23
ebuild-writing/functions/src_test: explicitly USE=test is only for so…

thesamesam:
…me cases

It's only needed where we have test dependencies, we want to conditionally
build tests, conditionally fetch files, or otherwise alter behaviour based
on the fact we're running tests (e.g. usage in RESTRICT).

Suggested-by: Florian Schmaus <flow@gentoo.org>


ebuild-writing/functions/src_test/text.xml
```diff
@@ -76,6 +76,14 @@ section on <uri link="::general-concepts/dependencies/#Test Dependencies"/>
 for more information.
 </p>
 
+<p>
+Note that the <c>test</c> USE flag is only necessary if there are
+test dependencies, tests are being built conditionally, or we otherwise want
```
sbraz, line 81:
```suggestion
test dependencies, tests are being built conditionally, or it is desirable
```

thesamesam: Done, thank you!

PR #232, opened 2021-06-23
general-concepts/ebuild-revisions: update revision bump examples

thesamesam:
- A revision bump is needed for a rebuild to occur when e.g. ABI is broken.

Without doing this, we're relying on Portage's preserve-libs feature which
does not have to be enabled and is not specified by PMS.

- Without revision bumping in cases where we change USE dep defaults in a
dependency, the package manager will refuse an upgrade because newer
versions of the dependency will not satisfy the requirements
in the package.

### DIFF
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -67,6 +67,7 @@ of thumb could be used as a guideline:
   <li>adding a patch to fix a runtime issue,</li>
   <li>installing additional files that could be useful to the user,</li>
   <li>adding a missing runtime dependency to one of the existing blocks,</li>
+  <li>adding a binding subslot operator (:=) to a dependency,</li>
   <li>
     adding a missing build-time dependency that contributed to
     a successfully yet incorrect build (e.g. missing some feature),
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -68,6 +68,7 @@ of thumb could be used as a guideline:
   <li>installing additional files that could be useful to the user,</li>
   <li>adding a missing runtime dependency to one of the existing blocks,</li>
   <li>adding a binding subslot operator (:=) to a dependency,</li>
+  <li>updating a dependency with default on/off USE flags,</li>
   <li>
     adding a missing build-time dependency that contributed to
     a successfully yet incorrect build (e.g. missing some feature),


Flowdalic:
I suggest also including a "add missing test dependency" to the examples, which, according to @juippis  in #gentoo-qa does not require a revbump (which would also been my estimation).

orlitzky:
> I suggest also including a "add missing test dependency" to the examples, which, according to @juippis in #gentoo-qa does not require a revbump (which would also been my estimation).

This can cause reinstalls to fail because the new dependency won't be pulled in but the test suite will still be run if the user enables `FEATURES=test` after the initial install succeeds without it (or if the missing dependency was installed "accidentally" at that time, and later depcleaned). If this is the _only_ test dependency and if `USE=test` is new, then `--changed-use` can catch that... but `--changed-use` is a portage-only feature (i.e. not part of the PMS).

The existing examples "adding a missing build-time dependency that caused a build failure" and "adding a new USE flag or removing an existing one..." are wrong for the same reasons.

juippis:
> 
> This can cause reinstalls to fail 

Use `FEATURES="test-fail-continue"`


> because the new dependency won't be pulled in but the test suite will still be run if the user enables `FEATURES=test` 

No one does that. 

> then `--changed-use` can catch that... but `--changed-use` is a portage-only feature (i.e. not part of the PMS).

*man emerge* tells me that:
```
       --changed-use, -U
              Tells emerge to include installed packages where USE flags have changed since  installation.  This  option
              also  implies the --selective option. Unlike --newuse, the --changed-use option does not trigger reinstal‐
              lation when flags that the user has not enabled are added or removed.

              NOTE: This option ignores the state of the "test" USE flag, since that flag has a special binding to  FEA‐
              TURES="test" (see make.conf(5) for more information about FEATURES settings).
```

> The existing examples "adding a missing build-time dependency that caused a build failure" and "adding a new USE flag or removing an existing one..." are wrong for the same reasons.

pkgcore is perfectly capable of doing --newuse.

Flowdalic:
> This can cause reinstalls to fail because the new dependency won't be pulled in but the test suite will still be run if the user enables `FEATURES=test` after the initial install succeeds without it (or if the missing dependency was installed "accidentally" at that time, and later depcleaned).

Fair point. Just to get the complete picture: This wouldn't be the case if portage is run with dyanmic deps (`--dyanmic-deps`) enabled, correct?

That said, even with static dependencies (which are what the PMS specifies IIRC) I am not sure if "adding a missing test dependency" justifies a revbump, given that many tests, even if we don't want it, are inherently nondeterministic. So if the test fails, uses will probably simply flip the feature off again for the package (and least that's what would do) or enable test-fail-continue (as suggested by @juppis). Furthermore, the revbump will cause rebuilds for users which have tests disabled, which are arguably the majority of our users, simply because this is the default.

orlitzky:
> > This can cause reinstalls to fail
> 
> Use `FEATURES="test-fail-continue"`
> 

If I'm still going to investigate the test failure, it's going to be a huge waste of time for me to figure out that a dependency is missing and that's it's in the ebuild now but wasn't when I installed it because the developer added it to the ebuild without a revbump. If I'm not going to investigate the test failure... why would I enable tests?


>  > The existing examples "adding a missing build-time dependency that caused a build failure" and "adding a new USE flag or removing an existing one..." are wrong for the same reasons.
> 
> pkgcore is perfectly capable of doing --newuse.

`--newuse` is also not in the PMS, but I'm not sure what it has to do with these examples anyway.  The build-dependencies example doesn't involve USE flags at all. The other explicitly says `--changed-use`.

orlitzky:
> > This can cause reinstalls to fail because the new dependency won't be pulled in but the test suite will still be run if the user enables `FEATURES=test` after the initial install succeeds without it (or if the missing dependency was installed "accidentally" at that time, and later depcleaned).
> 
> Fair point. Just to get the complete picture: This wouldn't be the case if portage is run with dyanmic deps (`--dyanmic-deps`) enabled, correct?

Right.


> That said, even with static dependencies (which are what the PMS specifies IIRC) I am not sure if "adding a missing test dependency" justifies a revbump, given that many tests, even if we don't want it, are inherently nondeterministic. So if the test fails, uses will probably simply flip the feature off again for the package (and least that's what would do) or enable test-fail-continue (as suggested by @Juppis). Furthermore, the revbump will cause rebuilds for users which have tests disabled, which are arguably the majority of our users, simply because this is the default.

There are exceptions, but that's what they are: exceptions. In general, you have to do a revbump to ensure that the package manager can figure out what to do. That's what the PMS says, and what the council said. Writing down the exception as if it's the rule is not just dishonest but actually harmful.

Rebuilding LLVM or Firefox for a new test dependency would be sadistic. But for most packages, people are doing `emerge -uDN @world` and walking away. Adding a few minutes to that for "pointless" rebuilds does no harm, and it avoids what everyone hates the most: having portage crash, and having to spend two days figuring out WTF went wrong. These are obscure problems that only occur when e.g. a subslot rebuild is taking place AND tests are enabled AND the developer messed up. How long do you think it's going to take a regular user to figure out why he can't update his machine? You wind up wasting a massive amount of real-people time to avoid a tiny bit of CPU time.

People spend way more time thinking up excuses for not doing the revbump than it would take them to just do the revbump and be safe.

juippis:
> If I'm still going to investigate the test failure, it's going to be a huge waste of time for me to figure out that a dependency is missing and that's it's in the ebuild now but wasn't when I installed it because the developer added it to the ebuild without a revbump. 

You are the 0.00000001 %. You re-merge the package, no one else has to suffer.

> If I'm not going to investigate the test failure... why would I enable tests?

Yeah, why would you enable it globally? Sounds silly.
I also can't figure out what you're trying to say, because:

> If I'm still going to investigate the test failure
> If I'm not going to investigate the test failure

you seem to answer yourself here?

> The build-dependencies example doesn't involve USE flags at all. 

It's most likely talking about BDEPEND. Since `RDEPEND="${DEPEND}"` warrants a revbump in most cases anyway. Besides, at least portage guards this with `@preserved-rebuild`. No idea if pkgcore has something similar.

> Adding a few minutes to that for "pointless" rebuilds does no harm, and it avoids what everyone hates the most: having portage crash, and having to spend two days figuring out WTF went wrong. These are obscure problems that only occur when e.g. a subslot rebuild is taking place AND tests are enabled AND the developer messed up. How long do you think it's going to take a regular user to figure out why he can't update his machine? You wind up wasting a massive amount of real-people time to avoid a tiny bit of CPU time.

This example makes no sense. Again, what about the 10 000 potential other users? No one runs their system with `FEATURES="test"` enabled globally. And if you do, you're most likely already prepared to do some extra work for it. I'm sure most of test failures do not happen because of missing deps anyway. So again, everyone has to rebuild because of some niche low-chance scenario for people who deliberately sabotage their own systems. Makes no sense.

> People spend way more time thinking up excuses for not doing the revbump than it would take them to just do the revbump and be safe.

We just like to apply some logic and common sense to our actions.

orlitzky:
> > If I'm still going to investigate the test failure, it's going to be a huge waste of time for me to figure out that a dependency is missing and that's it's in the ebuild now but wasn't when I installed it because the developer added it to the ebuild without a revbump.
> 
> You are the 0.00000001 %. You re-merge the package, no one else has to suffer.
> 

Re-merging the package doesn't change anything. That's the whole point of this discussion. The old version with the old (incorrect) dependencies can be merged, leading to another failure.


> > If I'm not going to investigate the test failure... why would I enable tests?
> 
> Yeah, why would you enable it globally? Sounds silly.
> I also can't figure out what you're trying to say, because:
> 
> > If I'm still going to investigate the test failure
> > If I'm not going to investigate the test failure
> 
> you seem to answer yourself here?

All I've done is give you two (exhaustive) cases. One case (enabling tests and then ignoring them completely) is nonsense. And in the other, your `FEATURES=test-fail-continue` does nothing to help, because I still have to reverse engineer what the developer fucked up.

> 
> > The build-dependencies example doesn't involve USE flags at all.
> 
> It's most likely talking about BDEPEND. Since `RDEPEND="${DEPEND}"` warrants a revbump in most cases anyway. Besides, at least portage guards this with `@preserved-rebuild`. No idea if pkgcore has something similar.
> 

Wouldn't it be great if we had a document that said what package managers were guaranteed to implement, and some kind of governing body to encourage people to develop against that document instead of whatever shit they just make up?

juippis:
> Re-merging the package doesn't change anything. That's the whole point of this discussion. The old version with the old (incorrect) dependencies can be merged, leading to another failure.
> 

If I fix it without revbumping and you re-merge it, of course you get the updated dep list? 

> 
> Wouldn't it be great if we had a document that said what package managers were guaranteed to implement, and some kind of governing body to encourage people to develop against that document instead of whatever shit they just make up?

It's great though we have this governing guide for developer workflow so we don't piss off majority of users over this :)

orlitzky:
> > Re-merging the package doesn't change anything. That's the whole point of this discussion. The old version with the old (incorrect) dependencies can be merged, leading to another failure.
> 
> If I fix it without revbumping and you re-merge it, of course you get the updated dep list?
> 

No.

juippis:
> > 
> > If I fix it without revbumping and you re-merge it, of course you get the updated dep list?
> 
> No.

I just tried it with emerge and pmerge, and it works for me. I don't know what you're doing wrong.

Steps: 
Add 
```
src_test() {
	echo "made it this far, die now" && die
}
```
to a package. emerge with 
`FEATURES="test test-fail-continue USE="test"`.

test phase will fail, but emerge will be successful.
Now update the ebuild, taking that `src_test` out and adding more test deps. When doing `FEATURES="test" USE="test" pmerge -av mypackage` or `FEATURES="test" USE="test" emerge -av mypackage` it wants to install the added deps first.

Flowdalic:
> Now update the ebuild, taking that `src_test` out and adding more test deps. When doing `FEATURES="test" USE="test" pmerge -av mypackage` or `FEATURES="test" USE="test" emerge -av mypackage` it wants to install the added deps first.

Is this also the case if you explicitly disable the dynamic deps feature of portage? That is

`FEATURES="test" USE="test" emerge -av --dynamic-deps n mypackage`

juippis:
I do not have dynamic deps enabled.

orlitzky:
> > > If I fix it without revbumping and you re-merge it, of course you get the updated dep list?
> > 
> > 
> > No.
> 
> I just tried it with emerge and pmerge, and it works for me. I don't know what you're doing wrong.
> 

I'm not doing anything wrong. You're throwing out anecdotes, whereas I've been arguing from the beginning that we shouldn't document behaviours that can break a package manager that follows the PMS.

All kinds of things can and do go wrong when you omit the revisions. For some real life examples, look back at all of the bug reports that triggered the re-enabling of dynamic-deps in portage. Regardless, trying to enumerate all of the ways that the two current versions of the two current package managers can (mis)handle things is a complete waste of time. There's a spec to follow. The easy and simple thing to do is to quit thinking, honor it, and do the revision bump unless you have a really good reason not to.

ulm (approved):
PR looks good to me.

juippis:
> The easy and simple thing to do is to quit thinking, honor it, and do the revision bump unless you have a really good reason not to.

I think not revbumping for added test deps is a really good reason not to. While it may be conceptually the right thing to do, in real life, applying the practice wouldn't help **anyone**.

ulm:
I have only skimmed that long discussion, but the package manager will use the dependencies from the ebuild when building a package. Therefore a revbump isn't needed when updating `DEPEND` or `BDEPEND`.

Think of it as a two-step process, building a binary image and then merging it to the system. The first step doesn't even require the package to be installed at all (think of building a binpkg), so the only consistent way to operate is to take dependencies from the ebuild, not from the VDB (where the package may or may not exist).

PR #235, opened 2021-07-11
ebuild-maintenance/general: new page

juippis:
@thesamesam is this how you envisioned the page to be?

Not sure if this belongs to devmanual, but as it was proposed once, here's an updated version.

Think I really shot myself in a foot with these `<subsection>s` and `<body>ies`, but oh well, seems to pass the xml validator...

Originally I had multiple "logical" commits updating the page, but it became a mess so just rebased
everything to a single update commit.

I can see there are a lot of trailing whitespaces, but can fix while fixing other requested changes.


juippis:
^ I'll fix those too after review rounds.

ulm:
Thank you. The criticism from PR #193 still applies. While I don't see major problems with the contents, from a point of view of document structure it doesn't make sense to reintroduce another "general" section with subsections that are only loosely related to each other.

We had a catch-all "ebuild maintenance" chapter in the past (which was added late by collecting different topics from the old devrel handbook), and we eventually split it into more specific sections. So please let's not repeat that mistake.

Also, please make the patch validate and conform to the [style guide](https://devmanual.gentoo.org/appendices/devbook-guide/index.html), so that we need not address these trivial issues during review.

juippis:
> Thank you. The criticism from PR #193 still applies. While I don't see major problems with the contents, from a point of view of document structure it doesn't make sense to reintroduce another "general" section with subsections that are only loosely related to each other.
> 
> We had a catch-all "ebuild maintenance" chapter in the past (which was added late by collecting different topics from the old devrel handbook), and we eventually split it into more specific sections. So please let's not repeat that mistake.

I admit I've become a bit blind to the structure already. How do you suggest changing it? Separate into own subpages, remove subsections and subtitles and just write everything as separate paragraphs..?

> 
> Also, please make the patch validate and conform to the [style guide](https://devmanual.gentoo.org/appendices/devbook-guide/index.html), so that we need not address these trivial issues during review.

I did try to follow this guide. Are you talking about the errors found by semaphoreci, or is there more?

ulm:
> I admit I've become a bit blind to the structure already. How do you suggest changing it? Separate into own subpages, remove subsections and subtitles and just write everything as separate paragraphs..?

Arguably the parent chapter's title "Ebuild Maintenance" is already a little problematic, because most of the Devmanual is about ebuild maintenance. So let's at least avoid anything with an unspecific name like "General", "Miscellaneous", or "Etcetera" below it. (In fact, in the past we also had "Ebuild Maintenance / Maintenance Tasks" which falls in the same category, so it was split.)

Maybe some subsections would fit better elsewhere in the devmanual (I haven't checked in detail, though)?

> I did try to follow this guide. Are you talking about the errors found by semaphoreci, or is there more?

On first glance:
- Validation fails
- "Opening tags with a single attribute may not be split between lines. For example, don't put a newline between `<uri` and its link attribute. Break the line before the `<uri>` tag instead."
- Most items in the list at the end are short and should be in a single line. (Then again, it's just a collection of external links, which aren't referenced from the text? We try to avoid such lists, or at least keep the number of external references at a minimum.)

juippis:
> Arguably the parent chapter's title "Ebuild Maintenance" is already a little problematic, because most of the Devmanual is about ebuild maintenance. So let's at least avoid anything with an unspecific name like "General", "Miscellaneous", or "Etcetera" below it. (In fact, in the past we also had "Ebuild Maintenance / Maintenance Tasks" which falls in the same category, so it was split.)

I'll try to think of something better.

> 
> Maybe some subsections would fit better elsewhere in the devmanual (I haven't checked in detail, though)?

I kind of agree, but that's also why this page has many links to external pages - to avoid repetition and replication.

> 
> > I did try to follow this guide. Are you talking about the errors found by semaphoreci, or is there more?
> 
> On first glance:
> 
>     * Validation fails
> 
>     * "Opening tags with a single attribute may not be split between lines. For example, don't put a newline between `<uri` and its link attribute. Break the line before the `<uri>` tag instead."

Cheers, on it next!

> 
>     * Most items in the list at the end are short and should be in a single line. (Then again, it's just a collection of external links, which aren't referenced from the text? We try to avoid such lists, or at least keep the number of external references at a minimum.)

Every link in the last section, "External resources", (should) have been mentioned in this page. I kind of wanted to provide a quick access.

sbraz (changes requested):

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,474 @@
...
+<body>
+
+<p>
+You may find yourself in a lucky position where enthusiastic users provide 
+contributions to your packages via Github pull requests. Using Github is not
+mandatory and Bugzilla is still the main channel for reporting enhancements to
+your packages. But consider trying out Github, as it's getting more and more
+popular and we're seeing more and more contributions pushed there. Since a
+distribution would be nothing without its users, we shouldn't ignore our
+users in popular platforms. Chances are, they end up becoming official
```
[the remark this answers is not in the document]
juippis: This is like saying "They might maybe" - I feel like it should be either or, not both. "They might end up" or "Chances are".

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,474 @@
...
+and finally, to <b>combine</b> these checks, you can do:
+<c>git grep -l developer@gentoo.org '**/metadata.xml' | cut -d/ -f1-2 | xargs
+pkgcheck --color true scan -c RedundantVersionCheck,StableRequestCheck</c>
+
+These two above checks are also shown in your maintainer page on <uri
+link="https://packages.gentoo.org/">packages.gentoo.org</uri>.
+
+Finally check available <e>PYTHON_COMPAT</e> updates to your packages with:
+<c>git grep -l developer@gentoo.org '**/metadata.xml' | cut -d/ -f1-2 | xargs
+pkgcheck --color true scan -k PythonCompatUpdate</c>
```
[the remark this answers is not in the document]
juippis: Well if you follow these steps and do stabilization and clean older versions occassionally, that shouldn't be a problem ;)
But yes, I get what you mean. Gonna think something for this.

juippis:
- fixed the xml errors,
- fixed 95 % of sbraz's suggestions

TODO:
 - I'm aware the `<uri link="::self/#` don't work, need to find an example and fix them
 - add more subsections to Github section, since it's a huge block of text and hard to follow
 - "break sync-tree status" needs rewording explaining if you break the tree with a new CI error, new metadata won't be generated to our sync repos and therefore users won't get the latest updates before CI error is fixed
 - fix the "Daily / Occassional pkgcheck scans" formatting, since `<br/>` doesn't work here :(

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,507 @@
...
+
+<p>
+This page aids you in getting familiar with the "everyday" tasks and
+responsibilites of a package maintainer in Gentoo. You may find this page
+useful if you're curious about stepping up as a maintainer, returning from a
+long hiatus, or simply looking for tips.
+
+Not everything listed in here need to be incorporated to your workflow, but
+following these steps should give users the best experience, and help your
+life as a maintainer.
```
juippis, line 18:
```suggestion
following these steps should keep the Gentoo repository in a clean working state, keep up your activity and ease your tasks as a maintainer. 
```

ebuild-maintenance/general/text.xml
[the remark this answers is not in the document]
sbraz: In that case, maybe "chances are they will end up"? Even if you want to keep the existing verb, the comma between the subject and the verb should be removed.

juippis:
That should be all done, happy with the outcome. 

If we're talking about re-structuring it, could perhaps the move the bugzilla and github sections out into their own subpages and go more into detail with each step.

ebuild-maintenance/general/text.xml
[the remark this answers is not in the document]
juippis: Removed the comma, and fixed all trailing whitespaces.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+
+<p>
+This page aids you in getting familiar with the "everyday" tasks and
+responsibilites of a package maintainer in Gentoo. You may find this page
+useful if you're curious about stepping up as a maintainer, returning from a
+long hiatus, or simply looking for tips.
+
+Not everything listed in here need to be incorporated to your workflow, but
+following these steps should keep the Gentoo repository in a clean working
+state, keep your activity up and ease your tasks as a maintainer.
```
thesamesam, line 18: Nit: Oxford comma.

```suggestion
state, keep your activity up, and ease your tasks as a maintainer.
```

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+</p>
+
+<p>
+Your general responsibilities as a maintainer consist of keeping your packages
```
thesamesam, line 22: Nit: maybe this and semicolons between the items ("upstream releases; ...")?
```suggestion
Your general responsibilities as a maintainer consist of: keeping your packages
```

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+
+<subsection>
+<title>Bugzilla</title>
+<body>
+
+<p>
+<uri link="https://bugs.gentoo.org/">Gentoo's Bugzilla</uri> is still the main
+communication channel to reach you with your package state. When a bug report
+gets assigned to you, the first thing you should attempt to do is to confirm
+and replicate it with the information attached to the bug. You should then
```
thesamesam, line 62: The [style](https://devmanual.gentoo.org/appendices/contributing/index.html#style-guidelines) guidelines mention third person. I personally find this a bit weird because the technical writing I'm used to allows gratuitous use of 'we', but...

So, we need to change instances of 'you' like this.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+
+<p>
+Keep an eye out for CI/tinderbox-style bugs too, which are found by testing the
+latest ebuilds (by commit) and random packages, respectively. They reveal
+breakages on different systems, mostly missing dependencies, and make you face
+issues you perhaps hadn't thought of before.
+</p>
+
+<p>
+When you make a commit to Gentoo's git tree, you can reference or close a
```
thesamesam, line 74:
```suggestion
When you make a commit to Gentoo's git tree, please remember to reference or close a
```

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+<p>
+<uri link="::keywording/#Stabilization rules">Stabilization can be
+done by a maintainer who is also a developer, and owns relevant hardware</uri>.
+It can also be <uri link="::keywording/#Moving from ~arch to arch">
+requested through Bugzilla</uri> for arch teams to handle, even for your own
+packages.
+</p>
+
+<p>
+It is recommended to set up your mail client's filtering properly. Separating
```
thesamesam, line 90: Maybe we should link to [this](https://wiki.gentoo.org/wiki/User:Wraeth/Sieve) or [this](https://wiki.gentoo.org/wiki/User:Wraeth/Sieve)?

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+
+<p>
+When writing and pushing ebuilds to Gentoo's repository tree (and why not to
+any overlay as well) you should definitely take advantage of available QA
+tools. <c>pkgcheck</c> and <c>repoman</c> being your choices here. They can
+both be found and installed from the main Gentoo repository with their
+respective names.
+</p>
+
+<!-- author's note to reviewers: I find the PMS link correct here, but
```
thesamesam, line 157: I'd link to the identifier from a specific EAPI just in case, so use 7 from now.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+<note>
+EAPI availability is usually dictated by the state of eclasses being used.
+</note>
+
+<p>
+After you've edited an ebuild or metadata.xml file, run <c>pkgcheck scan</c>
+and/or <c>repoman -x full</c> in the package directory.
+</p>
+
+<p>
```
thesamesam, line 178: Nice.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+<uri link="::ebuild-writing/common-mistakes"/> when writing ebuilds. In
+general, familiarize yourself with the Quality Assurance's
+<uri link="https://projects.gentoo.org/qa/policy-guide/">Policy guide</uri> and
+if writing Python ebuilds, the Python project's
+<uri link="https://dev.gentoo.org/~mgorny/python-guide/">Python guide</uri>.
+</p>
+
+<p>
+When committing your changes, <c>repoman commit</c> or <c>pkgdev commit</c>
+(from "pkgdev" package) can be used. They will properly sign-off your commit
```
thesamesam, line 189:
```suggestion
(from the <c>dev-util/pkgdev</c> package) can be used. They will properly sign-off your commit
```

```suggestion
(from "pkgdev" package) can be used. They will properly sign-off the commit
```

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+with accordance to Gentoo's
+<uri link="https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin">
+GLEP 76 requirements</uri>. These tools will also append the correct, GLEP 66
+compliant <uri link="::ebuild-maintenance/git/#Git commit message format">
+summary</uri> and fix Manifest or wrong copyright headers automatically.
+
+Commits also need to be GPG-signed, so keep your keys
```
thesamesam, line 196: Nit:
```suggestion
Commits also need to be PGP-signed, so keep your keys
```

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+<uri link="https://wiki.gentoo.org/wiki/Project:Infrastructure/Generating_GLEP_63_based_OpenPGP_keys#Invalid_expiration_date">
+up-to-date</uri> and
+<uri link="https://wiki.gentoo.org/wiki/Project:Infrastructure/Generating_GLEP_63_based_OpenPGP_keys#Submit_your_new_key_to_the_keyserver">
+available</uri> in Gentoo's infra.
+</p>
+
+<p>
+After committing, you can use <c>pkgcheck scan --commits</c> for a final check
```
thesamesam, line 204:
```suggestion
After committing, use <c>pkgcheck scan --commits</c> for a final check
```

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+</p>
+
+<p>
+<b>Joining the Gentoo GitHub organization:</b> To join the Gentoo GitHub
+organization, make sure you've set your <c>gentooGitHubUser</c> attribute in
+<uri link="https://wiki.gentoo.org/wiki/Project:Infrastructure/LDAP_Guide#Optional_custom_attributes">
+dev.gentoo.org's LDAP record</uri>. Join <c>#gentoo-infra</c> on the Libera IRC
+network to ask for a manual re-sync, and you will be added to the Gentoo
+organization in GitHub. You should be automatically added to teams you're part
+of, but you can request to join manually too. Remember to once again update
```
thesamesam, line 280: You can but AFAIK (I tried before) we should just poke infra to run the sync script instead, or it'll get lost if for some reason it's not recorded properly on the wiki:

```suggestion
of. Remember to once again update
```

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+should, of course ask before pushing if something is not fully clear to you.
+</p>
+
+</body>
+
+<subsubsection>
+<title>GitHub pull requests</title>
+<body>
+
+<p>
```
thesamesam, line 262: Nice.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,527 @@
...
+You may find yourself in a lucky position where enthusiastic users provide
+contributions to your packages via GitHub pull requests. Using GitHub is not
+mandatory and Bugzilla is still the main channel for reporting enhancements to
+your packages. But consider trying out GitHub, as it's getting more and more
+popular and we're seeing more and more contributions pushed there. Since a
+distribution would be nothing without its users, we shouldn't ignore our
+users in popular platforms. Chances are they end up becoming official
```
thesamesam, line 269:
```suggestion
users in popular platforms. Chances are, they end up becoming official
```

thesamesam (approved):
This is **fantastic** work and exactly what I envisioned. Huge, huge thanks for working on this.

I like the idea of a list of obligations and very common tasks, as opposed to burying some of this within reference-style material.

_If_ we do need to split this up (now or later), I will happily help you do that work, but I think this is pretty good as-is. If it ends up becoming a frankenpage and certain sections seem ripe for splitting out, we'll do that?

ebuild-maintenance/general/text.xml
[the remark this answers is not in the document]
radhermit: Filtering is somewhat supported internally so I assume you want something like ``pkgcheck scan -f latest -k PythonCompatUpdate``, but iirc that only works at category/repo scope atm and I think there's a open issue about it for generalizing to local pkg scope.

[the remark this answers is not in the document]
radhermit: Also just to note, ``pkgcheck scan`` doesn't require using xargs. It detects and handles reading from stdin by itself if you pipe it targets it understands and can generate restrictions from.

ebuild-maintenance/general/text.xml
juippis: Yeah, I tried to avoid using "we", "our" and "I" due to this. But I feel like the text can still be addressed to **you**. And there are *several* devmanual pages that read to "you".

ebuild-maintenance/general/text.xml
juippis: You gave the same link twice - but I guess https://wiki.gentoo.org/wiki/Project:Infrastructure/Developer_E-Mail/Sieve_Example would be the 2nd one? Yeah, a link to a **good** mail filter setup would be great.

ebuild-maintenance/general/text.xml
juippis: That did cross my mind but I thought if at some point more text was added to it, clarifications, updates etc, they don't make it to older versions. But this is a good idea, I'll do that.

juippis:
Sam's pointed out notes done.

ebuild-maintenance/general/text.xml
thesamesam: Sorry, yes!

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,530 @@
...
+<body>
+
+<p>
+You may find yourself in a lucky position where enthusiastic users provide
+contributions to your packages via GitHub pull requests. Using GitHub is not
+mandatory and Bugzilla is still the main channel for reporting enhancements to
+your packages. But consider trying out GitHub, as it's getting more and more
+popular and we're seeing more and more contributions pushed there. Since a
+distribution would be nothing without its users, we shouldn't ignore our
+users in popular platforms. Chances are, they end up becoming official
```
ulm, line 272: I see this section very critical. As a free software distro, we must not advertise a proprietary and [arguably unethical](https://www.gnu.org/software/repo-criteria-evaluation.html) Microsoft platform in our official documention.

Mentioning Github is o.k., but phrases like "you may find yourself in a lucky position" or "consider trying out GitHub, as it's getting more and more popular" definitely don't belong here. Indirectly, this would also put pressure on developers on using Github, while we have clarified multiple times (and our Social Contract mandates) that its use in entirely optional.

ebuild-maintenance/general/text.xml
```diff
@@ -0,0 +1,530 @@
...
+  <li>
+    <uri link="https://mgorny.pl/articles/the-ultimate-guide-to-eapi-7.html">
+      Ultimate guide to EAPI-7
+    </uri>
+  </li>
+  <li>
+    <uri link="https://mgorny.pl/articles/the-ultimate-guide-to-eapi-8.html">
+      Ultimate guide to EAPI-8
+    </uri>
+  </li>
```
ulm, line 521: The relevant parts of that guide are now in the EAPI 8 section (and there you'll find the link as well), so no need to link it from here as well.

In general, I see this whole large collection of links somewhat critical. It make the chapter look like a web page from the nineties. :grimacing:

ulm (changes requested):

ebuild-maintenance/general/text.xml
juippis:
```
You may find yourself in a position where users provide contributions to your 
packages via GitHub pull requests. Gentoo is seeing more and more contributions 
pushed there, as GitHub is getting more popular. That being said, using GitHub 
is not mandatory and Bugzilla is still the main channel for reporting 
enhancements to your packages.
```

ebuild-maintenance/general/text.xml
juippis: Just the way I like it ;)

sure I'll remove the block.

ebuild-maintenance/general/text.xml
ulm: WFM, if you word the second sentence in a more neutral way. Something along the lines of:
"As GitHub is a popular platform, Gentoo is seeing many contributions pushed there."

ebuild-maintenance/general/text.xml
juippis: Should be all done now.

ulm (changes requested):
As I said twice before, this doesn't fit well in the existing structure, so it needs to be split up.

juippis:
> As I said twice before, this doesn't fit well in the existing structure, so it needs to be split up.

I saw you wonder about it, but the request to split it wasn't clear to me.

Now I can imagine the whole "ebuilds & git workflow" section, including "Github pull requests" can be placed under *ebuild-maintenance/git*. From "External tools", I can see all pkgcheck calls be put under a new page *tools-reference/pkgcheck* digging deeper into pkgcheck's possibilities. Regardless of the outcome of this PR, such pkgcheck page would most likely be a welcomed addition. 

Now "Your responsibilities", "bugzilla", "packages.gentoo.org" and "qa-reports" sections are harder. BUT! Splitting this up potentially nullifies the whole purpose I proposed this update: A simple checklist to answer the question "What can I do? Are my packages in good condition?" 
I don't want **this** task to be overly complicated. I don't want it to be a page that takes you to multiple different places before you **find** all relevant checks / information. 

You said you've had similar page in devmanual before, and it didn't work. So I'm thinking maybe wiki would be a better place for this page after all. However I wanted it to be behind some "authoritative" media, and not be editable by *everyone*. 

If you can think of a way to make this work while splitting up the content, would you be willing to do it with the material provided in here? Since if I split it up, I'm only guessing what goes where and I fear it's a timesink for everyone involved. And if you think it can't work, let me know, let's not waste anyone's time here. I'll find another more suitable place for this.

ulm:
Hm, seems like I haven't explicitly repeated here that it should be split. I said so in the predecessor PR #193.

So yes, splitting off git workflow and pkgcheck would be a good start.

As for the rest, what I am most unhappy with in this PR is the `general` name, and the "Maintainer's checklist" heading which is still rather unspecific. Since we already have sections on "adding" and "removing" ebuilds and packages, maybe this could go under a more specific heading like "Updating a Package" or "Maintaining an Existing Package"? I'd also have no objections against splitting it up further, i.e. having two or three chapters directly under `ebuild-maintenance`, if you think that would make sense.

juippis:
So I've thought about it, and I think I'm gonna make a wiki page for now. I really want everything general-maintainer-related-tasks to be found in a single page. 

There are parts in this PR that should be applied to devmanual though, and I'll return with those, along with other new suggestions, after a while.

PR #236, opened 2021-07-15
profiles/{package, use}.mask: mention sorting mask entries chronologically

thesamesam:

### DIFF
--- a/profiles/package.mask/text.xml
+++ b/profiles/package.mask/text.xml
@@ -15,6 +15,12 @@ of the <c>package.mask</c> file is described in <c>man portage</c>.
 </p>
 
 <p>
+Entries are added chronologically <d/> that is, newer entries
+should be placed towards the top of the file, underneath any initial
+header comment block.
+</p>
+
+<p>
 This file can be used in subprofiles to mask packages only for certain setups.
 </p>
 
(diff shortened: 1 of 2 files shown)


profiles/package.mask/text.xml
```diff
@@ -14,6 +14,12 @@ have a comment detailing the specific reason for the mask. The format
 of the <c>package.mask</c> file is described in <c>man portage</c>.
 </p>
 
+<p>
+Entries should be added chronologically <d/> that is, newer entries
+should be placed towards the top of the file, underneath any initial
+header comment block.
```
ulm, line 20: I'd write "Entries are added chronologically ..." for a somewhat stronger statement.
No blocker for this PR though, it's also fine if you prefer the current wording.

profiles/use.mask/text.xml
```diff
@@ -32,7 +32,9 @@ defined purpose.
 </p>
 
 <p>
-Updates to <c>use.mask</c> should be handled via the relevant arch team.
+Updates to <c>use.mask</c> should be handled via the relevant arch team. Any
+additions should be sorted chronologically, starting at the top of the file
+(underneath any comment header blocks).
```
ulm, line 37: Same here.

ulm (approved):

profiles/package.mask/text.xml
thesamesam: Works for me, I'll change it, thanks!

PR #237, opened 2021-07-16
keywording: remove mentioning of STABLEREQ bugs.g.o keywords

Flowdalic:
This keyword is no longer used, NATTkA does not need it, and is
redundant to the "Stabilization Request" component of the "Gentoo
Linux" product, under which stabilization requests are filled.

Signed-off-by: Florian Schmaus <flow@gentoo.org>


ulm:
Thanks.

PR #240, opened 2021-08-01
profiles: mention PMS as the main source of documentation

thesamesam:
While `man portage` may be a useful source of information,
it's not the "source of truth" on profiles - PMS is.

Reported-by: Joshua Kinard <kumba@gentoo.org>

### DIFF
--- a/profiles/text.xml
+++ b/profiles/text.xml
@@ -5,8 +5,12 @@
 
 <body>
 <p>
-This section provides details on the <c>profiles/</c> directory. All of these files
-are also documented in <c>man portage</c>.
+This section provides details on the <c>profiles/</c> directory. All of these
+files are also documented in <c>man portage</c>, but the canonical reference
+for profiles is within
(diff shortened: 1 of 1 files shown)


profiles/text.xml
```diff
@@ -6,7 +6,10 @@
 <body>
 <p>
 This section provides details on the <c>profiles/</c> directory. All of these files
-are also documented in <c>man portage</c>.
+are also documented in <c>man portage</c>, but the canonical reference for profiles
+is within <uri link="https://projects.gentoo.org/pms/latest/pms.html#x1-410005">PMS</uri>.
```
ulm, line 10: As always, wrap lines at 80 characters at most.

Please use `8` instead of `latest`. The fraction identifier may not be the same for a future version.

ulm: Also, I just see that it points to the wrong section. The `profiles` dir is not described in chapter 5, but in section 4.4 "The Profiles Directory": https://projects.gentoo.org/pms/8/pms.html#x1-320004.4

PR #242, opened 2021-08-01
ebuild-writing/file-format: link to PMS for version specification

thesamesam:

### DIFF
--- a/ebuild-writing/file-format/text.xml
+++ b/ebuild-writing/file-format/text.xml
@@ -135,11 +135,12 @@ algorithm</uri> can be found in PMS.
 <body>
 
 <p>
-When packaging a snapshot of a source repository, there are two commonly used formats. The first
-treats the snapshot as a patch to the previous version, and so the ebuild version is in the format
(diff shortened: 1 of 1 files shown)


ebuild-writing/file-format/text.xml
```diff
@@ -120,18 +120,27 @@ EAPI 7 version commands</uri> may be used to manipulate and extract
 ebuild version components.
 </p>
 
+<p>
+The formal specification of version
+<uri link="https://projects.gentoo.org/pms/latest/pms.html#x1-250003.2">format
+</uri> and the comparison
+<uri link="https://projects.gentoo.org/pms/latest/pms.html#x1-260003.3">
+algorithm</uri> can be found in PMS.
```
ulm, line 128: Again, `8` instead of `latest` because TeX4ht doesn't provide immutable fragment identifiers. (I'm going to investigate if anything can be done about this.)

Also, could you put the word `format` on the next line? I guess it's within the guidelines as-is, but the `</uri>` at the beginning of the next line looks strange. :)

thesamesam: Understood! Yes, it's consistent this way too.

PR #244, opened 2021-08-01
ebuild-writing/common-mistakes: add calling external tools without die

thesamesam:
Suggested-by: Joonas Niilola <juippis@gentoo.org>

### DIFF
--- a/ebuild-writing/common-mistakes/text.xml
+++ b/ebuild-writing/common-mistakes/text.xml
@@ -14,6 +14,17 @@ writing ebuilds.
 <title>Common Ebuild Writing Mistakes</title>
 
 <subsection>
+<title>Unguarded external calls</title>
+<body>
+<p>
+Calls to external tools should be guarded with <c>|| die</c> (or use
+<c>assert</c>) in almost all cases to allow failure to be detected. See
+<uri link="::ebuild-writing/error-handling"/>.
+</p>
+</body>
+</subsection>
+
+<subsection>
 <title>Invalid use of <c>static</c> use-flag</title>
 <body>
 <p>


thesamesam:
@juippis Could you elaborate on what you envision for https://github.com/gentoo/devmanual/pull/223#issuecomment-882711204 re GLEP 81 and copy-pasting?

juippis:
For the GLEP part, it can most likely be disregarded now. I was kind of just looking for a way to "enforce" it, or made it more known through devmanual, as we still have devs pushing ebuilds using user.eclass. But with the late effort from @conikost I doubt the GLEP-81 suggestion would be relevant a year from now.

The copy-paste, it's really simple. Basically a way to say to "always take a look at the ebuild before testing/pushing", since if you just copypaste it, you end up pushing those EAPI-6 ebuilds in 2021 ;) + you might spot something unrelevant / outdated.
https://wiki.gentoo.org/wiki/Proxied_Maintainer_FAQ#.22Copy.2Fpaste.22_style_of_contributions
This is something we see ~daily in proxy-maint, and something we've seen even in retirement when checking activity.

ebuild-writing/common-mistakes/text.xml
```diff
@@ -13,6 +13,17 @@ writing ebuilds.
 <section>
 <title>Common Ebuild Writing Mistakes</title>
 
+<subsection>
+<title>Unguarded external calls</title>
+<body>
+<p>
+Calls to external tools should be guarded with <c>|| die</c> in almost
+all cases to allow failure to be detected. See
+<uri link="::ebuild-writing/error-handling"/>.
```
ulm, line 22: Is this really a common mistake? We don't even have a tracker bug for it (and AFAICS never had one).

If we are to include it, it should say "`die` or `assert`" I think.

ulm:
I don't understand how this PR is related to GLEP 81.

juippis:
> I don't understand how this PR is related to GLEP 81.

Don't worry, it's not.

ebuild-writing/common-mistakes/text.xml
thesamesam: It _really_ is. This is easily the most common issue that we have to comment on within proxy-maint@. But I appreciate why you're asking -- we've added a fair amount to this section recently and it's reasonable to be concerned about bloat.

I don't think we've ever bothered with tracker bugs given it's considered a "basic" QA issue and is also somewhat tricky to grep for.

thesamesam:
(Just for clarity: The GLEP 81 part was because I wanted to pick up on juippis' comments on the previous PR in a new one, but only one suggestion ended up being actionable for the time being. I was wondering if he wanted me to throw in anything else. Sorry for the confusion!)

ulm:
Merged, with fixed `uri` and whitespace.

PR #247, opened 2021-09-02
eclass-writing: @DEAD must be the first tag

ulm:


thesamesam (approved):
I thought we'd noted this explicitly at some point, but clearly not. Good spot.

PR #248, opened 2021-09-06
general-concepts/ebuild-revisions: mention EAPI bumps

juippis:


general-concepts/ebuild-revisions/text.xml
```diff
@@ -105,10 +102,25 @@ of thumb could be used as a guideline:
...
+    a dependency change that is a result of a package move (slot move),
+  </li>
+  <li>
+    a dependency change due to
+    <uri link="::ebuild-maintenance/removal/#removing-a-virtual-package">
+    removal of a virtual package.</uri>
   </li>
 </ul>
 
+<p>An ebuild's EAPI bump can be seen a bit controversial when it comes to
```
sbraz, line 114:
```suggestion
<p>An ebuild's EAPI bump can be a bit controversial when it comes to
```

general-concepts/ebuild-revisions/text.xml
```diff
@@ -105,10 +102,25 @@ of thumb could be used as a guideline:
...
+revision bumping. On the other hand an EAPI update might bring new <c>
```
sbraz, line 115:
```suggestion
revision bumping. On the one hand, an EAPI update might bring new <c>
```

general-concepts/ebuild-revisions/text.xml
```diff
@@ -105,10 +102,25 @@ of thumb could be used as a guideline:
...
+src_configure</c> and <c>src_install</c> defaults that changes the state of
+installed files, but on the other hand it can cause extra work for arch testing
```
sbraz, line 117:
```suggestion
installed files, but on the other hand, it can cause extra work for arch testing
```

general-concepts/ebuild-revisions/text.xml
```diff
@@ -105,10 +102,25 @@ of thumb could be used as a guideline:
...
+teams. You are encouraged to apply common sense and use your own judgement: If
```
sbraz, line 118:
```suggestion
teams. Developers are encouraged to apply common sense and use their own judgement: if
```

general-concepts/ebuild-revisions/text.xml
```diff
@@ -105,10 +102,25 @@ of thumb could be used as a guideline:
...
+the package has reverse dependencies, a revision bump is suggested. If you can
```
sbraz, line 119:
```suggestion
the package has reverse dependencies, a revision bump is suggested. If the package can be tested
to ensure that nothing breaks,  a revision bump might not be needed.
```

general-concepts/ebuild-revisions/text.xml
```diff
@@ -105,10 +102,25 @@ of thumb could be used as a guideline:
...
+test and be sure nothing breaks, a revision bump might not be needed. When
+merging user contributed EAPI ebuild bumps, it's always safer to revision bump.
```
sbraz, line 121: I don't agree. If you perform the same tests as if you were bumping yourself, there's no difference with a bump that you authored.

sbraz (changes requested):

ulm:
I fear that the paragraph on EAPI bumps is more confusing than it is helpful. I'd suggest to follow the KISS principle and add an item to the list of "examples of changes that warrant a new revision" along the lines of:
```
  <li>changing the EAPI (unless the changes to the ebuild are entirely trivial),</li>
```
The rationale is already given by the numbered list above, therefore I don't think that we need anything special in addition.

Second and third commit of the series look good.

general-concepts/ebuild-revisions/text.xml
juippis: My POV: I maintain packages that I use, and often I pick the m-n deps that are needed. I'm familiar with the package, and because I use it, I notice any breakage straight away. 

Now let's talk about m-n packages. When you merge those, do you test all revdeps? Especially if a trivial fix, like an EAPI bump, makes you compile chromium and libreoffice? Sometimes you have to think how much time you can give wrt the contribution's worth.  (FYI I do, but I don't expect everyone else does)

Nevertheless I still find the advice worth it, but what @ulm said works for me. However I'd like the rationale to be written open, since now you have to "compile" the information on your own.

juippis:
^ updated, also added a new commit.

general-concepts/ebuild-revisions/text.xml
sbraz: I think I get what you mean but the fact that a contribution is submitted by someone else doesn't imply that you aren't familiar with the package in question. For example, someone could be submitting a patch to one of the packages you maintain.

ulm (approved):

PR #251, opened 2021-09-17
Fix formatting issues

thesamesam:
Reported-by: Ulrich Müller <ulm@gentoo.org>


ulm:
Thanks.

PR #253, opened 2021-09-28
general-concepts/mirrors: update info about mirror://gentoo

juippis:
- make it very clear that d.g.o:/space/distfiles-local is not
   accessible anymore, and show the current way of hosting your own
   distfiles.

As noted by @thesamesam


general-concepts/mirrors/text.xml
```diff
@@ -128,14 +128,19 @@ where <c>myname</c> refers to the username of the developer.
...
+</p>
 
+<p>
 If the upstream download location for a package uses a non-standard TCP port
-(anything other than 21, 80 or 443), you <e>must</e> manually mirror the files. Not
-doing so can cause all kinds of problems with strict firewalls.
+(anything other than 21, 80 or 443), you <e>must</e> manually mirror the files.
+Not doing so can cause all kinds of problems with strict firewalls.
 </p>
 </body>
```
ulm, line 146: I wonder if we shouldn't drop the entire section. The second paragraph could be added at the end of the previous section.

I'd drop the paragraph about `/space/distfiles-local` altogether. It is only a historical curiosity which we need not document.

juippis: I really thought about it as well, but though it was nice to explain where possible `mirrors://gentoo` came from. Yeah, I can definitely combine the sections.

general-concepts/mirrors/text.xml
```diff
@@ -100,23 +100,24 @@ devspace.
...
+Previous policy was to use <c>mirror://gentoo</c> directly, but this is now
+deprecated, as that wouldn't allow to have long-term availability and
+traceability of the source files, which might be a requirement of the license.
 </p>
 
 <p>
-When you upload the file to <c>dev.gentoo.org:~/public_html</c>, you must ensure that it, and its
-parent directories, are world-readable. An example <c>SRC_URI</c> referencing
-a distfile mirrored this way is as follows:
+When you upload the file to <c>dev.gentoo.org:~/public_html</c>, you must ensure
```
sbraz, line 109: I thought we avoided "you".

general-concepts/mirrors/text.xml
```diff
@@ -100,23 +100,24 @@ devspace.
...
 </p>
 
 <codesample lang="ebuild">
 SRC_URI="https://dev.gentoo.org/~myname/distfiles/${P}.tar.gz"
 </codesample>
 
 <p>
-where <c>myname</c> refers to the username of the developer.
+where <c>myname</c> refers to the username of the developer. Make sure your
+distfile has the correct permissions, so it's accessible.
```
sbraz, line 120: Doesn't the above paragraph already address the possible permission problems?

general-concepts/mirrors/text.xml
ulm: [Third person form should be used rather than first.](https://devmanual.gentoo.org/appendices/contributing/index.html#style-guidelines)

Nothing about second person. :smile:

juippis:
All should be addressed.

general-concepts/mirrors/text.xml
```diff
@@ -100,15 +100,16 @@ devspace.
...
+prohibited, as that wouldn't allow to have long-term availability and
+traceability of the source files, which might be a requirement of the license.
 </p>
 
 <p>
-When you upload the file to <c>dev.gentoo.org:~/public_html</c>, you must ensure that it, and its
-parent directories, are world-readable. An example <c>SRC_URI</c> referencing
-a distfile mirrored this way is as follows:
+When you upload the distfile to <c>dev.gentoo.org:~/public_html</c>, ensure that
+your file, and its parent directories have the correct permissions, so they're
```
ulm, line 110:
```suggestion
your file and its parent directories have the correct permissions, so they're
```
I think there shouldn't be a comma.

PR #257, opened 2021-11-12
archs/amd64: Update information on profiles

ulm:
Profile information was way behind (2005); this updates it to 17.0 and 17.1.

@gentoo/amd64: Is the table in subsection "Libdir links" correct like this?

### DIFF
--- a/archs/amd64/text.xml
+++ b/archs/amd64/text.xml
@@ -232,21 +232,9 @@ this behaviour is controlled by the <c>MULTILIB_STRICT_EXEMPT</c> variable in
 <body>
 
 <p>
-In most cases, it's sufficient to use the <c>$(get_libdir)</c> function:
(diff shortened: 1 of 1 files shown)


archs/amd64/text.xml
```diff
@@ -143,141 +143,66 @@ variable and <c><uri link="::eclass-reference/multilib.eclass/"/></c>).
...
-    <ti>
-      d32
-    </ti>
-    <ti>
-      d64
-    </ti>
+    <ti>17.1/no-multilib</ti>
+    <ti>n/a</ti>
+    <ti>inexistent</ti>
+    <ti>amd64</ti>
```
thesamesam, line 196: nonexistent

ulm: Hm, that word was there before, and Oxford online lists it: https://www.lexico.com/definition/inexistent

Anyway, updated.

PR #260, opened 2022-01-01
ebuild-maintenance/package-moves: add common terms for pkgmove/slotmove

thesamesam:


ebuild-maintenance/package-moves/text.xml
```diff
@@ -5,9 +5,9 @@
 <body>
 
 <p>
-This chapter describes the use of package and slot moves. The package update
-mechanism is a powerful tool, and needs to be used with caution. In particular,
-the following must be noted:
+This chapter describes the use of package ("pkgmove" / "move") and slot moves
+("slotmove"). The package update mechanism is a powerful tool, and needs to be
+used with caution. In particular, the following must be noted:
```
ulm, line 10: Not sure if these technical terms should be mentioned already in the introductory sentence of the chapter. They're explained in their proper context below.

ulm (commented):
Rest looks good.

ebuild-maintenance/package-moves/text.xml
thesamesam: I take the point on _slotmove_, but wanted to include _pkgmove_ at least as it's an unofficial term which is often used on IRC and would help people search for it. But then figured if I was including one, I should mention the other 😄. What do you think?

ebuild-maintenance/package-moves/text.xml
```diff
@@ -77,6 +78,9 @@ and update them. These include:
   <li>
     open bug summaries
   </li>
+  <li>
+    wiki articles
```
ulm, line 82: IIUC the technical term is "wiki pages".

ulm (approved):
One small suggestion, otherwise LGTM.

ebuild-maintenance/package-moves/text.xml
thesamesam: Thank you!

PR #264, opened 2022-01-16
devbook.rnc: Change from DTD to RELAX NG compact schema

ulm:


ulm:
```
$ trang -I rnc -O dtd devbook.rnc devbook.dtd
/local/home/ulm/git/devmanual/devbook.rnc:110:26: error: sorry, cannot handle t\
his kind of "choice"
```
Success :smile:

thesamesam:
Would you mind explaining the rationale a bit in the commit message?

thesamesam:
> > Would you mind explaining the rationale a bit in the commit message?
> 
> Better now?

Brilliant, thank you for that!

ulm:
I think this is ready for production now. Please review.

The spec for the `uri` element is slightly more lax than it could be. This way, conversion into XML schema is still possible (in case we would ever need it). I still get a warning, because RELAX NG is more expressive (at least for that item) than XML schema:
```
$ trang -I rnc -O xsd devbook.rnc devbook.xsd
/local/home/ulm/git/devmanual/devbook.rnc:112:3: warning: choice between attributes and children cannot be represented; approximating
```

Makefile
```diff
@@ -82,10 +82,16 @@ install: all
...
-validate:
-	@xmllint --noout --dtdvalid devbook.dtd $(XMLS)
+# Not all versions of xmllint support --quiet, so test for it first
+validate: devbook.rng
+	@opt=--quiet; xmllint --help 2>&1 | grep -q -- --quiet || opt=; \
+	xmllint --noout $${opt} --relaxng $< $(XMLS)
 	@echo "xmllint validation successful"
 
+%.rng: %.rnc
+	trang -I rnc -O rng $< $@
```
thesamesam, line 92: Could you update `appendices/contributing/text.xml` to note this new dependency?

thesamesam (approved):

Makefile
ulm: Not sure if we should mention it. I went for the pragmatic approach to commit also `devbook.rng` to the repository, so trang won't normally be needed, neither for editing the XML sources nor for building the devmanual. Purists may object against having a generated file in the repository, but it is a small and machine-independent file, and it avoids app-text/trang as a dependency, which in turn depends on Java.

I very much trust that someone who will find themselves in the position to edit the RELAX NG schema will also know how to install trang. :smiley:

thesamesam: Okay, good point. WFM!

ulm:
Merged.

PR #265, opened 2022-01-21
general-concepts/use-flags: give guidance on IUSE defaults

thesamesam:

### DIFF
--- a/general-concepts/use-flags/text.xml
+++ b/general-concepts/use-flags/text.xml
@@ -203,6 +203,18 @@ Add <c>+</c> or <c>-</c> before the name of the use flag in <c>IUSE</c> to turn
 it on or off by default.
 </p>
 
+<p>
+IUSE defaults should be used sparingly. Reasons to exclude/default-disable a
+feature may include e.g. large build time for a dependency, or a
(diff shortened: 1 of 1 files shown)


orlitzky:
Because portage treats flags as off by default, and because of the default `USE_ORDER`, it's _much_ harder for people to disable flags that are default-on than it is for them to enable flags that are not.

Making this a de-facto policy is really annoying for the people who want to enable only what they want and leave everything else off. (That said, it's already a pain in the ass, because every maintainer does whatever he wants at the moment, and many use `IUSE` defaults as their own personal `make.conf` replacement.)

A policy like "`IUSE` defaults should only be set when the package would otherwise be severely handicapped" would be more in line with our tooling.

orlitzky:
Also, in my experience, many upstreams use `configure.ac` as their own personal `make.conf` replacement. In those cases, standardizing what the main developer likes to use on his machine isn't going to produce very consistent results.

sbraz:
> A policy like "IUSE defaults should only be set when the package would otherwise be severely handicapped" would be more in line with our tooling.

I like the general idea.

thesamesam:
yes, that alternative is fine with me

ulm (approved):

PR #267, opened 2022-01-22
keywording: note that stabilization rules are for actual testing

thesamesam:
These rules are for maintainers wishing to stable if they have
access to hardware -- they do _not_ mean maintainers must
follow these rules just to file a bug and stabilise a package
via CCing arch teams (the usual route).

Closes: https://bugs.gentoo.org/546942

### DIFF
--- a/keywording/text.xml
+++ b/keywording/text.xml
@@ -379,6 +379,11 @@ Vulnerability Treatment Policy</uri>.
 <title>Stabilization rules</title>
 <body>
 
+<p>
+These are rules for stabilizing packages <e>by yourself</e> on a particular
+architecture, <e>not</e> for filing bugs to request arch teams to handle it.
+</p>
+
 <ul>
   <li>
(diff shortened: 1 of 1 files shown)


keywording/text.xml
```diff
@@ -379,6 +379,11 @@ Vulnerability Treatment Policy</uri>.
 <title>Stabilization rules</title>
 <body>
 
+<p>
+These are rules for stabilzing packages <e>by yourself</e> on a particular
```
ulm, line 383:
```suggestion
These are rules for stabilizing packages <e>by yourself</e> on a particular
```

keywording/text.xml
```diff
@@ -379,6 +379,11 @@ Vulnerability Treatment Policy</uri>.
...
+architecture, <b>not</b> for filing bugs to request arch teams to handle it.
```
ulm, line 384: Please use `<e>` (_not_ `<b>`) for emphasis.

PR #270, opened 2022-01-24
keywording: it's preferred to rekeyword packages with new deps

thesamesam:
It's preferred to rekeyword packages with their new dependencies in the
same bug rather than using `package.use.mask` as this means the new
code path(s) using the new dependencies get tested rather than just
the new package/dependency in isolation.

Also, no tooling exists for arch testers to automatically remove
the relevant mask(s) once keywording is done, so often we end up
with stale unnecessary masks remaining in tree.

### DIFF
--- a/keywording/text.xml
+++ b/keywording/text.xml
@@ -239,6 +239,16 @@ with <c>repoman</c> or <c>pkgdev</c> <d/> instead please commit with
 git commit message.
 </p>
 
+<p>
+Note that it is preferred to drop keywords on the package and request
+rekeywording of it together with its new dependencies within the same bug to
+allow the new code path(s) in your package to be tested. This won't happen if
+the new dependency is requested for keywording by itself and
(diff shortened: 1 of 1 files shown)


keywording/text.xml
```diff
@@ -239,6 +239,16 @@ with <c>repoman</c> or <c>pkgdev</c> <d/> instead please commit with
...
 </p>
 
+<p>
+Note that it is preferred to drop keywords on the package and request
+rekeywording of it together with its new dependencies within the same bug to
+allow the new code path(s) in your package to be tested. This won't happen if
+the new dependency is requested for keywording by itself and
+<c>package.use.mask</c> is used to mask the relevant new USE flag: only the
+new package dependency will be tested by arch testers. Also, the mask has to be
+manually removed, and this step is not usually automated.
```
ulm, line 249: The last sentence sounds a little funny: manually already implies that it's not automated.

ulm (changes requested):

PR #272, opened 2022-02-22
general-concepts/projects: consult project lead on joining; mention project IRC channels

thesamesam:


general-concepts/projects/text.xml
```diff
@@ -72,7 +72,14 @@ lists the members of the project. Simply modifying the list is
 sufficient for adding or removing a developer. Note that different
 projects have different requirements and procedures for recruiting
 developers, which may require prior arrangements to be made before
-modifying the member list.
+modifying the member list. It is standard however to consult
+the project lead.
+</p>
+
+<p>
+If the project has an official IRC channel listed on the wiki,
```
ulm, line 80: Maybe "on their project page" instead of "on the wiki"?

thesamesam: Ah, yes!

PR #273, opened 2022-03-01
keywording: describe pending stabilization requests

thesamesam:
Describe how maintainers should keep track of (and know when to file)
pending stabilization requests.

### DIFF
--- a/keywording/text.xml
+++ b/keywording/text.xml
@@ -435,6 +435,36 @@ keyword, so packages are not to be marked stable for one of these architectures.
 </subsection>
 
 <subsection>
+<title>Keeping track of pending stabilizations</title>
+<body>
+
+<p>
+Maintainers need some method or system to organize and start pending
+stabilizations once they become due.
+</p>
+
+<p>
+There are several tools available that can help with this:
+</p>
+
+<ul>
+  <li>
+    Use <c>imlate</c> from app-portage/gentoolkit
+  </li>
+  <li>
+    Use packages.gentoo.org's maintainer pages (which have a Stabilization
+    tab)
+  </li>
+  <li>
+    Use <c>pkgcheck</c>'s <c>StableRequest</c> check, e.g.
+      <c>grep -ri "larry@" */*/metadata.xml -l | cut -d'/' -f1-2 | xargs pkgcheck scan -k StableRequest</c>
+  </li>
+</ul>
+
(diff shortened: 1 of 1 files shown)


keywording/text.xml
```diff
@@ -434,6 +434,36 @@ keyword, so packages are not to be marked stable for one of these architectures.
...
+stabilizations once they become due.
+</p>
+
+<p>
+There are several options available:
+</p>
+
+<ul>
+  <li>
+    Use <c>imlate</c> from app-portage/gentoolkit
```
laumann, line 452: Maybe a comment saying that imlate appears unmaintained and doesn't understand projects?

thesamesam: Oh, it does work, it just needs explicit email via -M.

keywording/text.xml
```diff
@@ -434,6 +434,36 @@ keyword, so packages are not to be marked stable for one of these architectures.
...
+  </li>
+  <li>
+    Use packages.gentoo.org's maintainer pages (which have a Stabilization
+    tab)
+  </li>
+  <li>
+    Use <c>pkgcheck</c>'s <c>StableRequest</c> check, e.g.
+      <c>grep -ri "larry@" */*/metadata.xml -l | cut -d'/' -f1-2 | xargs pkgcheck scan -k StableRequest</c>
```
mattst88, line 460: I'd love to add a script to [mgorny/mgorny-dev-scripts](https://github.com/mgorny/mgorny-dev-scripts) that just does this.

keywording/text.xml
```diff
@@ -434,6 +434,36 @@ keyword, so packages are not to be marked stable for one of these architectures.
...
+<title>Keeping track of pending stabilizations</title>
+<body>
+
+<p>
+Maintainers need some method or system to organize and start pending
+stabilizations once they become due.
+</p>
+
+<p>
+There are several options available:
```
ulm, line 447: One could read this as being the only options, which isn't the case. For example, I keep track of my version bumps in an old-fashioned list with Emacs' org-mode. :smiley:

Maybe something like the following instead?
```suggestion
There are several tools available that can help with this:
```

keywording/text.xml
thesamesam: Me too. It's _almost_ there with the components but not quite.

PR #274, opened 2022-03-11
Remove references to repoman

mattst88:
Signed-off-by: Matt Turner <mattst88@gentoo.org>

### DIFF
--- a/appendices/common-problems/text.xml
+++ b/appendices/common-problems/text.xml
@@ -161,7 +161,7 @@ directories Executable stacks ... binary files whose stack is marked with
 </section>
 
 <section>
-<title>Handling <c>repoman</c> Messages</title>
+<title>Handling <c>pkgcheck</c> Messages</title>
 <body>
 
 <todo>
--- a/ebuild-maintenance/git/text.xml
+++ b/ebuild-maintenance/git/text.xml
@@ -112,11 +112,10 @@ the historical commits after the initial git commit.
 <body>
 
 <p>
-The recommended way of committing to the Gentoo repository is to use <c>repoman
-commit</c> or <c>pkgdev commit</c> (then <c>pkgdev push</c>). It automatically
-performs the necessary QA checks on the package being committed and has other
-features helping with the Gentoo workflow. However, <c>repoman</c> is currently
-limited to creating a single commit to a single package.
+The recommended way of committing to the Gentoo repository is to use <c>pkgdev
+commit</c> (then <c>pkgdev push</c>). It automatically performs the necessary
+QA checks on the package being committed and has other features helping with
+the Gentoo workflow.
 </p>
 
 <p>
@@ -130,21 +129,19 @@ used. The valid uses of git include:
     Gentoo repository (eclasses, licenses, profiles…),
   </li>
   <li>
-    amending a commit created via <c>repoman commit</c> or <c>pkgdev commit</c>
-    with additional files or fixups,
+    amending a commit created via <c>pkgdev commit</c> with additional files
+    or fixups,
   </li>
   <li>
-    combining multiple commits created via <c>repoman commit</c> or
-    <c>pkgdev commit</c> using <c>git rebase</c>.
+    combining multiple commits created via <c>pkgdev commit</c> using
+    <c>git rebase</c>.
   </li>
 </ul>
 
 <p>
-Whenever <c>repoman</c> or <c>pkgdev</c> is not used to commit, you need to
-manually verify all packages affected by the commit using <c>repoman full</c> or
-<c>pkgcheck scan --commits</c>. When using <c>repoman</c>, it won't be aware of
-staged changes, so ensure that all files are included in the commit.
-Also, when using <c>git</c> manually, you must perform a manual sign-off to the
+Whenever <c>pkgdev</c> is not used to commit, you need to manually verify all
+packages affected by the commit using <c>pkgcheck scan --commits</c>. Also,
+when using <c>git</c> manually, you must perform a manual sign-off to the
 <uri link="::general-concepts/copyright-policy/#Certificate of origin"/> using
 the <c>-s </c> or <c>--signoff</c> option with your git commit commands. Make
 sure you have read and understand the actual certificate.
@@ -224,8 +221,8 @@ appropriately:
 
 <ul>
   <li>
-    <c>${CATEGORY}/${PN}:</c> Single Package (Note that <c>repoman commit</c>
-    and <c>pkgdev commit</c> will automatically insert this for you)
+    <c>${CATEGORY}/${PN}:</c> Single Package (Note that <c>pkgdev commit</c>
+    will automatically insert this for you)
   </li>
   <li><c>${CATEGORY}:</c> Package Category</li>
   <li><c>profiles:</c> Profile Directory</li>
@@ -286,15 +283,11 @@ are optionally used in Gentoo:
     referenced, it will be closed (as fixed) automatically with reference
     to the commit.
   </li>
-  <li>
-    <c>Package-Manager:</c> This is automatically inserted by
-    <c>repoman commit</c> and it specifies the version of
-    <c>sys-apps/portage</c> on the system.
+  <li>  
+    <c>Package-Manager:</c> Used by repoman to indicate Portage version
   </li>
   <li>
-    <c>RepoMan-Options:</c> This is automatically inserted by
-    <c>repoman commit</c> and records the options passed to repoman (such as
-    --force) for the commit.
+    <c>RepoMan-Options:</c> Used by repoman to indicate repoman options
   </li>
 </ul>
 
--- a/ebuild-maintenance/new-ebuild/text.xml
+++ b/ebuild-maintenance/new-ebuild/text.xml
@@ -54,10 +54,10 @@ examples.
 </p>
 
 <p>
-When committing to git, all developers should use <c>repoman commit</c> or
-<c>pkgdev commit</c> with <c>pkgdev push</c> instead of <c>git commit</c> to
-submit their ebuilds.  Before committing, please run <c>repoman full</c> or
-<c>pkgcheck scan</c> to make sure you didn't forget something.
+When committing to git, all developers should use <c>pkgdev commit</c> with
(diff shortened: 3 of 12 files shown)


keywording/text.xml
```diff
@@ -232,13 +232,6 @@ new bug</uri> to the architecture in question regarding this if no bug is
 already available.
 </p>
 
-<p>
-If there are no new dependencies, do not remove keywords if your commit fails
-with <c>repoman</c> or <c>pkgdev</c> <d/> instead please commit with
-<c>repoman -I</c> and file a bug to the broken architecture, noting it in your
-git commit message.
-</p>
-
```
mattst88, line 241: I discussed this with @floppym and @benkohler on `#gentoo-dev`. As best as we can figure, this is referring to a situation where the depgraph is *already* broken, and you're just trying to make a version bump.

The text doesn't make a ton of sense because it's telling you to commit with `repoman -I` but `repoman(1)` describes `-I, --ignore-masked` as "Ignore masked packages (not allowed with commit mode)".

So we figure the best thing to do is to simply drop the paragraph.

ebuild-maintenance/new-ebuild/text.xml
```diff
@@ -54,10 +54,10 @@ examples.
 </p>
 
 <p>
-When committing to git, all developers should use <c>repoman commit</c> or
-<c>pkgdev commit</c> with <c>pkgdev push</c> instead of <c>git commit</c> to
-submit their ebuilds.  Before committing, please run <c>repoman full</c> or
-<c>pkgcheck scan</c> to make sure you didn't forget something.
+When committing to git, all developers should use <c>pkgdev commit</c> with
+<c>pkgdev push</c> instead of <c>git commit</c> to submit their ebuilds.
+Before committing, please run <c>pkgcheck scan</c> to make sure you didn't
```
sbraz, line 59: Maybe repeat `pkgcheck scan --commits` here too?

mattst88: Good suggestion. Will do.

keywording/text.xml
```diff
@@ -196,8 +196,8 @@ a dependency of a package already keyworded.
 
 <p>
 Do <b>not</b> commit straight to <c>arch</c> ("stable"). If it is necessary
-to do so, when using <c>repoman</c>, use
-<c>repoman commit --straight-to-stable</c> and <b>not</b> <c>--force</c> as
+to do so, when using <c>pkgdev</c>, use
+<c>pkgdev commit --ask --scan</c> and <b>not</b> by dropping <c>--scan</c> as
```
juippis, line 200: You can `pkgdev commit` straight to stable just fine, it won't nag until you run `--scan`. Just saying for normal workflow you recommend `pkgdev commit` and this shouldn't be different, ie, you often catch the error only when doing `--scan`.

general-concepts/manifest/text.xml
```diff
@@ -15,10 +15,9 @@ by the package. This is used to verify integrity upon fetching them.
 </p>
 
 <p>
-To generate the <c>Manifest</c>, use <c>ebuild foo.ebuild manifest</c>,
-<c>repoman manifest</c>, or <c>pkgdev manifest -m</c>. You may want to set
-<c>GENTOO_MIRRORS=</c> while calling it to fetch distfiles from their original
-locations immediately.
+To generate the <c>Manifest</c>, use <c>ebuild foo.ebuild manifest</c> or
+<c>pkgdev manifest -m</c>. You may want to set <c>GENTOO_MIRRORS=</c> while
+calling it to fetch distfiles from their original locations immediately.
```
juippis, line 20: Encouraging `<c>pkgdev manifest -m</c>` and then saying `You may want to set <c>GENTOO_MIRRORS=</c>` conflicts with each other :)

I've never had problems with pure `pkgdev manifest`... even think it prefers GENTOO_MIRRORS.

appendices/common-problems/text.xml
```diff
@@ -161,7 +161,7 @@ directories Executable stacks ... binary files whose stack is marked with
 </section>
 
 <section>
-<title>Handling <c>repoman</c> Messages</title>
+<title>Handling <c>pkgcheck</c> Messages</title>
```
juippis, line 164:
```
<section>
<title>Handling <c>pkgcheck</c> Messages</title>
<body>

<todo>
write me
</todo>

</body>
</section>
```
Wanna go at it? ;)
Wonder if there should just be a link to pkgcheck documentation

juippis: Actually maybe we want a whole new devmanual page dedicated to pkgcheck/pkgdev, I could add some parts from #235 there.

general-concepts/manifest/text.xml
mattst88: I agree. I don't understand that either, but it's not my objective to resolve that in this patch.

Looks like @thesamesam added this in commit 00c1824be4d4bd2a72a54be9846846478d25ada1. Let's resolve that in a follow up.

appendices/common-problems/text.xml
mattst88: I think that would be great, and I'd be happy to contribute to it, but I don't want to block this patch on it.

juippis: I'm saying maybe this section can be deleted altogether now you're touching it, since "nothing existed here before".

general-concepts/manifest/text.xml
radhermit: GENTOO_MIRRORS usage is disabled for `pkgdev manifest` by default, passing the -m option enables them. See https://github.com/pkgcore/pkgdev/blob/main/src/pkgdev/scripts/pkgdev_manifest.py#L37 ... or in other words read the man page. :smile:

ebuild-maintenance/git/text.xml
```diff
@@ -286,16 +283,6 @@ are optionally used in Gentoo:
...
-  <li>
-    <c>Package-Manager:</c> This is automatically inserted by
-    <c>repoman commit</c> and it specifies the version of
-    <c>sys-apps/portage</c> on the system.
-  </li>
-  <li>
-    <c>RepoMan-Options:</c> This is automatically inserted by
-    <c>repoman commit</c> and records the options passed to repoman (such as
-    --force) for the commit.
-  </li>
```
ulm, line 298: This describes where these tags originate from, as they are present in existing commits. I suggest to keep both items.

ebuild-maintenance/git/text.xml
```diff
@@ -224,8 +221,8 @@ appropriately:
 
 <ul>
   <li>
-    <c>${CATEGORY}/${PN}:</c> Single Package (Note that <c>repoman commit</c>
-    and <c>pkgdev commit</c> will automatically insert this for you)
```
ulm, line 228: Again, this is simply a fact, no need to remove this information.

ulm (commented):
We should distinguish between updating the recommended workflow, and documentation of the repoman tool. I don't see a compelling reason to remove all mentions of the latter.

ebuild-maintenance/git/text.xml
mattst88: > We should distinguish between updating the recommended workflow, and documentation of the repoman tool. I don't see a compelling reason to remove all mentions of the latter.

I disagree. It's not a good idea to continue mentioning `repoman` in the devmanual when we want developers to stop using it.

ebuild-maintenance/git/text.xml
mattst88: I'm ambivalent, slightly leaning towards removing them.

If no one is using `repoman`, this only serves to document what the tags in git history mean -- which seem entirely self-explanatory. I don't see any value in keeping them in the devmanual.

If you feel that strongly about keeping them, I guess I could change their wording to use the past tense, e.g. instead of "This *is* automatically inserted by repoman commit" → "This *was* automatically inserted by repoman commit" to make it clear that repoman commit is the old, deprecated tool.

ebuild-maintenance/git/text.xml
mattst88: > Again, this is simply a fact, no need to remove this information.

But there's no value in keeping it. It only stands to lead readers to the false conclusion that repoman is a tool they should use.

keywording/text.xml
mattst88: Yeah. So how should I change this paragraph? repoman would have blocked committing straight to stable; pkgdev doesn't (unless you use --scan) and instead assumes that you'll run `pkgcheck scan` later.

I think I should probably just remove the paragraph.

thesamesam: Just keep the first sentence.

mattst88:
> We should distinguish between updating the recommended workflow, and documentation of the repoman tool. I don't see a compelling reason to remove all mentions of the latter.

I disagree. It's not a good idea to continue mentioning repoman in the devmanual when we want developers to stop using it.

ebuild-maintenance/git/text.xml
mattst88: Changed to use the past tense.

keywording/text.xml
mattst88: WFM.

ebuild-maintenance/git/text.xml
```diff
@@ -287,13 +284,13 @@ are optionally used in Gentoo:
...
-    <c>repoman commit</c> and it specifies the version of
+    <c>Package-Manager:</c> This was automatically inserted by
+    <c>repoman commit</c> and specified the version of
     <c>sys-apps/portage</c> on the system.
   </li>
   <li>
-    <c>RepoMan-Options:</c> This is automatically inserted by
-    <c>repoman commit</c> and records the options passed to repoman (such as
+    <c>RepoMan-Options:</c> This was automatically inserted by
+    <c>repoman commit</c> and recorded the options passed to repoman (such as
```
ulm, line 293: Past tense is somewhat misleading, because it suggests that this functionality was removed from repoman.

Maybe just "Automatically inserted by `repoman` and specifies ..."?

ebuild-maintenance/git/text.xml
ulm: wfm

ebuild-maintenance/git/text.xml
mattst88: Sounds good, thanks!

PR #276, opened 2022-03-15
profiles/package.mask: mention masking development versions

thesamesam:
Closes: https://bugs.gentoo.org/319927

### DIFF
--- a/profiles/package.mask/text.xml
+++ b/profiles/package.mask/text.xml
@@ -15,6 +15,23 @@ of the <c>package.mask</c> file is described in <c>man portage</c>.
 </p>
 
 <p>
+Development or unstable (per upstream declaration/categorization) versions of
+packages should usually be masked in <c>package.mask</c>. Upstreams may not
+deem such releases to be ready for general distribution (or safe to use), or
+may not be expecting bug reports from the wider userbase yet. The default
+should generally be to mask such versions, but it is acceptable to not mask
+in some circumstances <d/> e.g. upstream make very infrequent releases, the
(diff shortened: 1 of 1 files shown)


profiles/package.mask/text.xml
```diff
@@ -14,6 +14,23 @@ have a comment detailing the specific reason for the mask. The format
...
 </p>
 
+<p>
+Development or unstable (per upstream declaration/categorization) versions of
+packages should usually be masked in <c>package.mask</c>. Upstreams may not
+deem such releases to be ready for general distribution (or safe to use), or
+may not be expecting bug reports from the wider userbase yet. The default
+should generally be to mask such versions, but it is acceptable to not mask
+in some circumstances <d/> e.g. upstream make very infrequent releases, the
+changes are safe (reviewed by the Gentoo maintainer), and perhaps other
```
sbraz, line 24: Do you really mean "and" or "or", here?

ulm (commented):
LGTM.

Maybe add some sentence like the following?
"As an alternative to a development version, you may also consider backporting required upstream fixes to the released version."

profiles/package.mask/text.xml
thesamesam: I think I really did mean "and", but I think "or" is going to avoid ambiguity and it's clear you should exercise care when doing this. So I'll change to "or" for now.

thesamesam:
Done in 15dee0505f502a5eb3a248257b61f02f3f30bee9.

PR #278, opened 2022-03-16
eclass-writing: Add default src_configure to jmake example

laumann:

### DIFF
--- a/eclass-writing/text.xml
+++ b/eclass-writing/text.xml
@@ -667,7 +667,7 @@ src_compile() {
 <body>
 
 <p>
-A simple eclass which defines a number of functions and a default
+A simple eclass which defines a number of functions and default <c>src_confgure</c> and
 <c>src_compile</c> for the (hypothetical) <c>jmake</c> build system might look
 something like the following:
 </p>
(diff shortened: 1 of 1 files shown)


eclass-writing/text.xml
```diff
@@ -667,7 +667,7 @@ src_compile() {
 <body>
 
 <p>
-A simple eclass which defines a number of functions and a default
+A simple eclass which defines a number of functions and default <c>src_confgure</c> and
 <c>src_compile</c> for the (hypothetical) <c>jmake</c> build system might look
 something like the following:
```
laumann, line 672: I think this paragraph could be better, I could rework it a bit.

eclass-writing/text.xml
```diff
@@ -686,7 +686,7 @@ something like the following:
 # Demonstrates EXPORT_FUNCTIONS and defines simple wrappers for the
 # (hypothetical) jmake build system and a default src_compile.
```
thesamesam, line 687: We should update this comment too.

eclass-writing/text.xml
thesamesam: What are you thinking?

eclass-writing/text.xml
laumann: Done.

laumann:
@thesamesam Rebased on master as well.

eclass-writing/text.xml
thesamesam: Yeah, let's throw that in. Separate commit.

It reads easier :)

laumann: Done.

thesamesam (approved):
Much better, thank you!

PR #281, opened 2022-03-27
ebuild-writing/patches: Add "Conditional patching" subsection

laumann:
The added section under "Patches" tries to describe what patches are
intended for and why conditional patching should be avoided.

It was pointed out in a PR review that conditional patching in general
should be avoided as it introduces more variance and can make it harder
to debug breakage if a patch is only applied under certain conditions
(USE flags being the primary example).

---
The discussion/PR: https://github.com/gentoo/gentoo/pull/24709#discussion_r832361402

### DIFF
--- a/ebuild-writing/misc-files/patches/text.xml
+++ b/ebuild-writing/misc-files/patches/text.xml
@@ -113,6 +113,65 @@ from the <c>vim</c> patch tarball:
 </section>
 
 <section>
+<title>Conditional patching</title>
+<body>
+
+<p>
+Patching is ideally only done to make the package in question build properly,
+and should not be done to modify the runtime behaviour of the package. This is
+what USE flags and features of the package are for. As such, it is preferable to
+apply patches unconditionally and avoid conditional patching.
+</p>
+
+<p>
+There are a number of reasons to avoid conditional patching:
+</p>
+
+<ul>
+  <li>
+    It may go unnoticed that a patch fails to apply, if a package is not being
+    tested with a given flag
+  </li>
+  <li>
+    More variance is introduced and problems with a package can become much more
+    difficult to debug
+  </li>
+  <li>
+    Patches should preferably be upstreamed, but conditional patches cannot
+  </li>
+</ul>
+
+<p>
+Consider the following example <c>src_prepare()</c> implementation:
+</p>
+
+<codesample lang="ebuild">
+src_prepare() {
+	if use threads; then
+		PATCHES+=( "${FILESDIR}"/${P}-mt.patch )
+	fi
+	default
+}
+</codesample>
+
+<p>
+As this patch is only applied when <c>USE="threads"</c> is set, any developer
+creating new versions of this package might not detect whether the patch applies
+successfully if they don't test with the same flag.
+</p>
+
+<p>
+Although conditional patching is discouraged it can be unavoidable and as such,
+it is not considered a banned practice, but, to the extent possible, patches
+should be written such that they affect behaviour correctly based on e.g. build
+time definitions and configuration options rather than USE flags directly. This
+allows them to be applied unconditionally.
+</p>
+
+</body>
+</section>
+
+<section>
 <title>Clean Patch Howto</title>
 <body>
 


laumann:
@thesamesam Here's a first ~giraffe~ draft

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -331,6 +331,34 @@ scrub_patch some-patch-you-found.patch
...
+	if use threads; then
+		PATCHES+=( "${FILESDIR}/${P}"-mt.patch )
+	fi
+	default
+}
+</codesample>
+<p>
+As this patch is only applied when <c>USE="threads"</c> is set, any developer
+creating new versions of this package might not detect whether the patch applies
+successfully if they don't test with the same flag.
```
thesamesam, line 357: Let's add in a new paragraph that this can become unsustainable too, or even impossible to test, with USE flag combinatorics. So, while it's undesirable for even one or two patches, it becomes truly untestable and unfit for production with several flags if more than one applies patches.

We should also explicitly state that patches should be written such that they affect behaviour correctly based on e.g. build time definitions or config options rather than USE flags directly, allowing them to be applied unconditionally.

(Also, for e.g. musl patches, we want a portable fix, not a hack which is only applied _for_ musl, which kind of makes the problem worse.)

Nice work!

ulm:
Two comments:
- The "Clean patch howto" section is about formatting the patch itself (e.g. how to name it, put comment in it). So this seems misplaced as a subsection there. It should go in its own section instead (I'd suggest the end of the chapter, or before "Clean patch howto").
- I believe that the more common case is/was USE-conditional patching depending on an architecture flag (like `ppc64`). The reasoning may need some tweaking to cover that case too.

Also, we recommend conditional patching [elsewhere](https://github.com/gentoo/devmanual/blob/master/ebuild-writing/use-conditional-code/text.xml#L42). So this is a policy change. While I'm all for it, it may be better to discuss this in the gentoo-dev ML first.

laumann:
>  * The "Clean patch howto" section is about formatting the patch itself (e.g. how to name it, put comment in it). So this seems misplaced as a subsection there. It should go in its own section instead (I'd suggest the end of the chapter, or before "Clean patch howto").

Fair point. I'll move it, thanks. I might've figured that it's considered part of a "clean" patch :-)

> * I believe that the more common case is/was USE-conditional patching depending on an architecture flag (like `ppc64`). The reasoning may need some tweaking to cover that case too.

> Also, we recommend conditional patching [elsewhere](https://github.com/gentoo/devmanual/blob/master/ebuild-writing/use-conditional-code/text.xml#L42). So this is a policy change. While I'm all for it, it may be better to discuss this in the gentoo-dev ML first.

Yeah, I've linked this in a few places and it pretty much is exactly what this section is advising against :-) So I'm thinking it needs to be reworked if there's an overall agreement that conditional patching should be avoided.
 
> **Edit:** Another good argument against conditional patches may be that patches should be sent upstream, which cannot be done for a conditional patch.

Nice, good suggestion. I'll add this in.
 
> **Edit 2:** In a slightly different context, we have [this comment](https://gitweb.gentoo.org/archive/repo/gentoo-2.git/tree/eclass/eutils.eclass?id=50e8beda904760c773e5c67fdfe8242255e13495#n175) in `eutils.eclass` (now `epatch.eclass`) since 2010: "you should apply patches all the time and let architecture details be detected at configure/compile time". So maybe it isn't a policy change.

Good find, thanks.

I can bring it up on the gentoo-dev ML and ask for opinions/input?

ulm:
> I can bring it up on the gentoo-dev ML and ask for opinions/input?

Yeah, it certainly won't harm.

laumann:
OK, I've addressed some points that @ulm brought up

 - Moved to its own section (above "Clean patch howto")
 - Added a paragraph to say explicitly that, although discouraged, conditional patching can be unavoidable and is thus not a banned practice

akhuettel:
> 
> Also, we recommend conditional patching [elsewhere](https://github.com/gentoo/devmanual/blob/master/ebuild-writing/use-conditional-code/text.xml#L42). So this is a policy change. While I'm all for it, it may be better to discuss this in the gentoo-dev ML first.
> 

I was definitely taught during recruitment that while it's not completely banned it should be avoided as much as possible. Not a policy change. So maybe the devmanual is wrong here?

laumann:
> > Also, we recommend conditional patching [elsewhere](https://github.com/gentoo/devmanual/blob/master/ebuild-writing/use-conditional-code/text.xml#L42). So this is a policy change. While I'm all for it, it may be better to discuss this in the gentoo-dev ML first.
> 
> I was definitely taught during recruitment that while it's not completely banned it should be avoided as much as possible. Not a policy change. So maybe the devmanual is wrong here?

The goal of that page is to show different use cases for USE cases and some conditional logic, which I think is fine. But I think that it could be reworked either to omit the conditional patching (need to come up with other examples) and/or add a note/warning on the page directing the reader to this section on conditional patching.

ulm:
> > Also, we recommend conditional patching [elsewhere](https://github.com/gentoo/devmanual/blob/master/ebuild-writing/use-conditional-code/text.xml#L42). So this is a policy change. While I'm all for it, it may be better to discuss this in the gentoo-dev ML first.
> 
> I was definitely taught during recruitment that while it's not completely banned it should be avoided as much as possible. Not a policy change. So maybe the devmanual is wrong here?

Yeah, see the rest of my comment, regarding `epatch` documentation. Also the argument about upstreaming patches is pretty strong.

I think the conclusion here (and from gentoo-dev discussion) is that we want to strongly discourage conditional patching. However, we recognise that there are situations where it cannot be avoided.

ulm:
> The goal of that page is to show different use cases for USE cases and some conditional logic, which I think is fine. But I think that it could be reworked either to omit the conditional patching (need to come up with other examples) and/or add a note/warning on the page directing the reader to this section on conditional patching.

Right, the example there should be changed. I suggest that we leave that for a separate update and don't complicate this PR with it.

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -112,6 +112,42 @@ from the <c>vim</c> patch tarball:
 </body>
 </section>
 
+<section>
+<title>Conditional Patching</title>
```
ulm, line 116:
```suggestion
<title>Conditional patching</title>
```
@maffblaster is going to change all existing titles to sentence case soon, so I guess newly added ones should also follow that.

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -112,6 +112,42 @@ from the <c>vim</c> patch tarball:
...
+what USE flags and features of the package are for. As such, it is preferable to
+apply patches unconditionally and avoid conditional patching.
+</p>
+<p>
+Consider the following example <c>src_prepare()</c> implementation:
+</p>
+<codesample lang="ebuild">
+src_prepare() {
+	if use threads; then
+		PATCHES+=( "${FILESDIR}/${P}"-mt.patch )
```
ulm, line 130: The quoting looks a little strange to my eyes. Either quote only the variable where quoting is necessary:
```suggestion
		PATCHES+=( "${FILESDIR}"/${P}-mt.patch )
```
or quote the whole string:
```suggestion
		PATCHES+=( "${FILESDIR}/${P}-mt.patch" )
```

ulm (commented):
The reasoning about upstreaming patches should be added, too.

Also allow me some pedantic comments about the formatting. :smile:

ebuild-writing/misc-files/patches/text.xml
```diff
@@ -112,6 +112,42 @@ from the <c>vim</c> patch tarball:
 </body>
 </section>
 
+<section>
+<title>Conditional Patching</title>
+<body>
+<p>
```
ulm, line 118: Please add a blank line before `<p>` and `<codesample>` tags, see our [style guide](https://devmanual.gentoo.org/appendices/devbook-guide/index.html#internal-coding-style).

ebuild-writing/misc-files/patches/text.xml
laumann: Done.

ebuild-writing/misc-files/patches/text.xml
laumann: I went with the second proposal - did you have a preference?

ebuild-writing/misc-files/patches/text.xml
laumann: Done.

ebuild-writing/misc-files/patches/text.xml
ulm: Slight preference for `"${FILESDIR}"/${P}-mt.patch` here, but you'll find both styles in the tree.

laumann:
@ulm @thesamesam Regarding this "conditional patches can't be upstreamed". I think I don't fully understand this.

I get that the patch cannot verbatim be included in upstream, but it could be reworked to be a configuration option for the upstream, right? Or am I totally missing something here?

ulm:
> @ulm @thesamesam Regarding this "conditional patches can't be upstreamed". I think I don't fully understand this.
> 
> I get that the patch cannot verbatim be included in upstream, but it could be reworked to be a configuration option for the upstream, right? Or am I totally missing something here?

If the patch can be reworked, then it should be, and applied in its reworked form. Typically, the ebuild would then apply the patch unconditionally and add an `econf` option in addition.

laumann:
Expanded to include a short list of reasons why conditional patching should be avoid when possible. I'm not 100% on the wording, feedback is very welcome.

laumann:
@thesamesam @ulm ping, any additional input on this?

ulm (approved):

ulm:
Merged.

laumann:
\o/

PR #283, opened 2022-04-15
eclass-writing: doc inherit guards, EXPORT_FUNCTIONS warning

laumann:
Add a section to explain what inherit guards are and their usage, and
add a "rule of thumb" to put EXPORT_FUNCTIONS at the end of eclasses for
max portability.

This turned into a larger re-ordering of the sections so the full jmake
example can showcase all the features discussed in the preceding
sections, like EAPI guard, inherit guard, and EXPORT_FUNCTIONS at the
very end.

### DIFF
--- a/eclass-writing/text.xml
+++ b/eclass-writing/text.xml
@@ -638,11 +638,11 @@ eclass-defined defaults <d/> for example, say we had <c>fnord.eclass</c>:
 </p>
 
 <codesample lang="ebuild">
-EXPORT_FUNCTIONS src_compile
-
 fnord_src_compile() {
 	do_stuff || die
 }
+
+EXPORT_FUNCTIONS src_compile
 </codesample>
 
 <p>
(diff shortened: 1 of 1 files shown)


eclass-writing/text.xml
```diff
@@ -659,6 +659,122 @@ src_compile() {
 }
 </codesample>
 
+<p>
+<!-- TODO
+Eclasses may inherit other eclasses to make use of their functionality, and,
+historically there have been instances of eclasses calling
+<c>EXPORT_FUNCTIONS</c> and then inheriting another eclass. 
+the time, and 
+-->
```
laumann, line 668: i would like to put in a sentence or two to outline just _briefly_ that the interaction between `inherit` and `EXPORT_FUNCTIONS` had been not-fully-defined at some point and an implementation-defined behaviour had been dependend on (as I understand https://bugs.gentoo.org/399039). The conclusion of the paragraph is then to quote the PMS ("must not rely on any particular behaviour when inheriting another eclass after a call to EXPORT_FUNCTIONS"), maybe with a note to say: "hey, just put EXPORT_FUNCTIONS at the end from now on, and everybody's happy".

eclass-writing/text.xml
```diff
@@ -659,6 +659,122 @@ src_compile() {
...
+Old eclasses may put <c>EXPORT_FUNCTIONS</c> in other places, even before
+"inherit". They should <e>not</e> blindly be updated to follow the recommended
+pattern here, as it could result in significant breakage.
+</warning>
+
+</body>
+</section>
+
+<section>
+<title>Handling incorrect usage of an eclass</title>
```
laumann, line 751: This section is not new, just moved up in the file.

eclass-writing/text.xml
```diff
@@ -659,6 +659,122 @@ src_compile() {
...
+that define the same default phase function via <c>EXPORT_FUNCTIONS</c>. If an
+ebuild inherits both as <c>inherit foo bar</c>, then the default phases are
+defined by <c>bar.eclass</c>. If <c>foo.eclass</c> is then modified to inherit
+<c>bar</c> as well, then the ebuild's default functions could suddenly become
+those from <c>foo</c> if <c>EXPORT_FUNCTIONS</c> was placed inside the inherit
+guard.
+</p>
+
+<note>
+The rule of thumb here is: Put the call (if any) to <c>EXPORT_FUNCTIONS</c> at
```
sbraz, line 738:
```suggestion
The rule of thumb here is: put the call (if any) to <c>EXPORT_FUNCTIONS</c> at
```

sbraz (changes requested):

eclass-writing/text.xml
```diff
@@ -659,6 +659,122 @@ src_compile() {
...
+ensure that an eclass can be inherited multiple times and have its functions and
+variables defined only once.
+</p>
+
+<p>
+Although it is not nearly as fatal to omit an inherit guard as opposed to a C
+header guard, it is considered good practice to avoid potential confusing
+behaviour that can arise from an eclass being inherited more than once.
+<!-- FIXME expand with more arguments? -->
+</p>
```
ulm, line 693: An inherit guard is only needed for an eclass that can actually be inherited more than once. Otherwise, omitting it is no problem at all. So I'd avoid the word "fatal" here.

laumann: I removed the paragraph, and reworked the previous one a little. No more "fatal" :-)

eclass-writing/text.xml
laumann: done.

eclass-writing/text.xml
```diff
@@ -659,6 +659,113 @@ src_compile() {
...
+guard.
+</p>
+
+<note>
+The rule of thumb here is: put the call (if any) to <c>EXPORT_FUNCTIONS</c> at
+the very on of an eclass, outside of any inherit guards.
+</note>
+<warning>
+Old eclasses may put <c>EXPORT_FUNCTIONS</c> in other places, even before
+"inherit". They should <e>not</e> blindly be updated to follow the recommended
```
ulm, line 734:
```suggestion
<c>inherit</c>. They should <e>not</e> blindly be updated to follow the recommended
```

eclass-writing/text.xml
```diff
@@ -659,6 +659,113 @@ src_compile() {
...
+ebuild inherits both as <c>inherit foo bar</c>, then the default phases are
+defined by <c>bar.eclass</c>. If <c>foo.eclass</c> is then modified to inherit
+<c>bar</c> as well, then the ebuild's default functions could suddenly become
+those from <c>foo</c> if <c>EXPORT_FUNCTIONS</c> was placed inside the inherit
+guard.
+</p>
+
+<note>
+The rule of thumb here is: put the call (if any) to <c>EXPORT_FUNCTIONS</c> at
+the very on of an eclass, outside of any inherit guards.
```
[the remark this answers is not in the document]
laumann: Thanks, added.

eclass-writing/text.xml
laumann: Done.

eclass-writing/text.xml
```diff
@@ -659,6 +659,114 @@ src_compile() {
 }
 </codesample>
 
+<p>
+Eclasses may inherit other eclasses to make use of their functionality, and
+historically there have been instances of eclasses calling
+<c>EXPORT_FUNCTIONS</c> and then inheriting another eclass. As inherited
+eclasses may also execute <c>EXPORT_FUNCTIONS</c> it was not fully defined which
```
thesamesam, line 666:
```suggestion
eclasses may also execute <c>EXPORT_FUNCTIONS</c>, it was not fully defined which
```

thesamesam (approved):

eclass-writing/text.xml
laumann: done

PR #284, opened 2022-04-23
Indirect dependencies + update finding correct deps

juippis:


general-concepts/dependencies/text.xml
```diff
@@ -781,6 +789,23 @@ There are three kinds of circular dependencies:
...
+<section>
+<title>Indirect dependencies</title>
+<body>
+
+<p>
+Always list each direct dependency that your package needs to build and run
+correctly. Do not rely on dependency chains to meet the dependency
+requirements. For example, a package needs <c>dep1</c> and <c>dep2</c>, but
+<c>dep1</c> also depends on <c>dep2</c>. You might consider just adding
+<c>dep1</c> since it currently pulls <c>dep2</c> too, but in future <c>dep1</c>
```
sbraz, line 804:
```suggestion
<c>dep1</c> since it currently pulls <c>dep2</c> too, but in the future, <c>dep1</c>
```

general-concepts/dependencies/text.xml
```diff
@@ -781,6 +789,23 @@ There are three kinds of circular dependencies:
...
+might drop <c>dep2</c> as a dependency, or make it conditional with USE flags.
+This will then break building your ebuild.
```
sbraz, line 806:
```suggestion
This would then break building your ebuild.
```

general-concepts/dependencies/text.xml
```diff
@@ -781,6 +781,23 @@ There are three kinds of circular dependencies:
...
+</p>
```
Flowdalic, line 807: This is in general a very sensible advice. Even though I believe that this section is motivated by the recent discussion whether or not one should always declare the according `acct-group/user` dependency, and, if added as is to devmanual, will be used as argument that it should be done. I really dnot think that there is any reason to treat acct-user ebuilds as a black box that suddenly could drop the dependency to the according acct-group package (if any), as they are fully under our control. Therefore I would like to suggest to to exempt acct-group packages from this advise, unless they are required as DEPEND (as acct-user.eclass only declares a RDEPEND on the delcared acct-group packages).

sbraz (changes requested):

general-concepts/dependencies/text.xml
juippis: > This is in general a very sensible advice. Even though I believe that this section is motivated by the recent discussion whether or not one should always declare the according `acct-group/user` dependency, 

Yes and no. This pops up every so often, see https://github.com/gentoo/gentoo/pull/25039#discussion_r854811400 from 2 days ago. 


> and, if added as is to devmanual, will be used as argument that it should be done. 

Partially correct. One great thing about policies is, we don't have to debate about things based on our preferences. 

> I really dnot think that there is any reason to treat acct-user ebuilds as a black box that suddenly could drop the dependency to the according acct-group package (if any), as they are fully under our control. Therefore I would like to suggest to to exempt acct-group packages from this advise, unless they are required as DEPEND (as acct-user.eclass only declares a RDEPEND on the delcared acct-group packages).

I personally feel the opposite. Every **exactly** required deps should be listed. Simple as that, no need to make it more complicated or have conditions "do this in case of that". 

Overall yes I do think this is a nice principle to show/teach our contributors, but fear not, I will not go on a witchhunt throughout the tree checking each rdep of acct-* packages even if this is accepted to devmanual. You do you on your packages.

juippis:
Waiting for more feedback, merging on sunday if none.

PR #288, opened 2022-05-20
tools-reference/bash: Update examples to use modern syntax

ulm:


tools-reference/bash/text.xml
```diff
@@ -523,41 +365,25 @@ following are available (lifted from <c>man bash</c>):
...
+    or if file1 exists and file2 does not.
+  </ti>
+</tr>
+<tr>
+  <ti><c>file1 -ot file2</c></ti>
+  <ti>file1 is older than file2, or if file2 exists and file1 does not.</ti>
+</tr>
+<tr>
+  <ti><c>file1 -ef file2</c></ti>
+  <ti>file1 and file2 refer to the same device and inode numbers.</ti>
```
sbraz, line 385: Out of curiosity, why not use the same phrasing as bash's `help test` (file1 is a hard link to file2)? I find it slightly easier to understand.

ulm: Good point. The "inode" wording was from the GNU Info manual.

PR #293, opened 2022-06-04
devbook.xsl: Eclass documentation needs GPL-2 in the footer

ulm:
Eclasses are licensed GPL-2, therefore we need a different footer.
Hardcode this for all sub-pages of eclass-reference/.

Signed-off-by: Ulrich Müller &lt;ulm@gentoo.org&gt;

### DIFF
--- a/devbook.xsl
+++ b/devbook.xsl
@@ -665,7 +665,15 @@
               <small>
                 Gentoo is a trademark of the Gentoo Foundation, Inc.
                 The text of this document is distributed under the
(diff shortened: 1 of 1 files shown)


ulm:
Only one of #293 and #294 will be merged.

arthurzam:
> Only one of #293 and #294 will be merged.

I prefer this version. If one day we need to complicate the rule, we can always revisit it.

ulm:
Let go for this version. We don't want the license to be the author's choice, but it must be `CC-BY-SA-4.0` for all source pages. The converted pages are special, and we need special handling for ebuild(5) and make.conf(5) anyway which we'll still have to convert from *roff to html.

PR #294, opened 2022-06-04
devbook.{rnc,xsl}: Add license element

ulm:
Eclasses are licensed GPL-2, therefore we need a different footer.
Introduce an optional license element which defaults to CC-BY-SA-4.0.

Eclass documentation would then set the license with:
&lt;license&gt;GPL-2&lt;/license&gt;

Signed-off-by: Ulrich Müller &lt;ulm@gentoo.org&gt;

### DIFF
--- a/devbook.rnc
+++ b/devbook.rnc
@@ -18,9 +18,12 @@ start = guide
 
 guide = element guide {
   (attribute root { "true" } | attribute self { text }),
+  license?,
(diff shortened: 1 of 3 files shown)


ulm:
Only one of #293 and #294 will be merged.

ulm:
Closing for now.

PR #298, opened 2022-06-24
appendices/devbook-guide: Document the todo element

ulm:
I have this in a local branch since some time, but I am not even sure if documenting this is a good idea.

Any opinions?


thesamesam:
I think I'd rather we document TODOs on Bugzilla. It's too easy for these to get stale. What do you reckon?

PR #300, opened 2022-07-26
general-concepts/use-flags: make myconf an array instead of string

laumann:

### DIFF
--- a/general-concepts/use-flags/text.xml
+++ b/general-concepts/use-flags/text.xml
@@ -310,17 +310,17 @@ GnuTLS is more featureful than OpenSSL, it is favoured:
 
 <codesample lang="ebuild">
 src_configure() {
-	local myconf
+	local myconf=()
 
 	if use ssl; then
-		myconf+=" --enable-ssl --with-ssl=$(usex gnutls gnutls openssl)"
+		myconf+=( --enable-ssl --with-ssl="$(usex gnutls gnutls openssl)" )
 	else
-		myconf+=" --disable-ssl"
+		myconf+=( --disable-ssl )
 	fi
 
 	econf \
 		# Other stuff
-		${myconf}
+		"${myconf[@]}"
 }
 </codesample>
 


ulm:
What is wrong with the current example?

thesamesam:
1. Arrays are considered prettier for this sort of thing broadly
2. Using an array here is illustrative for newbies as it shows it's possible. The string case is kind of obvious, arrays not so much.

laumann:
@ulm nothing wrong per se, I've just gotten the impression that arrays are more of a de facto default than string manipulation (I'll admit that I also prefer arrays for this usage, but maybe I'm biased because I finally learned how to use bash arrays...)

juippis:
++, this is something I've wanted to write in the devmanual too. But I was looking at putting it under */ebuild-writing/common-mistakes* as missing a backlash (\) will break the argument chain, so using an array is preferred (also prettier). 

(all the examples should obviously be updated too!)

ulm:
The purpose of the example is to illustrate conflicting USE flags. I'd argue that the code should be as simple as possible, in order not to distract the reader. There's no need to use an array (or any other of the more fancy bash concepts) here.

Maybe a specific example for an array could be added elsewhere? It should come with an explanation that in general both scalar or array are fine, and that specific situations (like a parameter containing whitespace) may require an array.

The [`src_configure`](https://devmanual.gentoo.org/ebuild-writing/functions/src_configure/index.html) or [Configuring a package](https://devmanual.gentoo.org/ebuild-writing/functions/src_configure/configuring/index.html) section seem to be the obvious location for it.

thesamesam:
Yeah, it'd fit well there. Notably that section needs refreshing anyway for newer EAPIs.

laumann:
I have no strong opinion on this, the only reason I opened the PR was because it surprised me that the example used strings and _not_ an array :-) So I guess I was distracted by this :P

Regardless of the lesson being taught in any particular section of the devmanual, I'd think that any supporting code and examples should adhere to the general recommendations. And my understanding was that arrays are preferable to strings for flags and configuration parameters. But if there's no stated preference, then either it obviously fine.

> Maybe a specific example for an array could be added elsewhere? It should come with an explanation that in general both scalar or array are fine, and that specific situations (like a parameter containing whitespace) may require an array.

Works for me.

laumann:
Yay or nay on this one?

general-concepts/use-flags/text.xml
```diff
@@ -310,17 +310,17 @@ GnuTLS is more featureful than OpenSSL, it is favoured:
 
 <codesample lang="ebuild">
 src_configure() {
-	local myconf
+	local myconf=()
 
 	if use ssl; then
-		myconf+=" --enable-ssl --with-ssl=$(usex gnutls gnutls openssl)"
+		myconf+=( --enable-ssl --with-ssl="$(usex gnutls gnutls openssl" )
```
SoapGentoo, line 316:
```suggestion
		myconf+=( --enable-ssl --with-ssl="$(usex gnutls gnutls openssl)" )
```

SoapGentoo (changes requested):

laumann:
Good catch @SoapGentoo, thanks!

general-concepts/use-flags/text.xml
laumann: Huh, email replies on PRs don't go on the conversation... TIL.

laumann:
gentle ping here - yay or nay?

ulm:
As I said previously:

> Maybe a specific example for an array could be added elsewhere? It should come with an explanation that in general both scalar or array are fine, and that specific situations (like a parameter containing whitespace) may require an array.
> 
> The [`src_configure`](https://devmanual.gentoo.org/ebuild-writing/functions/src_configure/index.html) or [Configuring a package](https://devmanual.gentoo.org/ebuild-writing/functions/src_configure/configuring/index.html) section seem to be the obvious location for it.

So I think we should leave the example as it is and find a more suitable place to explain arrays.

laumann:
OK.

PR #301, opened 2022-08-17
ebuild-writing/variables: Add section about reserved variables

ulm:


ebuild-writing/variables/text.xml
```diff
@@ -600,55 +718,37 @@ table.
 </p>
 
 </body>
+</subsection>
 </section>
 
 <section>
-<title>ROOT</title>
+<title>Variables reserved for the package manager</title>
```
laumann, line 725: Small nit on the title: it doesn't seem to be reserved variable names, more like restrictions on variable and function names.

ulm: It's "Reserved commands and variables" in PMS. We could simply copy this (then again, it is the variables section).

ebuild-writing/variables/text.xml
```diff
@@ -185,6 +185,57 @@ for these variables when writing ebuilds.
...
+<p>
+Below is an example of an ebuild that uses <c>ROOT</c> in
+<c>pkg_postinst()</c> to conditionally print an error message if an
+old and obsolete configuration file still exists:
+</p>
+
+<codesample lang="ebuild">
+pkg_postinst() {
+	if [[ -e "${ROOT}/etc/oldconfig" ]]; then
+		ewarn "You still have the obsolete config file "
```
sbraz, line 229: Is the space before the quotation mark necessary?
```suggestion
		ewarn "You still have the obsolete config file"
```

ebuild-writing/variables/text.xml
```diff
@@ -568,23 +632,77 @@ not be listed.
...
+</codesample>
+<p>
+To state "exactly one of foo, bar, or baz must be set, but not several":
+</p>
+<codesample lang="ebuild">
+REQUIRED_USE="^^ ( foo bar baz )"
+</codesample>
+<p>
+Note that the last relationship is that of an Exclusive OR (XOR).
+While an XOR could be formed from usual DEPEND syntax,
```
sbraz, line 670: I don't really like "formed from", what do you think of this?
```suggestion
While an XOR could be formed with the usual DEPEND syntax,
```

ulm: Note that this is simply a reordering of sections while the wording was there before.

IMHO we can fix obvious small problems here (like the whitespace in the `ROOT` example), but other unrelated issues should go to their own PR.

BTW, "formed from" is my own wording from [bug 353624 comment 8](https://bugs.gentoo.org/353624#c8). :smile:

thesamesam (approved):

PR #302, opened 2022-08-26
general-concepts/ebuild-revisions: clarify interaction with adding/removing USE flags

thesamesam:
Note that when adding a USE flag where the functionality was previously on
(not off in build system), one must revbump
because a dep was unconditional that now became conditional.

Also, if removing a USE flag where the behaviour is now on, but was previously
always off (not on in build system), one must revbump also,
because people might have never set USE=foo (hence --changed-use wouldn't pick
it up) but their deps have now effectively changed.

Bug: https://github.com/pkgcore/pkgcheck/pull/435

### DIFF
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -103,8 +103,14 @@ of thumb could be used as a guideline:
     (unless it is also a runtime dependency),
   </li>
   <li>
-    adding a new USE flag or removing an existing one (since change
-    in USE flags is going to trigger <c>--changed-use</c> rebuild),
+    adding a new USE flag if it controls a USE-dependency where the
+    functionality was hard-disabled in the build system before,
+  </li>
+  <li>
+    removing an existing USE flag if it controls a USE-dependency where the
+    functionality is now disabled entirely, rather than always being enabled
+    (since the change in USE flags is going to trigger a <c>--changed-use</c>
+    rebuild),
   </li>
   <li>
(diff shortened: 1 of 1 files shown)


thesamesam:
cc @arthurzam

general-concepts/ebuild-revisions/text.xml
```diff
@@ -98,8 +98,11 @@ of thumb could be used as a guideline:
     adding a missing build-time dependency that caused a build failure,
   </li>
   <li>
-    adding a new USE flag or removing an existing one (since change
-    in USE flags is going to trigger <c>--changed-use</c> rebuild),
+    adding a new USE flag (only where the functionality was hard-disabled in
+    the build system before) or removing an existing one (only where the
+    functionality is now disabled entirely, rather than always being enabled)
+    since the change in USE flags is going to trigger <c>--changed-use</c>
+    rebuild,
```
ulm, line 105: Maybe it would be clearer to split this into separate items for adding and removing?

Also, [some say](https://thewritelife.com/when-to-use-parentheses/) that putting important information in parentheses isn't good style.

thesamesam: Yeah, I think you're right. And the parentheses indeed should be for elaboration, not for further conditions.

Thanks, I'll give that a go.

ulm (commented):
The commit message explains it better than the actual text. :smile:

Without mentioning USE-dependencies the meaning is not clear.

thesamesam:
Any better or still not there?

ulm (approved):

PR #303, opened 2022-10-06
Revert "ebuild-maintenance/removal: Process for virtual removal"

thesamesam:
ebuild-maintenance/removal: Change inconsistent virtual revbump policy

This partly reverts commit f28b85f069c70538c618a3cedece77a359601dd8.

There's no reason to treat virtuals differently here, there's no
exception for it in how Portage works, and there's been various
cases of it causing upgrade difficulties for users (including
but in-no-way limited to https://archives.gentoo.org/gentoo-dev/message/64c42804eb4cf4bc7d1161a2e9222c6a).

The last big examples being virtual/emacs & virtual/pam which led
to a long-tail of issues in #gentoo.

It's also led to extensive confusion with contributors & new
developers as we have to try to explain the inconsistency
between devmanual and policy for non-virtuals.

See: https://github.com/gentoo/gentoo/pull/27637

### DIFF
--- a/ebuild-maintenance/removal/text.xml
+++ b/ebuild-maintenance/removal/text.xml
@@ -113,10 +113,8 @@ In order to remove a virtual package, follow the following procedure:
     Wait the time appropriate for the last rites.
   </li>
   <li>
-    Update all ebuilds not to reference the virtual. Since there is
-    no urgent need to remove the virtual from user systems
-    and the resulting rebuilds would be unnecessary, do not bump ebuilds
-    when replacing the dependency.
+    Update all ebuilds not to reference the virtual, following normal
+    <uri link="::general-concepts/ebuild-revisions/"/> policy
   </li>
   <li>
     Remove the package directly
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -108,11 +108,6 @@ of thumb could be used as a guideline:
   <li>
     a dependency change that is a result of a package move (slot move),
   </li>
-  <li>
-    a dependency change due to
-    <uri link="::ebuild-maintenance/removal/#Removing a virtual package">
-    removal of a virtual package</uri>.
-  </li>
 </ul>
 
 </body>


SoapGentoo (approved):
Yes, the marginal gain is offset by all of the dead-ends.

ebuild-maintenance/removal/text.xml
```diff
@@ -113,16 +113,13 @@ In order to remove a virtual package, follow the following procedure:
     Wait the time appropriate for the last rites.
   </li>
   <li>
-    Update all ebuilds not to reference the virtual. Since there is
-    no urgent need to remove the virtual from user systems
-    and the resulting rebuilds would be unnecessary, do not bump ebuilds
-    when replacing the dependency.
+    Update all ebuilds not to reference the virtual with a new revision.
```
ulm, line 116: This is just not very practical for dependencies that are via an eclass.

Also I think the argument about the compile time is still valid.

ebuild-maintenance/removal/text.xml
```diff
@@ -113,16 +113,13 @@ In order to remove a virtual package, follow the following procedure:
...
   </li>
   <li>
-    Remove the package directly
+    Remove the package directly.
   </li>
   <li>
-    Perform the post-removal cleanup, as with regular packages
+    Perform the post-removal cleanup, as with regular packages.
```
ulm, line 122: No full stops there, see the [style guide](https://devmanual.gentoo.org/appendices/devbook-guide/index.html#external-coding-style).

ebuild-maintenance/removal/text.xml
thesamesam: It's no different for other non-virtual dependencies. A new revision is required. If you want to wait to do it until an ew version comes out of your package, that's fine, but you can't just elide the new revision. If you can suggest phrasing, go ahead.

ebuild-maintenance/removal/text.xml
thesamesam: OK.

thesamesam:
> > The last big examples being virtual/emacs & virtual/pam which led to a long-tail of issues in #gentoo.
> 
> Please remind me what were the issues with virtual/emacs removal. Can you point to a bug caused by it?

Conflicts from users who couldn't upgrade in #gentoo.

thesamesam:
> > If I install app-misc/i-get-rarely-updated which depends on virtual/emacs, then several months later, the ebuild is modified to drop the virtual/emacs dependency without a revbump, there's no reason that my system has to believe virtual/emacs is depcleanable. It'll hold onto it (and also then hold on to the deps via virtual/emacs) until it's told not to, by e.g. re-emerging i-get-rarely-updated manually.
> 
> Why would this be a problem? Resource usage of the installed virtual is negligible.

The PAM example is why it matters. It's relying on a hack and it's not guaranteed to work. It also means you have inconsistent behaviour across systems depending on when you installed. And, in time, like with PAM, you end up with upgrades held back if something held on by virtual/emacs ends up conflicting with a newer package.

ulm:
As I said, let's wait for eclass subrevisions then. This is a wider problem and not limited to virtuals.

However, virtuals rather pointedly illustrate the problem, because for them the revbump policy comes with considerable cost (for both developers and users) but close to zero benefit. IMHO.

thesamesam:
Yes, it's not limited to virtuals, and developers should revbump already as the devmanual says for non-virtual cases. This drops the odd exception for virtuals.

I doubt subrevisions for eclasses will be implemented.

ebuild-maintenance/removal/text.xml
```diff
@@ -113,10 +113,7 @@ In order to remove a virtual package, follow the following procedure:
     Wait the time appropriate for the last rites.
   </li>
   <li>
-    Update all ebuilds not to reference the virtual. Since there is
-    no urgent need to remove the virtual from user systems
-    and the resulting rebuilds would be unnecessary, do not bump ebuilds
-    when replacing the dependency.
+    Update all ebuilds not to reference the virtual with a new revision
```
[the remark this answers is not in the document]
thesamesam: Sure, that works. I'll change now.

ebuild-maintenance/removal/text.xml
```diff
@@ -113,17 +113,12 @@ In order to remove a virtual package, follow the following procedure:
     Wait the time appropriate for the last rites.
   </li>
   <li>
-    Update all ebuilds not to reference the virtual. Since there is
-    no urgent need to remove the virtual from user systems
-    and the resulting rebuilds would be unnecessary, do not bump ebuilds
-    when replacing the dependency.
+    Update all ebuilds not to reference the virtual, following normal
+    <uri link="::general-concepts/ebuild-revisions/"/> policy
```
ulm, line 117: As I said, you may want to drop the last bullet point from the [target page](https://devmanual.gentoo.org/general-concepts/ebuild-revisions/index.html), which refers back to this section and says:

> Examples of changes that can be done without a revision bump are:
> - [...]
> - a dependency change due to removal of a virtual package.

thesamesam: I interpreted your comment to mean "the last bullet point in the overall list", as you can see in the diff: https://github.com/gentoo/devmanual/compare/3d8fdf443d4dfc945b48b93bb1451d6bfa8be403..d9683815c7a740c30b5d668b778f40de25a8864f. I'll adjust it later.

ulm (commented):
I could live with this.

ulm:
For reference:
- previous PR #177
- previous ML discussion: https://archives.gentoo.org/gentoo-dev/message/30cc9c087b18e253b915a56ff9ed547c

ulm:
> I could live with this.

Not with the commit message though, because neither of the examples there applies:
- The issues with `virtual/pam` removal arose because it *didn't* follow the policy. (The virtual was package masked while the policy says otherwise.)
- There were no problems with `virtual/emacs` removal. See above.

thesamesam:
> > I could live with this.
> 
> Not with the commit message though, because neither of the examples there applies:
> 
>     * The issues with `virtual/pam` removal arose because it _didn't_ follow the policy. (The virtual was package masked while the policy says otherwise.)
> 
>     * There were no problems with `virtual/emacs` removal. See above.

Please suggest exactly what you'd like to see.

ulm:
> > > I could live with this.
> > 
> > 
> > Not with the commit message though, because neither of the examples there applies:
> > ```
> > * The issues with `virtual/pam` removal arose because it _didn't_ follow the policy. (The virtual was package masked while the policy says otherwise.)
> > 
> > * There were no problems with `virtual/emacs` removal. See above.
> > ```
> 
> Please suggest exactly what you'd like to see.

Just say that you want the same policy for virtuals as for regular packages. IMHO nothing else of the rationale in the commit message would hold up to scrutiny. (AFAICS also [the issue quoted in the first paragraph](https://archives.gentoo.org/gentoo-dev/message/64c42804eb4cf4bc7d1161a2e9222c6a) was caused by `virtual/pam` being masked contrary to policy.)

thesamesam:
OK.

PR #305, opened 2022-11-08
general-concepts/ebuild-revisions: note which build-time fixes need a…

thesamesam:
… new revision

### DIFF
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -89,7 +89,7 @@ of thumb could be used as a guideline:
     (e.g. <c>-Wformat</c> or <c>-Wimplicit-function-declaration</c> fixes);
     the package may have been miscompiled, or the change is substantial
     (if adding a huge patch to fix a problem, the chances of an unexpected
-    issue being introduced by it are larger).
+    issue being introduced by it are greater).
   </li>
   <li>adding a trivial documentation fix,</li>
   <li>
@@ -99,7 +99,8 @@ of thumb could be used as a guideline:
     is expected soon),
   </li>
   <li>
-    adding a missing build-time dependency that caused a build failure,
+    adding a missing build-time dependency that caused a build failure
+    (unless it is also a runtime dependency),
   </li>
   <li>
     adding a new USE flag or removing an existing one (since change
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -85,7 +85,11 @@ of thumb could be used as a guideline:
   <li>
     adding a patch to fix a build-time issue that prevented users from
     building the package (since it does not affect users who already
-    managed to build it),
+    managed to build it) unless: it affected runtime behaviour in some way
+    (e.g. <c>-Wformat</c> or <c>-Wimplicit-function-declaration</c> fixes);
+    the package may have been miscompiled, or the change is substantial
+    (if adding a huge patch to fix a problem, the chances of an unexpected
+    issue being introduced by it are larger).
   </li>
   <li>adding a trivial documentation fix,</li>
   <li>


ulm (commented):
Otherwise looks good. My main worry is that we should keep the section reasonably short, without going too much into micro-management.

general-concepts/ebuild-revisions/text.xml
```diff
@@ -85,7 +85,11 @@ of thumb could be used as a guideline:
   <li>
     adding a patch to fix a build-time issue that prevented users from
     building the package (since it does not affect users who already
-    managed to build it),
+    managed to build it) unless: it affected runtime behaviour in some way
+    (e.g. <c>-Wformat</c> or <c>-Wimplicit-function-declaration</c> fixes);
```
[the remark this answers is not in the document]
thesamesam: It's not purely cosmetic (despite it often seeming so, although some cases are):
- e.g. If a function returns a _Bool on amd64 but we assume it's an int, half of its bits get cleared.
- e.g. On arm64 (for Apple macs at least, with their different calling convention/ABI), there's been runtime corruption with passing arguments.

There's also been e.g. https://developers.redhat.com/blog/2019/04/22/implicit-function-declarations-flexs-use-of-reallocarray.

Florian covers it in his short talk at https://youtu.be/q5itHU2T5xU?t=2583 (but there's also a great LWN article: https://lwn.net/SubscriberLink/913505/b61bbc0c4296117f/).

All this to say: I wanted to call it out specifically because it's one of those things where something can _look_ like a pure build time fix, but actually affects code generation.

thesamesam:
I take your point about the section becoming a bit big. Would it be okay to keep this as-is, but I'll have a think about if we can rework it at all? My worry about moving the current information to be added is that folks might miss it.

I've seen a lot of instances of "build time fix => no revbump" even when it's obvious a revbump is needed (-Wformat-security or something) :(

general-concepts/ebuild-revisions/text.xml
```diff
@@ -95,7 +99,11 @@ of thumb could be used as a guideline:
     is expected soon),
   </li>
   <li>
-    adding a missing build-time dependency that caused a build failure,
+    adding a missing build-time dependency that caused a build failure that
+    is not a runtime dependency (e.g. a missing library dependency which
+    happened to fail in src_configure still needs a new revision because
+    it would manifest as an automagic dependency if a user had the package
+    installed already),
```
ulm, line 106: [For some reason, my previous comment disappeared.]

I think these examples aren't really necessary, and may distract from more important things (i.e. make the section too long so that people won't read it thoroughly any more). I also dislike the double "that". How about the following:
```
  <li>
    adding a missing build-time dependency that caused a build failure
    (unless it is also a runtime dependency),
  </li>
```

(Thinking about it, we probably should spell "build-time" and "run-time" consistently, i.e. both with or without a hyphen.)

ulm:
> I take your point about the section becoming a bit big. Would it be okay to keep this as-is, but I'll have a think about if we can rework it at all? My worry about moving the current information to be added is that folks might miss it.

That's fine for this item, but see my other comment (which was lost for some reason).

general-concepts/ebuild-revisions/text.xml
juippis: Wonder if there's room for a whole new section at the bottom of the page; concrete examples with rationale explained in detail (like it's not **really** explained why touching anything-RDEPEND always requires a revbump - at least in this page)

Or if such information(s) already exists, you could maybe add some sort of `[1]` internal links to these short sections.

Just an idea.

thesamesam: Yeah, I think that's a good point, although I'm not sure I've got it in me right now to start the new section. It's definitely something we should do though.

thesamesam:
Merged.

PR #306, opened 2022-11-21
ebuild-maintenance/git: update grafted repo URL

thesamesam:

### DIFF
--- a/ebuild-maintenance/git/text.xml
+++ b/ebuild-maintenance/git/text.xml
@@ -90,9 +90,9 @@ graft it into the repository:
 </p>
 
 <pre>
-git remote add history https://anongit.gentoo.org/git/repo/gentoo/historical.git/
+git remote add history https://anongit.gentoo.org/git/archive/repo/gentoo-2.git
 git fetch history
-git replace --graft 56bd759df1d0c750a065b8c845e93d5dfa6b549d history/master
+git replace --graft 56bd759df1d0c750a065b8c845e93d5dfa6b549d cvs-final-2015-08-08
 </pre>
 
 <p>


ebuild-maintenance/git/text.xml
```diff
@@ -90,7 +90,7 @@ graft it into the repository:
 </p>
 
 <pre>
-git remote add history https://anongit.gentoo.org/git/repo/gentoo/historical.git/
+git remote add history https://anongit.gentoo.org/git/archive/repo/gentoo-2.git
 git fetch history
 git replace --graft 56bd759df1d0c750a065b8c845e93d5dfa6b549d history/master
```
ulm, line 95: Thank you. I had updated the [wiki page](https://wiki.gentoo.org/wiki/Gentoo_git_workflow#Grafting_Gentoo_history_onto_the_active_repository) but apparently I missed this one.

The master branch in the historical repo contains one additional testing commit which shouldn't be included. There's a tag that can be used to avoid this:
```suggestion
git replace --graft 56bd759df1d0c750a065b8c845e93d5dfa6b549d cvs-final-2015-08-08
```

thesamesam: Ah, neat!

sbraz:
@thesamesam what does this add compared to the existing git conversion? The repo description ("Another conversion of the historical gentoo-x86 ebuild repository") isn't very explicit.

thesamesam:
> @thesamesam what does this add compared to the existing git conversion? The repo description ("Another conversion of the historical gentoo-x86 ebuild repository") isn't very explicit.

I believe there were a bunch of files that got lost with the original conversion. It was first announced at https://marc.info/?l=gentoo-dev&m=163008592103426&w=2 but it's not explicit about the actual rationale for the migration.

ulm:
> I believe there were a bunch of files that got lost with the original conversion. It was first announced at https://marc.info/?l=gentoo-dev&m=163008592103426&w=2 but it's not explicit about the actual rationale for the migration.

There are some notes about the difference in the [README file](https://gitweb.gentoo.org/dev/ulm.git/tree/README?h=gentoo-x86-conversion#n87) in my conversion repo. (It was linked in the announcement, but unfortunately marc.info has mangled the URL.)

PR #307, opened 2023-04-27
function-reference/install-functions: add note about paths to dostrip

matoro:
Found this extremely unintuitive especially compared to the documentation on the `dosym` function which is much clearer.

Closes: https://bugs.gentoo.org/888817

### DIFF
--- a/function-reference/install-functions/text.xml
+++ b/function-reference/install-functions/text.xml
@@ -431,6 +431,7 @@ The <c>*into</c> functions create the directory if it does not already exist.
       Normally executed to exclude from stripping.
       Eg. <c>dostrip -x /path/to/important.so</c>.  May also be used to include binaries
       to strip when <c>RESTRICT=strip</c> without the -x option.
+      Provided paths are relative to <c>${ED}</c>, even if they begin with a slash.
     </ti>
   </tr>
 </table>


function-reference/install-functions/text.xml
```diff
@@ -431,6 +431,8 @@ The <c>*into</c> functions create the directory if it does not already exist.
       Normally executed to exclude from stripping.
       Eg. <c>dostrip -x /path/to/important.so</c>.  May also be used to include binaries
       to strip when <c>RESTRICT=strip</c> without the -x option.
+      Note that in order to recursively exclude directories from stripping, the absolute
+      path including <c>${ED}</c> must be provided, as opposed to a relative path.
```
ulm, line 435: I don't think that's correct.
`dostrip` automatically prefixes its path with `${ED}`, see [PMS section 12.3.11](https://projects.gentoo.org/pms/8/pms.html#x1-13100012.3.11):

> Whether an item is to be excluded is determined as follows: For each item in the exclusion list, pretend it has the value of the ED variable prepended, then:
> - If it is a directory, act as if every file or directory immediately under this directory were in the exclusion list.
> - If the item is a file, the operation shall not be performed on it.
> - If the item does not exist, it is ignored.

matoro: Oops, I just reread ionen's comment and looked at the ebuild in question I pulled this from, you are right.  What this was meant to say is that you must use an *absolute path relative to `${ED}`*.

function-reference/install-functions/text.xml
```diff
@@ -431,6 +431,8 @@ The <c>*into</c> functions create the directory if it does not already exist.
       Normally executed to exclude from stripping.
       Eg. <c>dostrip -x /path/to/important.so</c>.  May also be used to include binaries
       to strip when <c>RESTRICT=strip</c> without the -x option.
+      Note that provided paths must be absolute relative to <c>${ED}</c>, i.e., as if
+      <c>${ED}</c> was the filesystem root.
```
ulm, line 435: Two small points:
- "Absolute relative" is an oxymoron :smile:. It really is a path _relative_ to `${ED}`.
- I think we don't put commas after `i.e.` because the devmanual is in British English.
```suggestion
      Note that provided paths must be relative to <c>${ED}</c>, i.e. as if
      <c>${ED}</c> was the filesystem root.
```

Maybe also change the example above to something other than `/path/to/important.so` which isn't very helpful. Something starting with `/usr` or `/opt` might be better there.

matoro: I don't think it's an oxymoron because in this case `dostrip -x /usr/lib/node_modules` works, but `dostrip -x usr/lib/node_modules` does not.  I would consider the second a "relative path".  Is there some sort of better phrasing you can think of to call the former?

ulm: How about the following wording for the whole item?
```
      Controls stripping of executables. Normally used to exclude from
      stripping, e.g. <c>dostrip -x /usr/$(get_libdir)/important.so</c>.
      Without the <c>-x</c> option, may also be used to include binaries to
      strip when <c>RESTRICT=strip</c>. Note that <c>${ED}</c> is prepended
      to provided paths, i.e. as if <c>${ED}</c> was the filesystem root.
```

Flowdalic: Assuming I understand this correctly, then what about adding the follow to what ulm suggests:

> This means that the path argument of dostrip must start with a leading slash ('/'), but the path itself is relativ to ED.

ulm: That's not correct either, as it works without the leading slash (as specified by the PMS). It is implemented like this in Portage (in phase-helpers.sh):
```
			[[ ${f:0:1} = / ]] || f="/${f}"
```

matoro: Sounds good, went with this phrasing.

function-reference/install-functions/text.xml
```diff
@@ -431,6 +431,7 @@ The <c>*into</c> functions create the directory if it does not already exist.
       Normally executed to exclude from stripping.
       Eg. <c>dostrip -x /path/to/important.so</c>.  May also be used to include binaries
       to strip when <c>RESTRICT=strip</c> without the -x option.
+      Provided paths are relative to <c>${ED}</c>, even if they being with a slash.
```
ulm, line 434: Typo:
```suggestion
      Provided paths are relative to <c>${ED}</c>, even if they begin with a slash.
```

Flowdalic (approved):
The added sentence is certainly an improvement. Thanks :)

That said, the whole description could be improved. But that is, of course, unrelated to your work.

ulm:
Thank you!

PR #308, opened 2023-04-29
function-reference/install-functions: Reword dostrip documentation

ulm:
As a followup to the previous change, use a better path for the example and reword. No longer mention EAPI 7 because all older EAPIs are deprecated.

### DIFF
--- a/function-reference/install-functions/text.xml
+++ b/function-reference/install-functions/text.xml
@@ -427,11 +427,11 @@ The <c>*into</c> functions create the directory if it does not already exist.
       <c>dostrip</c>
     </ti>
     <ti>
-      Introduced with EAPI=7, controls the stripping of executables.
-      Normally executed to exclude from stripping.
-      Eg. <c>dostrip -x /path/to/important.so</c>.  May also be used to include binaries
-      to strip when <c>RESTRICT=strip</c> without the -x option.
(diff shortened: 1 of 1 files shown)


function-reference/install-functions/text.xml
```diff
@@ -427,11 +427,11 @@ The <c>*into</c> functions create the directory if it does not already exist.
...
     <ti>
-      Introduced with EAPI=7, controls the stripping of executables.
-      Normally executed to exclude from stripping.
-      Eg. <c>dostrip -x /path/to/important.so</c>.  May also be used to include binaries
-      to strip when <c>RESTRICT=strip</c> without the -x option.
-      Provided paths are relative to <c>${ED}</c>, even if they begin with a slash.
+      Controls stripping of executables. Normally used to exclude from
+      stripping, e.g. <c>dostrip -x /usr/$(get_libdir)/important.so</c>.
+      May also be used without the <c>-x</c> option to include binaries
+      to strip when <c>RESTRICT=strip</c>. Provided paths are relative to
```
sbraz, line 433:
```suggestion
      to strip when <c>RESTRICT=strip</c> is set. Provided paths are relative to
```

ulm: I was reading it as "RESTRICT is equal to strip", but your suggestion is better.

PR #311, opened 2023-08-19
general-concepts/ebuild-revisions: License updates may require a revbump

ulm:


thesamesam (approved):

thesamesam:
I'd maybe mention the CMake example in the commit message so we don't wonder why we did this in a few years time, although it is kind of obvious. But lgtm thanks!

ulm:
> I'd maybe mention the CMake example in the commit message so we don't wonder why we did this in a few years time, although it is kind of obvious. But lgtm thanks!

Like the old [main page](https://github.com/gentoo/devmanual/blob/historical/text.rst) used to say: :smile:
```
A policy is a restrictive document to prevent a recurrence of a single
incident, in which that incident is never mentioned.

-- Kaufman's Law
```

Anyway, example added to the commit message.

PR #312, opened 2023-09-27
general-concepts/dependencies: use <c> tag for dev-util/ebuildtester

arkamar:
This change highlights the `ebuildtester` package in a manner consistent with how other packages are presented throughout the document.


ulm:
Thanks.

PR #316, opened 2023-10-21
general-concepts/copyright-policy: Clarify general guideline

ulm:
Use nested ordered lists instead of one description list. Hopefully, this will make the logic better understandable.

Add a diagram with a flowchart. Also commit the SVG output, so we don't need Graphviz as build dependency for a single file that isn't expected to change very often.


thesamesam (approved):
I like it - thanks!

ulm:
Currently blocked by `tidy` failing in `make check`:
```
Failed on general-concepts/copyright-policy/index.html:
line 128 column 1 - Warning: <ol> attribute "type" not allowed for HTML5
line 154 column 5 - Warning: <ol> attribute "type" not allowed for HTML5
line 170 column 9 - Warning: <ol> attribute "type" not allowed for HTML5
```

This isn't a problem of the document, but a bug in `tidy`:
https://github.com/htacg/tidy-html5/issues/1063

PR #317, opened 2023-10-23
[WIP] use pkgcore for eclassdoc -> devbook conversion

arthurzam:
See https://github.com/pkgcore/pkgcore/pull/346 for more info.

There are multiple ugly parts, and I know it. Makefiles aren't my forte.


ulm:
See https://codeberg.org/gentoo/devmanual/pulls/5 for a follow-up on this.

PR #319, opened 2023-10-27
devbook.xsl: Warn about missing terminating slash in internal links

ulm:
Fix all internal links accordingly.


thesamesam:
Thanks for doing this.

PR #325, opened 2024-01-14
eclass-writing: @USAGE is not required

ulm:


orlitzky:
Should the description say something like "Can be omitted (only) if the function does not take any arguments."?

ulm:
> Should the description say something like "Can be omitted (only) if the function does not take any arguments."?

We don't say that for other tags like `@RETURN` either.

orlitzky (approved):
OK, LGTM then, thanks.

PR #326, opened 2024-01-21
package-moves: Allow moving back, and describe how to chain updates

mgorny:
Make the rule forbidding reusing old package names less strict.  There is no real reason not to move packages back — we can safely assume that the package manager will handle that just fine.  The only real restriction is that no other package must reuse any of the old names, i.e. effectively cause the package manager to attempt merging the packages together.

Add explicit rules for updating old package move entries.  These match the logic used in pkgcheck, and therefore ensure both correct results and no warnings from pkgcheck.

### DIFF
--- a/ebuild-maintenance/package-moves/text.xml
+++ b/ebuild-maintenance/package-moves/text.xml
@@ -19,8 +19,10 @@ the following must be noted:
   </li>
   <li>
     Once an update entry is created, the old package name (or slot) cannot be
-    reused. Attempting to reuse it will cause updates to apply again,
-    to the reused name. This also means that updates cannot be undone.
+    reused for another package. Attempting to reuse it will cause updates to
+    apply again, to the reused name. This also means that while a package can
+    be moved back to its previous name, all the names historically used for
+    a package are reserved to it alone.
   </li>
   <li>
     Updates can only perform one-to-one moves. They cannot be used to merge
@@ -112,6 +114,49 @@ Make sure that you have fixed all the reverse dependencies and that
 you have updated every file in <c>profiles/</c> directory that
 happens to contain an entry which may be affected by your change.
 </p>
+
+</body>
+</section>
+
+<section>
+<title>Updating prior update entries when moving the package again</title>
+<body>
+
+<p>
+When the same package is moved again, the previous update entries should
+be updated appropriately. This is meant to make the situation more transparent
+to users reading update entries and to ensure that the process is handled
+efficiently even if the package manager does not implement updates in a robust
+way. This involves the following steps:
+</p>
+
+<ol>
+  <li>
+    The previous package moves for the package in question must be updated
+    to reference the final name. That is, rather than the chain A → B → C,
+    we want to have two update entries: A → C, and B → C.
+  </li>
+
+  <li>
+    If the package is being moved to a name that it used before, the original
+    move entry must be removed. That is, rather than the chain A → B → A,
+    we want to have the reverse entry: B → A. If the package manager did not
+    move A → B before, we don't want it to touch the package at all.
+  </li>
+
+  <li>
+    As a combination of the two aforementioned steps, a chain of A → B → C → A
+    would be replaced by two moves: B → A, and C → A.
+  </li>
+
+  <li>
+    If the package was slot-moved before, the slot moves should be updated
+    to use the final package name, and moved after the final package move.
+    That is, rather than the chain: A:S₁ → A:S₂, A → B; we prefer to have
+    the chain: A → B, B:S₁ → B:S₂. All package and slot move entries must reside
+    in the same file then, to guarantee sequential processing.
+  </li>
+</ol>
 
 </body>
 </section>


thesamesam:
Does the [relevant section](https://dev.gentoo.org/~ulm/pms/head/pms.html#section-4.4.4) of PMS need clarification?

It says:
> Any name that has appeared as the origin of a move must not be reused in the future

I suppose it doesn't say "has ever".

sbraz (approved):

mgorny:
> Does the [relevant section](https://dev.gentoo.org/~ulm/pms/head/pms.html#section-4.4.4) of PMS need clarification?
> 
> It says:
> 
> > Any name that has appeared as the origin of a move must not be reused in the future
> 
> I suppose it doesn't say "has ever".

Probably yes. I suppose it's going to be hard to formally convey the right meaning.

arthurzam (approved):
Easy to read and understand (even for me), and then examples make it excellent.

I also want to request verification if the `→` symbol is rendered correctly.

ebuild-maintenance/package-moves/text.xml
```diff
@@ -113,6 +115,48 @@ you have updated every file in <c>profiles/</c> directory that
...
+  <li>
+    As a combination of the two aforementioned steps, a chain of A → B → C → A
+    would be replaced by two moves: B → A, and C → A.
+  </li>
+
+  <li>
+    If the package was slot-moved before, the slot moves should be updated
+    to use the final package name, and moved after the final package move.
+    That is, rather than the chain: A:S₁ → A:S₂, A → B; we prefer to have
+    the chain: A → B, B:S₁ → B:S₂.
```
ulm, line 156: The order in which moves are executed isn't specified by PMS. For example, there's no defined order for A → B in `4Q-2023` and B:S₁ → B:S₂ in `1Q-2024`.

ulm (commented):
This would explicitly contradict what is (currently) specified by PMS.

Is this just a theoretical exercise, or a real problem we need to solve? Package moves are rare, and moving packages back to their previous name practically never happens.

ebuild-maintenance/package-moves/text.xml
mgorny: Not even if they're in the same file? Should we specify slotmoves for both old and new package name then?

mgorny:
> Is this just a theoretical exercise, or a real problem we need to solve? Package moves are rare, and moving packages back to their previous name practically never happens.

It does happen. And when it does, I think it makes sense to be explicit on how to make it work without stuff accidentally exploding in user's face.

ulm:
We had a previous discussion in [bug 296713](https://bugs.gentoo.org/296713), where the conclusion was that mistakes can be corrected before they hit the rsync mirrors. Apart from that case, moves are pretty much permanent and cannot be reverted.

Are we sure that this change won't open up possible undefined behaviour, especially when considering multiple repositories and binary packages?

ebuild-maintenance/package-moves/text.xml
ulm: Within the same file should be fine, so maybe explicitly say so?
https://archives.gentoo.org/gentoo-pms/message/4fde4acef1dd0c85e82f23dfb5ae0322

mgorny: Updated to say that all entries must be in the same file.

mgorny:
> Are we sure that this change won't open up possible undefined behaviour, especially when considering multiple repositories and binary packages?

I can't think of anything worse than what we have right now. The worst mess would be if overlay copies our `pkgmove` entry so ::gentoo would specify 'B → A' and the overlay 'A → B'. However, there's really no reason why'd they do that, and we can't protect against them doing something like that anyway.

My understanding is that if you move package back, and remove the original move entry, then:
- packages (installed or binary) that didn't have the original move applied won't have any changes done to them (which is fine)
- packages that have the original move applied will be moved back

ulm:
What do we do for PMS (which I'd rather not expand too much)? Would something like the following be sufficient?
```
-Any name that has appeared as the origin of a move must not be reused in the future.
+Any name that has ever appeared as the origin of a move must not be reused by another package in the future.
```

thesamesam:
> What do we do for PMS (which I'd rather not expand too much)? Would something like the following be sufficient?
> 
> ```
> -Any name that has appeared as the origin of a move must not be reused in the future.
> +Any name that has ever appeared as the origin of a move must not be reused by another package in the future.
> ```

Wouldn't this imply we can never cleanup `updates/`, or at least must have `pkgcheck` check git history or similar, to avoid accidental reuse by e.g. restoring a last-rited package, or adding a new package with an ancient name?

mgorny:
> What do we do for PMS (which I'd rather not expand too much)? Would something like the following be sufficient?
> 
> ```
> -Any name that has appeared as the origin of a move must not be reused in the future.
> +Any name that has ever appeared as the origin of a move must not be reused by another package in the future.
> ```

That would work for me. Though TBH I think PMS is stepping too much into the "policy" area here. Perhaps it would be cleaner to explain what the consequences would be.

ulm:
Thinking about it, I'd also be happy to revert https://gitweb.gentoo.org/proj/pms.git/commit/?id=3ba90e4a4918cfe820bb7c627bef2fc0d5a9fed5 and leave the policy for the devmanual.

thesamesam (approved):
OK with me.

wrt PMS: I feel like we still need to say something about not reusing names (we don't want conflicting updates entries?), just not about how long they are there for? Or do we have something for that already?

ulm:
> wrt PMS: I feel like we still need to say something about not reusing names (we don't want conflicting updates entries?), just not about how long they are there for? Or do we have something for that already?

How about this?
"At any given time, a name that appears as the origin of a move may not be used as a qualified package name in the repository. A slot that appears as the origin of a slot move may not be used by packages matching the spec of that slot move."

thesamesam:
perfect, thanks!

mgorny:
Yes, that sounds about right. Also clearly splits specification from policy.

ebuild-maintenance/package-moves/text.xml
```diff
@@ -113,6 +115,49 @@ you have updated every file in <c>profiles/</c> directory that
...
+  <li>
+    As a combination of the two aforementioned steps, a chain of A → B → C → A
+    would be replaced by two moves: B → A, and C → A.
+  </li>
+
+  <li>
+    If the package was slot-moved before, the slot moves should be updated
+    to use the final package name, and moved after the final package move.
+    That is, rather than the chain: A:S₁ → A:S₂, A → B; we prefer to have
+    the chain: A → B, B:S₁ → B:S₂. All package and slot move entries must reside
```
ulm, line 156: I think that Unicode superscript and subscript characters should be avoided. They are a workaround for plain text but have no place in a markup language like XML or HTML. (In other words, a `1` is a `1` regardless of its appearance in regular text or in a subscript.)

ulm:
Merged (with the suggested markup change) in commit 89b2bf7f24c04546bc9d25177ee7739f528f872c.

PR #329, opened 2024-04-13
Rename epatch chapter and update its wording

ulm:


ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,75 @@
...
+    <c>eapply</c> will not unpack patches for you.
+  </li>
+  <li>
+    The default patch level is <c>-p1</c>. Other patch levels must be specified
+    manually or the command will fail.
+  </li>
+  <li>
+    When specifying a directory, at least one file with a name ending in
+    <c>.patch</c> or <c>.diff</c> must exist or the command fails. Other files
+    are ignored.
```
mgorny, line 28: This sounds a bit backwards. Perhaps start by saying that it applies files that …

Also, perhaps it's worth mentioning that it's not recursive.

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,75 @@
...
+  </li>
+</ul>
+
+<p>
+Note that distributing modified tarballs rather than a vanilla tarball
+and patches is <e>highly</e> discouraged.
```
mgorny, line 34: Again, it would reverse it. "Is is highly preferable to distribute patches and apply them over unpacked sources, rather than distribute an already patched tarball."

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<guide self="ebuild-writing/functions/src_prepare/eapply/">
+<chapter>
+<title>Patching with eapply</title>
+<body>
+
+<p>
+The canonical way of applying patches in ebuilds is to use the package
+manager's <c>eapply</c> function.
```
mgorny, line 9: and perhaps "… or the `PATCHES` array combined with `default_src_prepare`".

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,75 @@
...
+</p>
+
+<p>
+<c>eapply</c> was added in EAPI=6. It differs from the previously available
+<c>epatch</c> in several ways:
```
mgorny, line 14: While this is kinda useful, it feels a bit backwards to start by telling the differences. I think we should focus on what the function does (as an information for new users), and perhaps leave differences to the old function in a separate section at the end.

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,75 @@
...
+</ul>
+
+<p>
+Note that distributing modified tarballs rather than a vanilla tarball
+and patches is <e>highly</e> discouraged.
+</p>
+</body>
+
+<section>
+<title>Basic <c>eapply</c></title>
```
mgorny, line 39: Maybe something like "Applying straightforward patches".

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,75 @@
...
+	eapply -p2 "${WORKDIR}/${P}-suse-update.patch"
+	eapply -p0 "${FILESDIR}/${PV}-no-TIOCGDEV.patch"
+	eapply "${FILESDIR}/${PV}-gcc-6.patch"
+	eapply_user
+}
+</codesample>
+</body>
+</section>
+</chapter>
+</guide>
```
mgorny, line 75: Neither of the examples cover applying patch directories.

ulm:
I retract this.

PR #331, opened 2024-04-21
Drop epatch command and rework eapply explanation

ulm:
Take 2, completely reworked (instead of #329 which was too simplistic).

### DIFF
--- a/ebuild-writing/functions/src_prepare/eapply/text.xml
+++ b/ebuild-writing/functions/src_prepare/eapply/text.xml
@@ -1,54 +1,113 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <guide self="ebuild-writing/functions/src_prepare/eapply/">
 <chapter>
-<title>Patching with epatch and eapply</title>
-
+<title>Patching with eapply</title>
 <body>
-<p>
-The canonical way of applying patches in ebuilds is to
-use <c>epatch</c> (from <c>epatch.eclass</c>, which you must make sure
-to inherit!) inside <c>src_prepare</c>. This function automatically
-handles <c>-p</c> levels, <c>gunzip</c> and so on as necessary.
-</p>
 
 <p>
-Starting with EAPI=7, this function is banned and <c>eapply</c> must be used.
+The canonical way of applying patches in ebuilds is to use the package
+manager's <c>eapply</c> command, either by calling it explicitly, or by
+assigning the <c>PATCHES</c> variable supported by the default
+<c>src_prepare</c> implementation.
 </p>
 
+<important>
+Applying patches to the sources from the upstream tarball is <e>strongly</e>
+preferred to distributing your own modified tarball.
+</important>
+
 <p>
-Beginning with EAPI=6, a new function <c>eapply</c> was added to apply patches
-without the need for an eclass.
-This function differs from epatch in several ways:
+The <c>eapply</c> command takes one or more regular file or directory paths as
+its arguments. Optionally, these can be preceded by GNU <c>patch</c> options.
 </p>
+
+<note>
+The <c>--</c> delimiter indicates the end of options. This is useful if a
+filename begins with a hyphen.
+</note>
 
 <ul>
-<li><c>eapply</c> will not unpack patches for you.</li>
-<li>
-The default patch level is -p1.
-Other patch levels must be specified manually or the command will fail.
-</li>
-<li>
-When specifying a directory, at least one file with a name ending in .patch or .diff
-must exist or the command fails.  Other files are ignored.
-</li>
+  <li>
+    If an argument is a regular file, it will be applied it in the working
+    directory by calling GNU <c>patch</c> with patch level <c>-p1</c>.
+    Specifying an explicit <c>-p<e>N</e></c> option will override the default
+    patch level.
+  </li>
+  <li>
+    For a directory, <c>patch -p1</c> applies all patch files with names ending
+    in <c>.diff</c> or <c>.patch</c> in that directory, in POSIXbetical order
+    of their names. Any other files in the directory are ignored.
+    Again, <c>-p<e>N</e></c> can be used to override the default patch level.
+    Note that <c>eapply</c> will not recurse into subdirectories.
+  </li>
 </ul>
 
 <p>
-Note that distributing modified tarballs rather than a vanilla tarball
-and patches is <e>highly</e> discouraged.
+<c>eapply</c> was added in EAPI 6. It differs from the previously available
+<c>epatch</c> in several ways:
 </p>
+
+<ul>
+  <li>
+    <c>eapply</c> will not unpack patches for you.
+  </li>
+  <li>
+    The patch level is no longer detected automatically. Patch levels other
+    than <c>-p1</c> must be specified manually.
+  </li>
+  <li>
+    When specifying a directory, at least one file with a name ending in
+    <c>.diff</c> or <c>.patch</c> must exist or the command fails.
+  </li>
+</ul>
 </body>
 
 <section>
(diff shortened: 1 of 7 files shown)


ebuild-writing/functions/src_prepare/eapply/text.xml~
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
thesamesam, line 1: Oops :)

thesamesam (approved):
LGTM once backup file is removed.

ebuild-writing/functions/src_prepare/eapply/text.xml~
ulm: Fixed, and `*~` added to `.gitignore`.

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,123 @@
...
+The canonical way of applying patches in ebuilds is to use the package
+manager's <c>eapply</c> command, either by calling it explicitly, or by
+assigning the <c>PATCHES</c> variable supported by the default
+<c>src_prepare</c> implementation.
+</p>
+
+<important>
+Applying patches to the sources from the vanilla tarball is <e>strongly</e>
+preferred to distributing a modified tarball.
+</important>
```
mgorny, line 17: To be honest, I still find this a little bit confusing, but perhaps because I'm in the position where applying patches is the "obvious" solution. Perhaps make it clear that we're talking about distributing patches vs. creating our own tarball. We don't want to prohibit users from using upstream "patched" tarballs (such as newer kernel sources vs. taking 5.10.0 along with 200+ patches).

While at it, I wonder if we should also point out that for big patches, we prefer hosting them externally and fetching via `SRC_URI` rather than hogging up `FILESDIR`.

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,123 @@
...
+    directory by calling GNU <c>patch</c> with patch level <c>-p1</c>.
+    Specifying an explicit <c>-p<e>N</e></c> option will override the default
+    patch level.
+  </li>
+  <li>
+    For a directory, <c>patch -p1</c> applies all patch files with names ending
+    in <c>.diff</c> or <c>.patch</c> in that directory, in POSIXbetical order
+    of their names. Any other files in the directory are ignored.
+    Again, <c>-p<e>N</e></c> can be used to override the default patch level.
+    Note that <c>eapply</c> will not recurse into directories.
```
mgorny, line 41:
```suggestion
    Note that <c>eapply</c> will not recurse into subdirectories.
```

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,123 @@
...
+
+<p>
+In the following example taken from <c>www-client/firefox</c>, <c>eapply</c>
+is called with a directory argument. It applies all patches found in that
+directory:
+</p>
+
+<codesample lang="ebuild">
+	eapply "${WORKDIR}/firefox-patches"
+</codesample>
```
mgorny, line 87: Considering what I said above, perhaps include a larger snippet showing how to fetch a patchset via `SRC_URI` and note how it's unpacked.

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,123 @@
...
+The default <c><uri link="::ebuild-writing/functions/src_prepare"/></c>
+function will look for a global PATCHES array to apply a list of patches
+for you.
+</p>
+
+<codesample lang="ebuild">
+PATCHES=(
+	"${FILESDIR}/${P}-destdir.patch"
+	"${FILESDIR}/${P}-parallel_build.patch"
+)
```
mgorny, line 99: Perhaps add some example comments and upstream patch links, to make it clear that we really want people documenting why they apply patches, and where they get them from.

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,123 @@
...
+<p>
+This example shows how different patch levels can be applied:
+</p>
+
+<codesample lang="ebuild">
+src_prepare() {
+	eapply -p2 "${WORKDIR}/${P}-suse-update.patch"
+	eapply -p0 "${FILESDIR}/${PV}-no-TIOCGDEV.patch"
+	eapply "${FILESDIR}/${PV}-gcc-6.patch"
+	eapply_user
```
mgorny, line 117: Perhaps make it clear that either that or `default` works.

We could even go further to demonstrate that you can combine `PATCHES` and explicit `eapply` but that's kinda ugly, so perhaps not.

ebuild-writing/functions/src_prepare/eapply/text.xml
ulm: I've reworded it so that "vanilla tarball" is replaced by "upstream tarball" and "a modified tarball" by "your own modified tarball".

ebuild-writing/functions/src_prepare/eapply/text.xml
ulm: Done.

ebuild-writing/functions/src_prepare/eapply/text.xml
ulm: I've added `SRC_URI+=` and a much simplified `src_prepare()`.

ebuild-writing/functions/src_prepare/eapply/text.xml
ulm: I've added a reference to the Patches chapter where all these things are explained. IMHO it doesn't make much sense to repeat it here.

ebuild-writing/functions/src_prepare/eapply/text.xml
ulm: IMHO explicit `eapply_user` should be preferred whenever an explicit `eapply` occurs. Having `default` alongside `eapply` looks skewed to me. :smile:

ebuild-writing/functions/src_prepare/eapply/text.xml
```diff
@@ -0,0 +1,134 @@
...
+
+<p>
+In the following simplified example taken from <c>www-client/firefox</c>,
+a patchset is added to <c>SRC_URI</c> in order to fetch and unpack it.
+<c>eapply</c> is then called with a directory argument. It applies all patches
+found in that directory:
+</p>
+
+<codesample lang="ebuild">
+SRC_URI+="https://dev.gentoo.org/~larry/patchsets/firefox-125-patches-01.tar.xz"
```
mgorny, line 87:
```suggestion
SRC_URI+="https://dev.gentoo.org/~larry/patchsets/firefox-${PV}-patches-01.tar.xz"
```

:-)

ebuild-writing/functions/src_prepare/eapply/text.xml
mgorny: Sure, but let's add some example comments nevertheless, for good measure.

ebuild-writing/functions/src_prepare/eapply/text.xml
ulm: Done (using `${P}` rather than `firefox-${PV}`).

ebuild-writing/functions/src_prepare/eapply/text.xml
ulm: Better now?

mgorny (approved):
LGTM

PR #332, opened 2024-04-24
eclass-writing: Update documention of DEAD tag

ulm:


eclass-writing/text.xml
```diff
@@ -114,8 +113,10 @@ batch push them at once where possible.
 <title>Removing Eclasses</title>
 <body>
 
-<p>No longer used eclasses may be removed from the tree, but developers must
-adhere to the following process:</p>
+<p>
+No longer used eclasses may be removed from the tree, but developers must
+adhere to the following process:
+</p>
```
thesamesam, line 119: I wonder if we should require a `@DEPRECATED` tag exist for some period beforehand (at the very least a month before removal, like we do for `@DEAD`), given otherwise nobody gets a pkgcheck warning about it...

ulm: It's mentioned in the `<note>` below the list.

Maybe we could clarify "Do this at least 30 days before removing the eclass." instead of "Do this for a period before removing the eclass."?

thesamesam: That would WFM.

eclass-writing/text.xml
```diff
@@ -114,8 +113,10 @@ batch push them at once where possible.
 <title>Removing Eclasses</title>
 <body>
 
-<p>No longer used eclasses may be removed from the tree, but developers must
-adhere to the following process:</p>
+<p>
+No longer used eclasses may be removed from the tree, but developers must
```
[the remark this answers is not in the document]
ulm: How about "Eclasses that are no longer in use"?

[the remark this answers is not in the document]
orlitzky: That's good too. So long as we avoid the reader having to back and parse the first three words as no-longer-used.

PR #334, opened 2024-04-29
general-concepts/licenses: Licenses must be plain text

ulm:
Historically we had some PDF files in the repository, but the last of them was removed in 2011.


thesamesam (approved):
Oh, wow, I didn't know we ever had non-plain-text. I'm interested in what those cases were but let's discuss that elsewhere.

PR #337, opened 2024-04-29
devbook-guide: Document that sentence case should be used in titles

ulm:


maffblaster (approved):
Looks great. Thank you, @ulm!

PR #342, opened 2024-06-06
eclass-writing: @DEPRECATED first

vaukai:


eclass-writing/text.xml
```diff
@@ -390,7 +390,10 @@ variables are as follows:
   <ti><c>@DEPRECATED:</c></ti>
   <ti>No</ti>
   <ti>Optionally, the name of any replacement variable</ti>
-  <ti>Declares that this variable should no longer be used in ebuilds</ti>
+  <ti>
+    Declares that this variable should no longer be used in ebuilds.
+    Must appear before the <c>@ECLASS_VARIABLE:</c> tag.
```
ulm, line 395: I don't think that's correct. The script will warn about an 'Unexpected tag in "funcvar" state' when `@DEPRECATED:` occurs before `@ECLASS_VARIABLE:`.

vaukai: > I don't think that's correct. The script will warn about an 'Unexpected tag in "funcvar" state' when `@DEPRECATED:` occurs before `@ECLASS_VARIABLE:`.

pls compare https://github.com/gentoo/gentoo/pull/37022/commits/74e7987ebf1b95629a8c82df36f3266552057379
i got `"EclassDocError: java-ant-2: failed parsing eclass docs: '@DEPRECATED:', line 84: missing inline arg"` when putting it after.

ulm: That's because the `@DEPRECATED:` tag needs an argument (use `none` when there's no replacement).

ulm:
Closing (per previous comments).

PR #345, opened 2024-08-12
Fix multiple typos and grammatical errors

SigHunter:
While reading the devmanual, I took some notes when I believe I found mistakes.
Those are - in my opinion - relatively straightforward corrections, mostly grammar and typos. 

I put it all in a single commit, I hope that is OK.

### DIFF
--- a/quickstart/text.xml
+++ b/quickstart/text.xml
@@ -74,7 +74,7 @@ At the top of the ebuild is a header block. This is present in all ebuilds.
 </p>
 
 <p>
-Ebuilds are indented using tabs, with each tab representing four places.
+Ebuilds are indented using tabs, with each tab representing four spaces.
 See <uri link="::ebuild-writing/file-format/"/>.
 </p>
 </body>
--- a/tasks-reference/pam/text.xml
+++ b/tasks-reference/pam/text.xml
@@ -80,15 +80,15 @@ The statement is composed of 3 or 4 tokens:
     methods.  Some documentation states that we need to specify the full path of
     the module, but this creates problems because not all the systems install the
     modules in the same place: Linux-PAM on Gentoo is generally set up to load
-    them from <c>/lib/security</c>, but for example on AMD64 this become
+    them from <c>/lib/security</c>, but for example on AMD64 this becomes
     <c>/lib64/security</c>. The
     result is that providing the full path will lead to non-working <c>pamd</c>
-    files, and the right way to handle this is just states the module name <d /> the
+    files, and the right way to handle this is to just state the module name <d /> the
     PAM implementation will take care of finding the module.
   </li>
   <li>
-    The last token, which can consist of multiple items, describe the parameters
-    passed to the module. These are module-dependent.
+    The last token, which can consist of multiple items, describes the
+    parameters passed to the module. These are module-dependent.
   </li>
 </ul>
 <ul>
@@ -124,7 +124,7 @@ source for some services, such as mail and ftp servers.
 
 <p>
 But for desktop systems, all the different services, such as mail servers, ftp
-servers, ssh and so on, just need to authenticate in the same way the users logs
+servers, ssh and so on, just need to authenticate in the same way the users log
 in to the system.
 </p>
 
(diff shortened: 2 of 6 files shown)


tasks-reference/pam/text.xml
```diff
@@ -80,14 +80,14 @@ The statement is composed of 3 or 4 tokens:
     methods.  Some documentation states that we need to specify the full path of
     the module, but this creates problems because not all the systems install the
     modules in the same place: Linux-PAM on Gentoo is generally set up to load
-    them from <c>/lib/security</c>, but for example on AMD64 this become
+    them from <c>/lib/security</c>, but for example on AMD64 this becomes
     <c>/lib64/security</c>. The
     result is that providing the full path will lead to non-working <c>pamd</c>
-    files, and the right way to handle this is just states the module name <d /> the
+    files, and the right way to handle this is just to state the module name <d /> the
```
thesamesam, line 86:
```suggestion
    files, and the right way to handle this is to just state the module name <d /> the
```

tools-reference/echo/text.xml
```diff
@@ -29,10 +29,10 @@ the same as the former, but they won't print the trailing newline
 
 <p>
 All usage of the form <c>echo ${somevar} | grep substring</c> just to
-check if the contents of the <c>${somevar}</c> variable
-contains <c>substring</c>, or more often, <c>echo ${somevar} |
+check if the content of the <c>${somevar}</c> variable
```
thesamesam, line 32: The changes in this file feel more like style and should perhaps be split into their own commit.

SigHunter: true, I got carried away :-)
I will keep the "third person S correction" in this commit and move the other stuff (commonly, rarely) to what else I'm preparing

tasks-reference/pam/text.xml
SigHunter: adapted, ty

eclass-writing/text.xml
```diff
@@ -187,7 +187,7 @@ to properly break lines when generating eclass manpages.
 
 <p>
 An eclass is a <c>bash</c> script which is sourced within the ebuild environment.
-Files should be a simple text file with a <c>.eclass</c> extension. Allowed
+Files should be a simple text file with an <c>.eclass</c> extension. Allowed
```
ulm, line 190: This can be pronounced as "a dot eclass extension", in which case "a" is correct.

SigHunter:
first round of edits done from my side, please review, thanks

quickstart/text.xml
```diff
@@ -74,7 +74,7 @@ At the top of the ebuild is a header block. This is present in all ebuilds.
 </p>
 
 <p>
-Ebuilds are indented using tabs, with each tab representing four places.
+Ebuilds are indented using tabs, with each tab representing four spaces.
```
ulm, line 77: That's not necessarily correct. A tab may represent less than 4 spaces if it doesn't start at a column that is divisible by 4. How about the following wording?
```suggestion
Ebuilds are indented using tabs, with a tab stop every four positions.
```

tasks-reference/pam/text.xml
```diff
@@ -80,14 +80,14 @@ The statement is composed of 3 or 4 tokens:
...
+    them from <c>/lib/security</c>, but for example on AMD64 this becomes
     <c>/lib64/security</c>. The
     result is that providing the full path will lead to non-working <c>pamd</c>
-    files, and the right way to handle this is just states the module name <d /> the
+    files, and the right way to handle this is to just state the module name <d /> the
     PAM implementation will take care of finding the module.
   </li>
   <li>
-    The last token, which can consist of multiple items, describe the parameters
+    The last token, which can consist of multiple items, describes the parameters
```
ulm, line 90: Long line. Please keep them below 80 characters.

tasks-reference/pam/text.xml
```diff
@@ -230,7 +230,7 @@ correct directory to use for the current implementation/arch.
 <p>
 When the PAM module doesn't provide a way to install the package by itself, such
 as a <c>Makefile</c> or an installation script, there are also the <c>dopammod</c> and
-<c>newpammod</c> functions which takes care of install the module in the right
+<c>newpammod</c> functions, which take care of installing the module in the right
```
ulm, line 233: Long line.

ulm (changes requested):

quickstart/text.xml
SigHunter: I was using [ebuild-writing/file-format](https://devmanual.gentoo.org/ebuild-writing/file-format/index.html#indenting-and-whitespace) as reference, where it says spaces

ulm: Hm, I guess for indenting it is correct.

eclass-writing/text.xml
SigHunter: I see. reverted

tasks-reference/pam/text.xml
SigHunter: should fit now, thanks for spotting

ulm (approved):

ulm:
Thank you, good to go from my side.
If there are no further comments from @thesamesam then I'm going to merge this.

thesamesam (approved):

ulm:
Merged in commit c08a1dc1072740182e5c4d059967db697897bed6.
Thanks again.

PR #351, opened 2024-10-14
general-concepts/ebuild-revisions: mention balancing examples against…

thesamesam:
… stable

Explicitly say that even 'no revision bump' examples should be balanced against whether a stable ebuild is being modified, as we want to minimize any risk of regressions there.

i.e. It might be that a change would not need a revbump for an ~arch-only ebuild, but if the ebuild is stable, the principle of being conservative ofr stable would mean a revbump is warranted.

### DIFF
--- a/general-concepts/ebuild-revisions/text.xml
+++ b/general-concepts/ebuild-revisions/text.xml
@@ -125,6 +125,11 @@ of thumb could be used as a guideline:
   </li>
 </ul>
 
+<p>
+Note that the examples above should still be balanced against the general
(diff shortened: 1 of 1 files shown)


ulm (commented):
Good point, but having it as a separate paragraph after the examples feels wrong to me (i.e. all important general information should be before the examples).

The 1st item of the numbered list above already says in parentheses "in the case of packages marked stable, every non-trivial change is classified as such". Maybe this could be split off into a separate item instead, so that it would be more pronounced?

ulm:
I have filed #355 for the new wording.

PR #352, opened 2024-10-24
Rename the document element from "guide" to "devbook"

ulm:
This has bugged me since quite some time, mainly because nXML mode in Emacs (or rather, `rng-loc`) locates the schema based on the XML document element. So, when looking at old documentation, it would see `<guide>`, load `devbook.rnc`, and would end up highlighting supposedly bad syntax.

Renaming the document element to `devbook` will avoid these problems, because `guide` can then be dropped from nXML mode's schema locating, and the vacuous schema will be used for GuideXML documents (which is not a problem when looking at them).

While at it, rename `<contentsTree/>` to `<contents/>` because it is the only element whose name is in camelCase. (No strong opinion about the name here, `contents` is what is used by Texinfo. So please speak up if you prefer `tableofcontents` (as in LaTeX), `toc` (as in Mediawiki), or even `childlist` (used in the old RST Devmanual).)

Timeline: If there are no objections, I would start with a transitional release of `app-emacs/nxml-gentoo-schemas` that would accept both the old `guide` and the new `devbook` element. The Devmanual would be updated some time after stabilisation of that version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,9 +87,9 @@ install: all
 
 validate: devbook.rng
 	@xmllint --noout --quiet --relaxng $< $(XMLS)
-	@# Check if /guide/@self agrees with the document path
+	@# Check if /devbook/@self agrees with the document path
 	@for file in $(XMLS); do \
-	  self=$$(xmllint --quiet --xpath 'string(/guide/@self)' $${file}); \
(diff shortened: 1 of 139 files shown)


thesamesam (approved):
This sounds OK. I trust you on the transition mechanics.

PR #355, opened 2024-10-29
general-concepts/ebuild-revisions: Almost always revbump when stable

ulm:
This reiterates PR #351.


thesamesam (approved):
Thank you!

On a procedural note, I wonder if it'd make sense to cherry-pick my commit here and then adjust it on top, but I really don't mind either way (the strategy may be useful in general even if not appropriate here).

ulm:
@thesamesam: Is this what you had in mind?

thesamesam:
Yes, thank you!

PR #359, opened 2024-11-04
Makefile: Secondary expansion is not needed

ulm:
Drop .SECONDEXPANSION and use a static pattern rule instead.

The problem addressed by secondary expansion was that % in a "normal" pattern rule doesn't match the empty string, therefore %index.html won't match the top-level index.html target. By contract, % in a static pattern rule matches any part of the target name.

Using a pattern rule like "ind%x.html: t%xt.xml ..." would also be possible but seems a little hackish.


orlitzky (approved):
Looks like an improvement, thanks.

PR #361, opened 2024-12-07
Add internal links to eclass documentation

ulm:


thesamesam (approved):
Lovely!

Flowdalic:
I really like that it's linked. However I wonder if we shouldn't introduce an `<eclass/>` element? So that
```xml
 <c><uri link="::eclass-reference/optfeature.eclass/"/></c>
```
becomes
```xml
 <eclass>optfeature.eclass</eclass>
```

ulm:
> I really like that it's linked. However I wonder if we shouldn't introduce an `<eclass/>` element? So that
> 
> ```
>  <c><uri link="::eclass-reference/optfeature.eclass/"/></c>
> ```
> 
> becomes
> 
> ```
>  <eclass>optfeature.eclass</eclass>
> ```
> 
> But that's just a suggestion.

I'd much prefer keeping `<uri>` for all links:
- Eclass links are just a type of internal links, and inventing a new tag for them would cause code duplication in several places of the XSL stylesheet. For example, all internal cross references are converted to _relative links_ in the HTML output.
- Introducing special cases for eclasses would complicate future changes, e.g. a single-page version, or directly generating XML from eclass documentation.
- There are instances with a different link text, which an `<eclass>` tag cannot do. Also, once we do XML from eclassdoc, we might want fragment identifiers, like `<uri link="::eclass-reference/elisp-common.eclass/#elisp-compile"/>`.
- A separate tag would make searching the source for internal references (and testing their validity) more difficult. (Currently it's as simple as `xmllint --xpath //uri/@link`.)

PR #363, opened 2024-12-16
general-concepts/slotting: New 'Considerations for slotting' section

thesamesam:
This is a section I've been wanting to write for quite some time and formalises a talk I've had to give many contributors.

Slotting is a powerful tool but easy to abuse and there's a number of non-obvious pitfalls that we should document: attempt to do that here.

### DIFF
--- a/general-concepts/slotting/text.xml
+++ b/general-concepts/slotting/text.xml
@@ -46,9 +46,125 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
 </body>
 
 <section>
+<title>Considerations for slotting</title>
+<body>
+
+<p>
+Slotting is a powerful tool and as such, is easily abused. It's critical
+that when slotting, package maintainers carefully reflect on implications
+for <e>reverse dependencies</e>.
+</p>
+
+<p>
+Do not be fooled by the ease in adapting the package itself to slotting:
+slotting the package itself may well be the easiest part.
+</p>
+
+<p>
+Before proceeding at all, consider the wider implications beyond Gentoo.
+Especially examine other distributions and their handling of the problem in
+question. Unilaterally moving software can have devastating consequences for
+both users and other developers.
+</p>
+
+<p>
+Suppose every possible consumer of a slotted package can be adapted in
+Gentoo: there is still more work to be done. Keep in mind that users may
+develop or test software on Gentoo systems which may be deployed elsewhere.
+They both want to be able to build that software without having to adapt to
+Gentooisms, and they also don't want to unintentionally write their software
+such that it relies on Gentooisms.
+</p>
+
+<p>
+For example, slotting software by inventing a Gentoo-specific pkgconfig
+file and then patching all packages in the Gentoo repository is usually
+the wrong decision, because users won't be able to build software manually
+without patching it, and they may ship software that then fails to build
+on other distributions.
+</p>
+
+<p>
+Even systems that lend themselves to slotting, such as Guile or Lua, cannot
+be naively slotted by just adapting the package itself. Other packages
+must be primed to handle a slotted landscape: an eclass mechanism usually
+needs to exist for packages to declare which versions they support, as well
+as force building against the version currently depended upon to avoid
+automagic dependencies.
+</p>
+
+<p>
+For example, for slotted Guile, it may be trivial to make <c>dev-scheme/guile</c>
+itself install itself as <c>guile-3.0</c> (and other <c>-3.0</c> suffixes
+on its binaries), but packages depending on Guile will need an eclass to
+control <e>which</e> Guile the installed files are compatible with.
+</p>
+
+<p>
+All reverse dependencies of a package should be tested and examined when
+considering any slotting approach, in particular asking the following:
+</p>
+
+<ul>
+  <li>how does this software use the to-be-slotted package;</li>
+  <li>how the build system should locate the (right) version of this package;</li>
+  <li>
+      how the package (and the right version of it) should be located at runtime,
+      if required.
+  </li>
+</ul>
+
+</body>
+
+<subsection>
+<title>Libraries</title>
+<body>
+
+<p>
+As a rule of thumb, libraries should not be slotted unless upstream
+explicitly intend (major) versions to be installable in parallel. The
+presumption should be that a library cannot be slotted until proven otherwise.
+</p>
+
+<p>
+If a library is to be slotted, please consult with upstream wherever possible:
+ask them to begin installing pkgconfig files to aid porting build systems
+to detecting the library portably.
+</p>
+
+<p>
+ABI should also be considered with slotted libraries: if a consumer of the
+library (in)directly links against another version of the library, loading
+two ABI-incompatible versions of the library in the same process, unpredictable
+runtime behaviour may occur (including crashes, or worse).
+</p>
+</body>
+</subsection>
+
+<subsection>
+<title>Executables</title>
+<body>
+
+<p>
+For utilities / executables, slotting may be acceptable as long as an
+unversioned, generic symlink or wrapper is made available for the benefit
+of user scripts or tools, and build systems which expect the generic name
+to be available.
+</p>
+
+<p>
+An <c>eselect</c> module is often made available for these cases, though
+ebuilds normally shouldn't depend on its use.
+</p>
+</body>
+</subsection>
+</section>
+
+<section>
 <title>Sub-slots</title>
 
 <body>
+
 <p>
 Sometimes a package installs a library that changes interfaces between versions,
 but it's undesirable or inconvenient to allow some of these versions to be installed
--- a/general-concepts/slotting/text.xml
+++ b/general-concepts/slotting/text.xml
@@ -17,6 +17,13 @@ in the ebuilds. This is purely a convention; the package manager does not treat
 <c>0</c> any different from other slot values.
 </p>
 
+<p>
+"Fake slotting" where multiple SLOTs exist for the same package but they
+are mutually exclusive (i.e. block each other) is strongly discouraged, given
+that the existence of multiple SLOTs implies to users that parallel installation
+is possible.
+</p>
+
 <note>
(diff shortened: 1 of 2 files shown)


thesamesam:
Not done a pass for formatting yet. Looking more for feedback on the higher-level layout and missing points (I suspect there's some more which will come to me shortly).

SoapGentoo:
thanks for providing something we can use as a reference for people suggesting slotting

SoapGentoo (approved):

CheyenneWills:
I don't have anything that creates a slotted install, but slots are very helpful and I do use slotted compilers (gcc and clang) quite a bit for testing.  

Would be nice to have a bit more standardized way of selecting the binary for slotted commands -- for example:

gcc -- gcc-config
clang -- env.d tricks
python - eselect

thesamesam:
Yeah, it's worth us listing methods for the binary itself (and maybe mentioning environment variables too and `python-exec`).

For libraries: we should perhaps mention things like `PKG_CONFIG_PATH` as we do for Guile too.

general-concepts/slotting/text.xml
```diff
@@ -45,6 +52,102 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+
+<p>
+Do not be fooled by the ease in adapting the package itself to slotting:
+slotting the package itself is always the easiest part.
+</p>
+
+<p>
+Before proceeding at all, consider the impact on the wider ecosystem
+beyond Gentoo. Especially examine other distributions and their handling
+of the problem in question. Gentoo cannot unilaterally move software to
```
mgorny, line 73: Well, technically speaking it can. I think the point's more like "it's not always a good idea".

general-concepts/slotting/text.xml
```diff
@@ -45,6 +52,102 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+dependencies.
+</p>
+
+<p>
+All reverse dependencies of a package should be tested and examined when
+considering any slotting approach, in particular asking the following:
+</p>
+
+<ul>
+  <li>how does this software use this package;</li>
```
mgorny, line 100: "Software" and "package" here can be ambiguous.

general-concepts/slotting/text.xml
```diff
@@ -45,6 +52,102 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+
+<p>
+Slotting is a powerful tool and as such, is easily abused. It's critical
+that when slotting, package maintainers carefully consider the impact
+on <e>reverse dependencies</e>.
+</p>
+
+<p>
+Do not be fooled by the ease in adapting the package itself to slotting:
+slotting the package itself is always the easiest part.
```
ulm, line 67: I think "always" is too strong here. How about
```suggestion
Do not be fooled by the ease in adapting the package itself to slotting:
slotting the package itself is often the easiest part.
```
or
```suggestion
Do not be fooled by the ease in adapting the package itself to slotting:
slotting the package itself may well be the easiest part.
```
?

general-concepts/slotting/text.xml
```diff
@@ -45,6 +52,102 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+</p>
+
+<p>
+Before proceeding at all, consider the impact on the wider ecosystem
```
ulm, line 71: I really dislike that "ecosystem" metaphor in computing (especially in free-software) context. It is a fuzzy and ambiguous term.

general-concepts/slotting/text.xml
```diff
@@ -45,6 +52,102 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+to Gentooisms, and they also don't want to unintentionally write their
+software such that it relies on Gentooisms.
+</p>
+
+<p>
+Ecosystems that lend themselves to slotting, such as Guile or Lua, cannot
+be naively slotted by just adapting the package itself. Other packages
+must be primed to handle a slotted world: an eclass mechanism usually needs
+to exist for packages to declare which versions they support, as well as
+force building against the version currently depended on to avoid automaigc
```
ulm, line 90: Typo:
```suggestion
force building against the version currently depended on to avoid automagic
```

general-concepts/slotting/text.xml
```diff
@@ -45,6 +52,102 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+<p>
+Even if every possible consumer of a slotted package has been adapted in
+Gentoo, remember that people may develop or test software on Gentoo systems:
+they both want to be able to build that software without having to adapt
+to Gentooisms, and they also don't want to unintentionally write their
+software such that it relies on Gentooisms.
+</p>
+
+<p>
+Ecosystems that lend themselves to slotting, such as Guile or Lua, cannot
```
ulm, line 86:
```suggestion
Systems that lend themselves to slotting, such as Guile or Lua, cannot
```

general-concepts/slotting/text.xml
```diff
@@ -45,6 +52,102 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+be naively slotted by just adapting the package itself. Other packages
+must be primed to handle a slotted world: an eclass mechanism usually needs
```
ulm, line 88: "World" has a specific meaning in Gentoo context, maybe better avoid?

general-concepts/slotting/text.xml
```diff
@@ -45,6 +52,102 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+two ABI-incompatible versions of the library in the same process, unpredictable
+runtime behaviour may occur (including crashes, or worse).
+</p>
+</body>
+</subsection>
+
+<subsection>
+<title>Executables</title>
+<body>
+<p>
```
ulm, line 136: Blank line before `<p>` please.

general-concepts/slotting/text.xml
```diff
@@ -45,6 +52,102 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+For utilities / executables, slotting may be acceptable as long as an
+unversioned, generic symlink or wrapper is made available for the benefit
+of user scripts or tools, and build systems which expect the generic name
+to be available.
+</p>
+
+<p>
+An <c>eselect</c> module is often made available for these cases, though
+ebuilds normally shouldn't depend its use.
```
ulm, line 145: Is the word "on" missing here?

general-concepts/slotting/text.xml
thesamesam: Can you suggest an alternative? In this case, I really want people to think "how is this going to affect other people", and "is there a thing this is going to break far beyond Gentoo"?

ulm: "Before proceeding, consider the wider implications beyond Gentoo."

general-concepts/slotting/text.xml
```diff
@@ -45,6 +45,102 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+<subsection>
+<title>Libraries</title>
+<body>
+
+<p>
+As a rule of thumb, libraries should not be slotted unless upstream
+explicitly intend (major) versions to be installable in parallel. The
+presumption should be that a library cannot be slotted until proven otherwise.
+</p>
+
```
vaukai, line 117: Is `dev-java/bcprov` regarded a `library`? If so, what' should have been the alternative to slotting in commit [901da76177](https://gitweb.gentoo.org/repo/gentoo.git/commit/dev-java/bcprov/bcprov-1.74.ebuild?id=901da761770cc5e16a9303ddcb0722f5b3654773) followed-up by commit [154e9224cf](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=154e9224cf297219ed0a5baf3cbc1e5eb5c42456), all was done according to [this wiki article](https://wiki.gentoo.org/wiki/Java_Developer_Guide#Slots)?

Flowdalic: I'd say Java libraries are by construction intended to be installable in parallel (they are often not mean to be consumed in parallel, but that is unrelated to slotting).

ulm:
Ping. No visible progress since half a year; do you still intend for this to be included?

thesamesam:
Yeah, I'll come back to this soon.

general-concepts/slotting/text.xml
thesamesam: I'll do "landscape".

general-concepts/slotting/text.xml
thesamesam: Yes.

general-concepts/slotting/text.xml
```diff
@@ -17,6 +17,13 @@ in the ebuilds. This is purely a convention; the package manager does not treat
 <c>0</c> any different from other slot values.
 </p>
 
+<p>
+"Fake slotting" where multiple SLOTs exist for the same package but they
+are mutually exclusive (i.e. block each other) are strongly discouraged, given
```
ulm, line 22: `"Fake slotting"` is the subject of the sentence, so the verb should be singular, I think:
```suggestion
"Fake slotting" where multiple SLOTs exist for the same package but they
are mutually exclusive (i.e. block each other) is strongly discouraged, given
```

general-concepts/slotting/text.xml
```diff
@@ -45,10 +52,108 @@ To <c>DEPEND</c> upon a package in a specific slot, refer to
...
+</p>
+
+<p>
+Systems that lend themselves to slotting, such as Guile or Lua, cannot
+be naively slotted by just adapting the package itself. Other packages
+must be primed to handle a slotted landscape: an eclass mechanism usually
+needs to exist for packages to declare which versions they support, as well
+as force building against the version currently depended upon to avoid
+automagic dependencies.
+</p>
```
ulm, line 92: I find these three paragraphs (but especially the last one) hard to understand. The text is somewhat vague and lacks specificity. Maybe it could profit from an example?

ulm: Yes, that's better.

PR #364, opened 2025-01-03
Pair emake -j1 with --shuffle=none

Flowdalic:
Thanks to trofi, GNU make 4.4 gained the --shuffle option to detect unsound Makefiles [1]. Gentoo's tinderboxes use --shuffle to find and report such issues. However, they also would report broken Makefiles where the ebuild author knew that the Makefile is broken, and hence added -j1 to the make invocation [2]..

To stop Tinderboxes from creating noise by reporting known issues, this change devmanual to suggest pairing -j1 with --shuffle=none for the *compile* case. Devmanual also suggest to consider -j1 for resource intensive testsuites, but this does not require --shuffle.

1: https://trofi.github.io/posts/249-an-update-on-make-shuffle.html
2: https://bugs.gentoo.org/947303#c10
Signed-off-by: Florian Schmaus <flow@gentoo.org>

### DIFF
--- a/ebuild-writing/functions/src_compile/building/text.xml
+++ b/ebuild-writing/functions/src_compile/building/text.xml
@@ -19,8 +19,8 @@ cleanly, it should be patched.
 </p>
 
 <p>
-If patching <e>really</e> isn't an option, <c>emake -j1</c> should be
-used. However, when doing this, please remember that you are seriously
+If patching <e>really</e> isn't an option, <c>emake -j1 --shuffle=none</c> should
+be used. However, when doing this, please remember that you are seriously
 hurting build times for many non-x86 users in particular. Forcing
 <c>-j1</c> can increase build times from a few minutes to an hour on
 some MIPS and SPARC systems.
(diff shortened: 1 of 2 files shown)


ulm:
This looks like an option that has a special and very narrow purpose.

Maybe this should be fixed on the tinderbox side instead?

Flowdalic:
> Maybe this should be fixed on the tinderbox side instead?

I would have hoped this as well and this was briefly discussed in #-dev and #-tinderbox, but I guess its not that easy, even though probably possible. Tinderboxes would need to hook into the `make` invocation and determine if `-j1` is active.

ulm:
> > Maybe this should be fixed on the tinderbox side instead?
> 
> I would have hoped this as well and this was briefly discussed in #-dev and #-tinderbox, but I guess its not that easy, even though probably possible. Tinderboxes would need to hook into the `make` invocation and determine if `-j1` is active.

Still sounds better to me than modifying many ebuilds. Also, what about `make` variants that don't support the option? https://trofi.github.io/posts/249-an-update-on-make-shuffle.html mentions passing the option via the `GNUMAKEFLAGS` variable rather than the command line.

ulm:
Closing because `--shuffle` is a GNUism and presumably shouldn't be unconditionally added to emake options.

PR #365, opened 2025-01-05
eclass-writing: Specify @BLURB more precisely

ulm:
The @BLURB description appears in the NAME section of the generated man page, and should therefore follow the rules from man-pages(7):

"All words in this line (including the word immediately following the '\-') should be in lowercase, except where English or technical terminological convention dictates otherwise."

Since it is a section heading, it should not end with a full stop either.


eclass-writing/text.xml
```diff
@@ -266,8 +266,9 @@ summarizes the available documentation tags:
   <ti>Yes</ti>
   <ti>Single line freetext</ti>
   <ti>
-    Contains a short description for the eclass. Must be on the same
-    line with the tag.
+    Contains a short description for the eclass. Must be on the same line
+    with the tag. All words, including the first, must be in lowercase, except
```
sbraz, line 270:
```suggestion
    as the tag. All words, including the first, must be in lowercase, except
```

sbraz (changes requested):

eclass-writing/text.xml
ulm: Well, that was part of the existing text. :smile: Updated.

orlitzky (approved):

PR #366, opened 2025-04-21
general-concepts/dependencies: Mention BDEPEND

e-luks:
It can be confusing that only DEPEND is mentioned here, as some dependencies belong in BDEPEND behind a test USE flag instead.

### DIFF
--- a/general-concepts/dependencies/text.xml
+++ b/general-concepts/dependencies/text.xml
@@ -743,8 +743,10 @@ tarball.
 
 <p>
 Packages often have optional dependencies that are needed only when running
-tests. These should be specified in DEPEND behind a USE flag. Often, the
-'test' USE flag is used for this purpose.
+tests. These should be specified in <c>BDEPEND</c> or <c>DEPEND</c> behind a
+USE flag
+(see also <uri link="::general-concepts/dependencies/#Build dependencies"/>).
+Often, the <c>test</c> USE flag is used for this purpose.
 </p>
 
 <p>


orlitzky:
Bug: https://bugs.gentoo.org/693916

general-concepts/dependencies/text.xml
```diff
@@ -743,8 +743,8 @@ tarball.
 
 <p>
 Packages often have optional dependencies that are needed only when running
-tests. These should be specified in DEPEND behind a USE flag. Often, the
-'test' USE flag is used for this purpose.
+tests. These should be specified in DEPEND or BDEPEND behind a USE flag. Often,
+the 'test' USE flag is used for this purpose.
```
ulm, line 747: First of all, sorry for the delay.

I only have a small comment about style: `BDEPEND` is a literal string, so should be written as `<c>BDEPEND</c>`. While at it, can you fix `DEPEND` and `test` too? (The latter should be just `<c>test</c>`, i.e. drop the quotation marks.)

https://devmanual.gentoo.org/appendices/devbook-guide/index.html#c-b-e-sub-and-sup

Flowdalic:
Shouldn't we also elaborate on when to use `DEPEND` and when to use `BDEPEND` for `USE=test` conditional dependencies? Otherwise, reading the devmanual may yield more questions than answers.

ulm:
> Shouldn't we also elaborate on when to use `DEPEND` and when to use `BDEPEND` for `USE=test` conditional dependencies? Otherwise, reading the devmanual may yield more questions than answers.

This is explained above in https://devmanual.gentoo.org/general-concepts/dependencies/index.html#build-dependencies, so adding a cross- reference to that section might be enough.

general-concepts/dependencies/text.xml
e-luks: No worries :slightly_smiling_face:

general-concepts/dependencies/text.xml
```diff
@@ -743,8 +743,8 @@ tarball.
 
 <p>
 Packages often have optional dependencies that are needed only when running
-tests. These should be specified in DEPEND behind a USE flag. Often, the
-'test' USE flag is used for this purpose.
+tests. These should be specified in <c>BDEPEND</c> or <c>DEPEND</c> behind a USE flag (see also <uri link="::general-concepts/dependencies/#build-dependencies"/>). Often,
+the <c>test</c> USE flag is used for this purpose.
```
ulm, line 747: Hm, internal cross references are a bit special in the devmanual, as they use the original section title, i.e. instead of `#build-dependencies` it should be `#Build dependencies`. Also avoid lines longer than 80 chars (if at all possible).
```suggestion
tests. These should be specified in <c>BDEPEND</c> or <c>DEPEND</c> behind
a USE flag
(see also <uri link="::general-concepts/dependencies/#Build dependencies"/>).
Often, the <c>test</c> USE flag is used for this purpose.
```

ulm (approved):

ulm:
Thanks, merged.

PR #368, opened 2025-06-04
ebuild-maintenance/removal: fix ebuild-name command list diff

henning-schild:
The list of commands talks about dev-util/pmk while actually dev-qt/qtphonon would get removed by the commands listed.

Fixes: 74b1623b6414 ("ebuild-maintenance: remove the Gentoo-Bug tag in the package removal subsection")

Related-to: https://bugs.gentoo.org/640788

### DIFF
--- a/ebuild-maintenance/removal/text.xml
+++ b/ebuild-maintenance/removal/text.xml
@@ -59,7 +59,7 @@ When removing packages follow these steps:
 </ol>
 
 <p>
-Here is a list of commands that will delete <c>dev-util/pmk</c>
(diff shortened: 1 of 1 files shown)


henning-schild:
@floppym came across this when looking into https://bugs.gentoo.org/show_bug.cgi?id=957136

but just reading so far

floppym (approved):

ulm:
Thanks.

henning-schild:
https://github.com/gentoo/devmanual/commit/6d86da88d43eb3c926738bccf4c177d3217614d7

PR #370, opened 2025-07-25
 keywording: Explain how Stabilization groups works, mention pkgdev bugs, remove mention of ia64

arthurzam:
- Finally found the time to explain how stabilization groups works [1], how to use them. I think this is useful info for devmanual, especially since they are already used a lot in ::gentoo.

- Also dropped the mentions of ia64, found other arch to mention in each place (didn't want to just drop since it breaks the wrapping and I was too lazy to handle it)

- Also added sample invocation of pkgdev bugs to file bugs for stabilization candidates by maintainer email.

[1] https://public-inbox.gentoo.org/gentoo-dev/49ec2455-9055-c9f7-9619-723f0d2a7ebf@gentoo.org/

### DIFF
--- a/keywording/text.xml
+++ b/keywording/text.xml
@@ -502,6 +502,11 @@ There are several tools available that can help with this:
     Use <c>pkgcheck</c>'s <c>StableRequest</c> check, e.g.
       <c>grep -ri "larry@" */*/metadata.xml -l | cut -d'/' -f1-2 | xargs pkgcheck scan -k StableRequest</c>
   </li>
+  <li>
+    Use <c>pkgdev bugs</c> for finding and filing stabilization bugs, e.g.
+    <c>pkgdev bugs --edit-graph --filter-stablereqs
+    --auto-cc-arches larry@gentoo.org --find-by-maintainer larry@gentoo.org</c>
+  </li>
 </ul>
 
 </body>


keywording/text.xml
```diff
@@ -421,14 +421,53 @@ dependency for your package).
...
+
+<p>
+Stabilization groups are sets of packages defined to help organize and automate
+stabilization requests. These groups are stored under
+<c>metadata/stabilization-groups/</c>, and all files within this directory
+(including subdirectories) are included recursively.
+</p>
+
+<p>
+Each group is named according to its path, for example <c>@gnome/vte</c>. This
```
sbraz, line 444:
```suggestion
Each group is named after its path, for example <c>@gnome/vte</c>. This
```

keywording/text.xml
```diff
@@ -421,14 +421,53 @@ dependency for your package).
...
+naming convention helps avoid collisions and makes referencing groups
+straightforward. The files use a simple format: one <c>cat/pkg</c> per line,
+with lines starting with <c>#</c> treated as comments. Whitespace before and
+after entries is ignored.
+</p>
+
+<p>
+When invoking tools like <c>pkgdev bugs</c>, passing a set (anything starting
```
sbraz, line 452: I assume that only works locally, meaning I couldn't use a set directly in the stabilization bug's package list?

keywording/text.xml
```diff
@@ -421,14 +421,53 @@ dependency for your package).
...
+</p>
+
+<p>
+During bug creation, packages whose latest matching version is already
+stabilized are dropped from the list. The tool then builds the full dependency
+graph as usual, adding any additional required packages. For each defined set,
+all packages included in both the graph and the set are merged into a single
+bug. For example, if <c>@gnome/vala</c> contains <c>dev-libs/vala-common</c> and
+<c>dev-lang/vala</c>, and both are required, they will be grouped into one bug.
+Packages in the set that are not required or requested are not included.
```
sbraz, line 465: How can a package in the set be "not requested"? I don't understand that part.

keywording/text.xml
arthurzam: Waiting for ACK from anyone else (I'm also far from native speaker)

thesamesam: Either are OK but sbraz's suggestion sounds better.

keywording/text.xml
arthurzam: @sbraz so the usage of stabilization groups is two fold:

1. expand mentions of `@set` into all packages "as requested". But you can also manually request packages, including subset of the group.
2. force-group packages from the same group into one bug, so they are done together.

thesamesam (approved):

keywording/text.xml
```diff
@@ -421,14 +421,62 @@ dependency for your package).
...
+Stabilization groups are sets of packages defined to help organize and automate
+stabilization requests. These groups are stored under
+<c>metadata/stabilization-groups/</c>, and all files within this directory
+(including subdirectories) are included recursively.
+</p>
+
+<p>
+The main purpose of stabilization groups is to ensure that all packages in the
+same group are handled together in a single stabilization bug. This allows
+them to be stabilized in an "atomic" way-either all packages in the group are
```
sbraz, line 446:
```suggestion
them to be stabilized in an "atomic" way—either all packages in the group are
```
I would go for an em dash based on https://en.wikipedia.org/wiki/Dash#Simple_equivalence_(or_near-equivalence)_of_paired_parenthetical_marks

sbraz: Or a colon? The standard hyphen just looks weird IMHO:
```suggestion
them to be stabilized in an "atomic" way: either all packages in the group are
```

keywording/text.xml
arthurzam: Yes, https://pkgcore.github.io/pkgdev/man/pkgdev/bugs.html

ulm:
LGTM. Merged.

PR #373, opened 2025-09-17
*: introduce <pkg> tag for nicer package links

arthurzam:
When I was writing the part for destabaling arch, I noticed we don't have any nice link/render for package (for the text "use qkeyword from app-portage/portage-utils").

I've decided to add a new tag `<pkg>` which renders in similar way to how we do it in Gentoo Wiki, with link to packages.g.o, a small nice "package" icon after it.

Then I changed all the places where we refer to specific package (and not as just a generic example or source for example). When a reader will want to go to the package page.

@ulm I'm find with rejecting this PR if you think this is useless or not needed.

### DIFF
--- a/appendices/contributing/text.xml
+++ b/appendices/contributing/text.xml
@@ -55,10 +55,10 @@ for those who prefer to submit patches using pull requests.
 
 <p>
 To build the devmanual, simply run <c>make</c> in the top directory of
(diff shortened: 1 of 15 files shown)


ulm:
Ping. Any other opinions? I'll work on this if anybody is going to second the request. Otherwise, I'll close this PR.

ulm:
Let's continue in https://codeberg.org/gentoo/devmanual/pulls/6.

PR #376, opened 2025-09-20
general-concepts/dependencies: mention common pitfalls

arthurzam:
This commit copies from @mgorny 's blog posts about common pitfalls in writing dependencies, and adds them to the devmanual.

I've added his sign off (based on https://bugs.gentoo.org/587526#c2) but it might have been wrongly done - but I don't really know the correct way to attribute here (I do want to mention him as the super main source).

Closes: https://bugs.gentoo.org/587526


general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+
+<codesample lang="ebuild">
+# Bad (if app-misc/foo is slotted):
+DEPEND="&gt;=app-misc/foo-2
+	&lt;app-misc/foo-5"
+</codesample>
+
+<p>
+If <c>app-misc/foo</c> is slotted and two installed slots each satisfy one
+side of the range, the resolver can legally use them separately. Likewise,
```
sbraz, line 885: I'd add a line break or some kind of separator before "Likewise"

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+# Bad (if app-misc/foo is slotted):
+RDEPEND="app-misc/foo[foo]
+	bar? ( app-misc/foo[baz] )"
+</codesample>
+
+<p>
+may be met by different slots with different USE configurations. For
+packages that are not truly multi-slotted (i.e. slots represent mutually
+incompatible ABIs), the safe fix is to state the intended slot explicitly or
+to express the intent via any-of groups limited per slot.
```
sbraz, line 899: Maybe provide examples for both proposed solutions? The examples below correspond to other cases, right?

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+</subsection>
+
+<subsection>
+<title>The <c>:=</c> slot operator with multiple slots</title>
+<body>
+
+<p>
+The <c>:=</c> operator records the slot/sub-slot of the best matching
+installed version for the given atom (see PMS). If that atom can match
+newer slots than the versions you explicitly allowed elsewhere, it may bind
```
sbraz, line 940:
```suggestion
slots newer than the versions you explicitly allowed elsewhere, it may bind
```
sounds better IMHO

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+<body>
+
+<p>
+For extended discussion with examples, see the
+<uri link="https://blogs.gentoo.org/mgorny/2016/06/21/dependency-pitfalls-regarding-slots-slot-ops-and-any-of-deps/"
+>blog post</uri>
+and its
+<uri link="https://blogs.gentoo.org/mgorny/2016/06/22/dependency-classes-and-allowed-dependency-types/"
+>follow-up</uri>,
+and the relevant sections of PMS on any-of dependency specifications and slot
```
sbraz, line 1150: Can you link to https://projects.gentoo.org/pms/8/pms.html#x1-760008.2.3 and https://projects.gentoo.org/pms/8/pms.html#x1-820008.3.3 while you're at it?

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+</body>
+</subsection>
+
+<subsection>
+<title>The <c>:=</c> slot operator across classes</title>
+<body>
+
+<p>
+The <c>:=</c> operator is technically valid in all classes but is useful only
+when the matching package is installed at the time metadata are recorded
```
sbraz, line 1072: Should we use the plural for metadata? Why not "metadata is recorded"?

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+    in <c>RDEPEND</c> rather than used purely in <c>DEPEND</c>.
+  </li>
+
+</ul>
+
+<codesample lang="ebuild">
+# Weak blocker — meaningful in RDEPEND
+RDEPEND="!app-misc/foo"
+
+# Strong blocker — enforced pre-build or pre-install
```
sbraz, line 1132: Maybe add a third example where it's used in DEPEND and change this description?
```suggestion
# Strong blocker — enforced pre-install
```

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+
+<p>
+Multiple independent atoms are not guaranteed to be satisfied by the same
+slot of a slotted package. For example, a common version range like the
+following can be satisfied by two different slots:
+</p>
+
+<codesample lang="ebuild">
+# Bad (if app-misc/foo is slotted):
+DEPEND="&gt;=app-misc/foo-2
```
mgorny, line 879: Since you're using `db` example below, perhaps use it here as well?

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+		dev-libs/A
+		dev-libs/B
+		dev-libs/C
+	)
+"
+</codesample>
+
+<p>
+If more than one alternative is installed, the package manager may use any
+of them. Replacing one alternative with another later still satisfies the
```
mgorny, line 996: I would be careful with "package manager using" anything, given that the package manager doesn't really do anything here, beyond making sure that at least one of the atoms is satisfied. It's the ebuild / package code itself deciding which will be used.

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+"
+
+# Better:
+RDEPEND="
+	|| (
+		dev-libs/A
+		dev-libs/B
+	)
+	dev-libs/A:=
+"
```
mgorny, line 1038: Yeah, I don't think this illustrates anything. In the end, B is completely redundant here, except that it technically may end up being installed via some misbehavior.

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+<c>:=</c> should be present in <c>RDEPEND</c> to express the rebuild
+relationship, and <c>DEPEND</c> must guarantee that a matching package is
+installed at the relevant time.
+</p>
+
+<codesample lang="ebuild">
+# Typical pattern — ensure the provider is present at build time and tie
+# rebuilds to its sub-slot changes
+DEPEND="dev-libs/foo:="
+RDEPEND="dev-libs/foo:="
```
mgorny, line 1083:
```suggestion
RDEPEND="${DEPEND}"
```

Perhaps with a comment that it evalutes to `dev-libs/foo:=`.

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+</codesample>
+
+<note>
+The atoms need not be textually identical as long as <c>DEPEND</c> guarantees
+that some package matching the <c>RDEPEND</c> <c>:=</c> atom is installed when
+metadata are recorded.
```
mgorny, line 1089: Maybe add a `DEPEND="dev-libs/foo"` example with `# also valid`.

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+  </li>
+  <li>
+    Strong blockers (<c>!!</c>) must be resolved before the dependency is
+    considered satisfied. They therefore make sense in both <c>DEPEND</c>
+    (before building) and <c>RDEPEND</c> (before installation).
+  </li>
+  <li>
+    In <c>PDEPEND</c>, weak blockers act effectively like those in
+    <c>RDEPEND</c>, while strong blockers behave more like weak ones because
+    satisfaction may be delayed post-install.
```
mgorny, line 1119: I don't think that's true, is it?

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+  </li>
+  <li>
+    Remember the general caveat from above: weak blockers should be included
+    in <c>RDEPEND</c> rather than used purely in <c>DEPEND</c>.
+  </li>
+
+</ul>
```
ulm, line 1126: No blank line here.
```suggestion
  </li>
</ul>
```

general-concepts/dependencies/text.xml
ulm, line 1132: Please, no literal em-dashes in the source. Use `<d/>` where it's possible, and maybe a colon in code samples.
```suggestion
# Weak blocker: meaningful in RDEPEND
RDEPEND="!app-misc/foo"

# Strong blocker: enforced pre-build or pre-install
```

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,309 @@ USE flags. This would then break building your ebuild.
...
+
+<p>
+For extended discussion with examples, see the
+<uri link="https://blogs.gentoo.org/mgorny/2016/06/21/dependency-pitfalls-regarding-slots-slot-ops-and-any-of-deps/"
+>blog post</uri>
+and its
+<uri link="https://blogs.gentoo.org/mgorny/2016/06/22/dependency-classes-and-allowed-dependency-types/"
+>follow-up</uri>,
+and the relevant sections of PMS on any-of dependency specifications and slot
+dependencies.
```
ulm, line 1151:
```suggestion
<uri link="https://blogs.gentoo.org/mgorny/2016/06/21/dependency-pitfalls-regarding-slots-slot-ops-and-any-of-deps/">
blog post</uri>
and its
<uri link="https://blogs.gentoo.org/mgorny/2016/06/22/dependency-classes-and-allowed-dependency-types/">
follow-up</uri>,
and the relevant sections of PMS on any-of dependency specifications and slot
dependencies.
```

arthurzam:
@ulm please merge it if you are fine with this PR

general-concepts/dependencies/text.xml
```diff
@@ -850,5 +850,312 @@ USE flags. This would then break building your ebuild.
...
+	|| (
+		dev-libs/A:=
+		dev-libs/B:=
+	)
+"
+</codesample>
+
+<codesample lang="ebuild" caption="Example of correct use of := outside any-of">
+IUSE="a b"
+REQUIRED_USE="^^ (a b)"
```
thesamesam:
```suggestion
REQUIRED_USE="^^ ( a b )"
```

general-concepts/dependencies/text.xml
thesamesam: I think we should drop it. I'm not convinced it's guaranteed.

ulm:
> @ulm please merge it if you are fine with this PR

@thesamesam still had some comments. (@arthurzam: I can apply these two changes when merging if that's o.k.)

ulm:
Merged, with the whitespace change applied and the `PDEPEND` bullet point dropped.

